PR #55, opened 2023-03-23
Fix 5.7 replication throttler

morgo:
The 5.7 replication throttler was not working correctly, and there were no tests. This fixes it and adds the most basic end to end tests.

See #56 #57 and #58 for additional work that will be needed later.

### DIFF
--- a/.github/workflows/mysql5712.yml
+++ b/.github/workflows/mysql5712.yml
@@ -1,0 +1,22 @@
+name: MySQL 5.7.12 (Aurora version)
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.19
+
+      - name: Deploy MySQL
+        run: scripts/dbdeployer_install_5712.sh
+
+      - name: Test
+        run: MYSQL_DSN="msandbox:msandbox@tcp(127.0.0.1:5712)/test" go test -v ./...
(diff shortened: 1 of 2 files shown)


pkg/throttler/basereplicationthrottler.go
```diff
@@ -0,0 +1,32 @@
+package throttler
```
daniel-nichter: I'd change this filename and type name because it's an egregious case of "Go stutter". Since Go leverages (and encourages) package context naming, `base.go` and `type Base struct` are idiomatic. Then you get `throttler.Base` which is very Go-esque.

Related, I highly suggest avoiding https://hackmysql.com/golang/go-antipatterns/#narrative-naming

daniel-nichter: >Long names don't automatically make things more readable.
https://go.dev/doc/effective_go#names

pkg/throttler/basereplicationthrottler.go
```diff
@@ -0,0 +1,32 @@
...
+
+type BaseReplicationThrottler struct {
+	replica          *sql.DB
+	lagToleranceInMs int64
+	currentLagInMs   int64
+	logger           loggers.Advanced
+}
+
+func (l *BaseReplicationThrottler) IsThrottled() bool {
+	return atomic.LoadInt64(&l.currentLagInMs) >= l.lagToleranceInMs
```
daniel-nichter: The "InMs" suffix is varying the style. Other vars use just a "Ms" suffix. But really the suffix is a type of anti-pattern and flawed design: use `time.Duration` instead. There was an article someone wrote about this, but of course now I can't find it. The TL;DR of that article was just thing: `lag time.Duration` not `lagMs int`  or `lagS int` etc.

morgo: I will fix this later. I have https://github.com/squareup/spirit/issues/28 to track the chunker target in ms, and will fix the replication target at the same time.

PR #60, opened 2023-03-24
Factor out checks package

### LINKED ISSUE #56: Write preflight checks for replication is healthy
If there is a replication throttler, it should pre-flight check replication is HEALTHY (IO and SQL Thread running: Yes) and fail the migration if not.

### LINKED ISSUE #58: Add tests for minimal privilege (main process and replica throttler DSN)
We should verify what these are and document them.

There should be preflight checks that validate that the privileges exist.

### LINKED ISSUE #59: 24: factor out preflight checks.

morgo:
This moves the various checks and tests into their own package, where they can easily be tested. The test-coverage is a little bit weak on them right now, but it can easily be improved in future.

Fixes #58 
Fixes #56
Fixes #59
Fixes https://github.com/squareup/spirit/issues/62

This also fixes a potential leak in the throttler package that I noticed when thinking about if the replica connection leaks.

### DIFF
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -5,12 +5,13 @@ package check
 import (
 	"context"
 	"database/sql"
+	"sync"
 
 	"github.com/siddontang/loggers"
 	"github.com/squareup/spirit/pkg/table"
 )
 
-// ScopeFlag a checker's scope
+// ScopeFlag scopes a check
 type ScopeFlag uint8
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/check/check.go
```diff
@@ -0,0 +1,56 @@
...
+	name     string
+	callback func(context.Context, Resources, loggers.Advanced) error
+	scope    ScopeFlag
+}
+
+var checks []check
+
+// registerCheck registers a check (callback func) and a scope (aka time) that it is expected to be run
+func registerCheck(name string, callback func(context.Context, Resources, loggers.Advanced) error, scope ScopeFlag) {
+	checks = append(checks, check{name: name, callback: callback, scope: scope})
```
joycse06: I don't see how it can happen with current setup, but it's probably better to guard against registering a check `> 1` times by using a map and check for the key?

pkg/check/configuration.go
```diff
@@ -0,0 +1,42 @@
+package check
+
+import (
+	"context"
+	"errors"
+
+	"github.com/siddontang/loggers"
+)
+
+func init() {
```
joycse06, line 10: Is this guaranteed to just run once?

morgo: Yeah, I think so! The current behavior is also just one execution. I did look it up though, and I couldn't see it the exactly only one semantic described though.

FWIW I copied this design from how RegisterRule works in skeema. But it [does look like they are stored in a map.](https://github.com/skeema/skeema/blob/main/internal/linter/linter.go#L163) I'll fix it, it's simple enough.

pkg/check/check.go
morgo: I changed this to a map, and also added a mutex. It should be safe now.

PR #70, opened 2023-03-30
Fix bug in resume-from-checkpoint, workaround large-last-chunk

morgo:
This fixes a bug where the resume from checkpoint seems to have recently been broken. It is regrettably difficult to E2E test due to https://github.com/squareup/spirit/issues/69

This also includes a workaround (not a fix) for https://github.com/squareup/spirit/issues/68 (lower `tableStatUpdateInterval`).

I changed the checkpoint `checkpointDumpInterval` and `statusInterval` to be less frequent as well.

### DIFF
--- a/pkg/row/copier.go
+++ b/pkg/row/copier.go
@@ -35,6 +35,7 @@ type Copier struct {
 	CopyChunksCount   int64
 	EtaRowsPerSecond  int64
 	isInvalid         bool
+	isOpen            bool
 	Throttler         throttler.Throttler
 	logger            loggers.Advanced
 }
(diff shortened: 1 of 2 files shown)


pkg/row/copier.go
```diff
@@ -133,8 +135,12 @@ func (c *Copier) isHealthy() bool {
 
 func (c *Copier) Run(ctx context.Context) error {
 	c.CopyRowsStartTime = time.Now()
-	if err := c.chunker.Open(); err != nil {
-		return err
+	if !c.isOpen {
+		// For practical reasons resume-from-checkpoint
```
daniel-nichter, line 139: What are "practical reasons"? i.e. future engineers won't know (and we'll forget in a few months)

morgo: The resume func has to call `OpenAtWatermark()` (with copyRowsAt) instead of `Open()`. If we changed Open to here we'd need to if it's a resume do this, otherwise do that and keep track of the initial copyRowsAt in the struct. This is more messy than tracking if `isOpen` or not.

PR #74, opened 2023-04-02
A number of fixes for large migrations

morgo:
- Fixes https://github.com/squareup/spirit/issues/67 (parallel flushing, exhaustive flushing *after* checksum)
- Fixes https://github.com/squareup/spirit/issues/66 - a problem where the status output was missing during ApplyChangeset and Checksum progress. 
- Removes the spam from the log by canal.


pkg/repl/logwrapper.go
```diff
@@ -0,0 +1,114 @@
+package repl
+
+import (
+	"github.com/siddontang/loggers"
+)
+
+// Canal accepts a loggers.advanced. If we pass our loggers.Advanced directly,
+// it will write a bunch of spam to the log. So we use this hack
+// to filter out the noisiest messages.
```
daniel-nichter, line 9: :-(  Some day soon, let's replace `canal`, then. This is heavy-weight noise (almost dead code) just to work around other code.

morgo: I agree with that - it's not turned out to be the most successful technical choice.

PR #77, opened 2023-04-03
Small cleanups from long running migration

morgo:
This adds a few small features:
- Record the checksum time, and in the last line summary show checksum + copy time independently
- Cleanup log format consistency
- Rename a few copier variables to remove stutter, and rename "MigrateChunk" to "CopyChunk".
- Fixes https://github.com/squareup/spirit/issues/73

The default threshold for the replication flushing to be considered "trivial" is increased to 10K. This is because this number of rows can be flushed pretty quickly. 1000 proved to be racy in testing where the number hovers just above it because of batched insert clients.

### DIFF
--- a/pkg/repl/logwrapper.go
+++ b/pkg/repl/logwrapper.go
@@ -1,6 +1,8 @@
 package repl
 
 import (
+	"fmt"
+
 	"github.com/siddontang/loggers"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/repl/logwrapper.go
```diff
@@ -39,8 +41,11 @@ func (c *LogWrapper) Warnf(format string, args ...interface{}) {
 func (c *LogWrapper) Errorf(format string, args ...interface{}) {
 	// Noisy bug on close, can be ignored.
 	// https://github.com/squareup/spirit/pull/65
-	if format == "canal start sync binlog err: %v" && args[0] == "Sync was closed" {
-		return
+	if len(args) == 1 {
+		message := fmt.Sprintf("%s", args[0])
+		if format == "canal start sync binlog err: %v" && message == "Sync was closed" {
```
daniel-nichter, line 46: Didn't notice this before. Another reason to stop using `canal` as soon as we can.

morgo: Because it's a race, I didn't notice it wasn't filtering. It needs to convert `args[0]` to a string for this to match and filter out.

Issue #78, opened 2023-04-05
Improve time estimates


morgo:
The sampling average might be too small?

Frequently the ETA is off by 2x versus the actual time. At least it over estimates:
```
info migration status: state=copyRows, copy-progress=8.92%, binlog-deltas=0, time-total=37m37.122483812s, eta=13h41m17s, copier-is-throttled=false
```

PR #85, opened 2023-04-10
fix copy estimates

morgo:
This makes the following changes:
- Move copy estimates into 'row' package (the responsibility of copier)
- Use uint64 for row counts, since wrapping could quite possibly occur.
- Fix https://github.com/squareup/spirit/issues/84
- Fix https://github.com/squareup/spirit/issues/78
- We wait 1 minute before providing any estimates (since the initial part of the copy is always much faster)

The estimates now use one of two methods:
1. If the PK is auto_inc it uses the "logical row copy count" and divides by the highest auto_inc value in the table.
2. Otherwise, it uses the "row copy count" and divides by the estimated rows.

The first method matches a lot of our tables, and is more stable since the estimated rows can jump around a lot. The estimates for the second are improved because of the bug in #78, but also the progress doesn't just shows a percentage but a number _over_ total. User's can now intuitively see progress even if the percentage decreases because the numerator is still growing.

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -765,6 +765,7 @@ func (m *Runner) updateTableStatisticsContinuously(ctx context.Context) {
 		if err := m.table.UpdateTableStatistics(ctx); err != nil {
 			m.logger.Errorf("error updating table statistics: %v", err)
 		}
+		m.logger.Infof("table statistics updated: estimated-rows=%d pk[0].max-value=%v", m.table.EstimatedRows, m.table.MaxValue())
 	}
 }
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Spirit is a _clone_ of the schema change tool [gh-ost](https://github.com/github
 
 It works very similar to gh-ost except:
 - It only supports MySQL 5.7 and MySQL 8.0 (it is anticipated only 8.0 will be supported once Aurora v2 reaches end of life)
-- It is multi-threaded in the row-copying phase
+- It is multi-threaded in both the row-copying and the binlog applying phase
 - It supports resume-from-checkpoint
 
 The goal of spirit is to apply schema changes much faster than gh-ost. This makes it unsuitable in the following scenarios:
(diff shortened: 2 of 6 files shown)


pkg/migration/runner.go
```diff
@@ -520,7 +517,7 @@ func (m *Runner) createCheckpointTable(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	query = fmt.Sprintf("CREATE TABLE `%s`.`%s` (id int NOT NULL AUTO_INCREMENT PRIMARY KEY, copy_rows_at TEXT, binlog_name VARCHAR(255), binlog_pos INT, copy_rows BIGINT, alter_statement TEXT)",
+	query = fmt.Sprintf("CREATE TABLE `%s`.`%s` (id int NOT NULL AUTO_INCREMENT PRIMARY KEY, low_watermark TEXT,  binlog_name VARCHAR(255), binlog_pos INT, rows_copied BIGINT, rows_copied_logical BIGINT, alter_statement TEXT)",
```
joycse06, line 520: How is this going to work for existing tables or do we always clean up checkpoint tables after a migration succeeds/fails?

pkg/row/copier.go
```diff
@@ -181,6 +192,78 @@ func (c *Copier) SetThrottler(throttler throttler.Throttler) {
...
+		if err != nil {
+			maxValue = c.table.EstimatedRows
+		}
+		pct := float64(copyRows) / float64(maxValue) * 100
+		return copyRows, maxValue, pct
+	}
+	// This is the legacy estimation method, which is not as accurate as the one above.
+	// It is required for scenarios like VARBINARY primary keys. The downside here is that
+	// the estimated rows can jump around a lot on a big table with a high variability of row size.
+	// Because we include the CopyRowsCount to users at least it will *at least*
```
joycse06: Remove one "at least" in this line.

pkg/migration/runner.go
morgo: It will cause resume to fail, which auto-restarts the migration from scratch. It just means we can't deploy while active migrations are running, which is nothing new :-)

PR #90, opened 2023-04-11
Move routines out of migration package and fix slow last chunk

morgo:
Fixes https://github.com/squareup/spirit/issues/86
Fixes https://github.com/squareup/spirit/issues/89

**Edit:** Also fixes https://github.com/squareup/spirit/issues/68 now

### DIFF
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -20,6 +20,7 @@ const (
 	ScopePostSetup   ScopeFlag = 1 << 1
 	ScopeCutover     ScopeFlag = 1 << 2
 	ScopePostCutover ScopeFlag = 1 << 3
+	ScopeTesting     ScopeFlag = 1 << 4
 )
 
 type Resources struct {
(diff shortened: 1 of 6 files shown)


pkg/table/tableinfo.go
```diff
@@ -230,14 +237,51 @@ func (t *TableInfo) setMinMax(ctx context.Context) error {
...
+// Close closes the tableInfo and stops the goroutine that updates the table statistics.
+func (t *TableInfo) Close() error {
+	t.isClosed = true
+	return nil
+}
+
+// AutoUpdateStatistics starts a goroutine that updates the table statistics every interval.
+// This will continue until Close() is called on the tableInfo.
+func (t *TableInfo) AutoUpdateStatistics(ctx context.Context, interval time.Duration, logger loggers.Advanced) {
+	go func() {
```
daniel-nichter: https://hackmysql.com/golang/go-antipatterns/#hidden-goroutine

`go tbl.AUpdateStatistics(...)` is more clear clear at the caller/when called, and it makes the func more testable.

PR #103, opened 2023-04-25
Refactor migration.Runner to store reference to Migration

joycse06:
Fixes https://github.com/squareup/spirit/issues/101

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -97,7 +97,7 @@ type Runner struct {
 
 func NewRunner(m *Migration) (*Runner, error) {
 	r := &Runner{
-		migration: migration,
+		migration: m,
 		logger:    logrus.New(),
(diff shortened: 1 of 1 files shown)


pkg/migration/runner.go
```diff
@@ -97,164 +91,144 @@ type Runner struct {
...
-	optAttemptInplaceDDL bool
-	optChecksum          bool
-	optReplicaDSN        string
-	optReplicaMaxLag     time.Duration
-
 	// Attached logger
 	logger loggers.Advanced
 }
 
 func NewRunner(migration *Migration) (*Runner, error) {
```
morgo: I think we can single letter this too:
```suggestion
func NewRunner(m *Migration) (*Runner, error) {
```

Then change the reference on line 100 to:
```
migration: m,
```

PR #114, opened 2023-05-01
table: improve coverage

morgo:
This increases the table test coverage from 75.2% to 86.9%

Almost everything is tested except `ExtractPrimaryKeyFromRowImage`, but we can do that in https://github.com/squareup/spirit/issues/107

### DIFF
--- a/pkg/check/dropadd.go
+++ b/pkg/check/dropadd.go
@@ -1,0 +1,51 @@
+package check
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/pingcap/tidb/parser"
+	"github.com/pingcap/tidb/parser/ast"
+	"github.com/siddontang/loggers"
+)
+
+func init() {
+	registerCheck("dropadd", dropAddCheck, ScopePreflight)
+}
+
+// dropAddCheck checks for a DROP and then ADD in the same statement.
+// This is unsupported per https://github.com/squareup/spirit/issues/102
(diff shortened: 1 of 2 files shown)


pkg/table/datum_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	newsigned := signed.Add(10)
+	newunsigned := unsigned.Add(10)
+	assert.Equal(t, "11", newsigned.String())
+	assert.Equal(t, "11", newunsigned.String())
+
+	assert.True(t, newsigned.GreaterThanOrEqual(signed))
+	assert.True(t, newunsigned.GreaterThanOrEqual(unsigned))
+
+	// Test for overflow on add operation.
```
jemiahw: Reading this for the first time, it would be helpful if the comment said something about expected behavior. e.g. "// Test that add operations do not overflow" or something like that.

pkg/table/datum_test.go
```diff
@@ -0,0 +1,46 @@
...
+	overflowSigned := newDatum(uint64(math.MaxInt64), signedType) // wrong type, converts.
+	overflowUnsigned := newDatum(uint64(math.MaxUint64), unsignedType)
```
jemiahw: Consider initializing these not to the max value, but to something like max - 10. That would catch errors like e.g. if add ignored the addition and returned the _current_ value or something like that (it doesn't, I looked, but still seems smart to cover it).

PR #116, opened 2023-05-01
check: add foreign key check

morgo:
Fixes https://github.com/squareup/spirit/issues/115

### DIFF
--- a/pkg/check/dropadd.go
+++ b/pkg/check/dropadd.go
@@ -1,0 +1,51 @@
+package check
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/pingcap/tidb/parser"
+	"github.com/pingcap/tidb/parser/ast"
+	"github.com/siddontang/loggers"
+)
+
+func init() {
+	registerCheck("dropadd", dropAddCheck, ScopePreflight)
+}
+
+// dropAddCheck checks for a DROP and then ADD in the same statement.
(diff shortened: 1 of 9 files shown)


pkg/check/foreignkey.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return fmt.Errorf("could not parse alter table statement: %s", sql)
+	}
+	stmt := &stmtNodes[0]
+	alterStmt, ok := (*stmt).(*ast.AlterTableStmt)
+	if !ok {
+		return errors.New("not a valid alter table statement")
+	}
+	for _, spec := range alterStmt.Specs {
+		if spec.Constraint != nil && spec.Constraint.Refer != nil {
```
joycse06, line 49: In the context of alter statments, what's the difference between `spec.Constraint` and `spec.NewConstraints`?

morgo: It looks like it is used here:
https://github.com/pingcap/tidb/blob/275696aa449bd7bff5c85ba3f8013d162dc690a4/parser/parser.y#L1982C4-L2000

I saw there were two different forms, so I added it for both.

morgo: In theory it should fire for this statement: 
```
ADD COLUMN foo INT references customers (id)
```

But that doesn't seem to be the case. This statement was parse-but-ignore in MySQL for a long time, so there might be some flag that needs to be enabled that I am missing.

PR #120, opened 2023-05-03
Add metrics interface

joycse06:
Fixes https://github.com/squareup/spirit/issues/118

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -44,21 +44,18 @@ type Sink interface {
 	Send(context.Context, *Metrics) error
 }
 
-// noopSink is the default sink which does nothing
-type noopSink struct{}
+// NoopSink is the default sink which does nothing
+type NoopSink struct{}
 
-func (s *noopSink) Send(ctx context.Context, m *Metrics) error {
+func (s *NoopSink) Send(ctx context.Context, m *Metrics) error {
 	return nil
 }
 
-var _ Sink = &noopSink{}
+var _ Sink = &NoopSink{}
 
-func NewNoopSink() *noopSink {
-	return &noopSink{}
+func NewNoopSink() *NoopSink {
+	return &NoopSink{}
 }
-
-// NoopSink is a Noop Sink instance to be used in tests
-var NoopSink = NewNoopSink()
 
 // logSink logs metrics
 type logSink struct {
--- a/pkg/row/copier.go
+++ b/pkg/row/copier.go
@@ -70,7 +70,7 @@ func NewCopierDefaultConfig() *CopierConfig {
 		FinalChecksum:   true,
 		Throttler:       &throttler.Noop{},
 		Logger:          logrus.New(),
-		MetricsSink:     metrics.NoopSink,
+		MetricsSink:     &metrics.NoopSink{},
 	}
 }
 
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -105,7 +105,7 @@ func NewRunner(m *Migration) (*Runner, error) {
 	r := &Runner{
 		migration:   m,
 		logger:      logrus.New(),
-		metricsSink: metrics.NoopSink,
+		metricsSink: &metrics.NoopSink{},
 	}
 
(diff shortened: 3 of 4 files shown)


pkg/row/copier.go
```diff
@@ -70,7 +73,7 @@ func NewCopierDefaultConfig() *CopierConfig {
 }
 
 // NewCopier creates a new copier object.
-func NewCopier(db *sql.DB, tbl, newTable *table.TableInfo, config *CopierConfig) (*Copier, error) {
+func NewCopier(db *sql.DB, tbl, newTable *table.TableInfo, config *CopierConfig, metricsSink metrics.Sink) (*Copier, error) {
```
morgo: I think metricsSink can be part of copierConfig. If it's `nil`, we use the noop sink implicitly.

joycse06: Yeah, I have moved it to `copierConfig`

pkg/row/copier_test.go
```diff
@@ -55,6 +71,9 @@ func TestCopier(t *testing.T) {
 	err = db.QueryRow("SELECT COUNT(*) FROM copiert2").Scan(&count)
 	assert.NoError(t, err)
 	assert.Equal(t, 1, count)
+
+	// Verify that testMetricsSink.Send was called 2 times
+	assert.Equal(t, 2, testMetricsSink.called)
```
joycse06, line 76: This is not that good of a test, but can't think of a good form of test as we are not doing much inside spirit. It's just providing a interface and plumbing.

morgo: Yeah, I think it's fine. Another option is to have the noop metrics sink retain the last metric sent, and then have a way to retrieve that.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,86 @@
...
+// Sink sends metrics to an external destination.
+type Sink interface {
+	// Send sends metrics to the sink. It must respect the context timeout, if any.
+	Send(context.Context, *Metrics) error
+}
+
+// noopSink is the default sink which does nothing
+type noopSink struct{}
+
+func (s noopSink) Send(ctx context.Context, m *Metrics) error {
```
morgo: Are you sure this isn't a situation where you want a pointer?
```suggestion
func (s *noopSink) Send(ctx context.Context, m *Metrics) error {
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+var _ Sink = noopSink{}
+
+func NewNoopSink() *noopSink {
+	return &noopSink{}
+}
+
+// NoopSink is a Noop Sink instance to be used in tests
+var NoopSink = NewNoopSink()
```
morgo: Was your motivation in having the helper to _not_ export the `noopSink` implementation? I think it's fine to export, since the net symbol count should be one less.

I think we can remove these lines, and just have the migration code do:
```
sink: &metrics.NoopSink{}
```

pkg/metrics/metrics.go
joycse06: Converted to a pointer. Thought this wasn't doing anything with the receiver.

pkg/row/copier_test.go
```diff
@@ -8,10 +8,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/squareup/spirit/pkg/metrics"
+
```
morgo, line 12: I think your imports look broken to include these above the others? but if go-fmt doesn't complain then I won't either.

pkg/metrics/metrics.go
joycse06: > Was your motivation in having the helper to not export the noopSink implementation

Yeah kind of but I am not too attached to that idea. Will update the use the struct instead.

pkg/row/copier_test.go
joycse06: I ran `golangci-lint run` locally and it was asking me to `goimport`-ed, so I added a file watcher in `goland` to run it automatically on files I touches. `golangci-lint` didn't have any complaints after that.

PR #127, opened 2023-05-07
fix resume from checkpoint

morgo:
Fixes https://github.com/squareup/spirit/issues/126
Fixes https://github.com/squareup/spirit/issues/125

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -91,8 +91,9 @@ type Runner struct {
 
 	// Used by the test-suite and some post-migration output.
 	// Indicates if certain optimizations applied.
-	usedInstantDDL bool
-	usedInplaceDDL bool
+	usedInstantDDL           bool
+	usedInplaceDDL           bool
+	usedResumeFromCheckpoint bool
 
 	// Attached logger
 	logger loggers.Advanced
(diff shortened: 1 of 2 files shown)


pkg/table/tableinfo_test.go
```diff
@@ -162,15 +162,15 @@ func TestLowWatermark(t *testing.T) {
 	chunker.Feedback(chunk, time.Second)
 	watermark, err := chunker.GetLowWatermark()
 	assert.NoError(t, err)
-	assert.Equal(t, "{\"Key\":\"id\",\"ChunkSize\":1000,\"LowerBound\":{\"Value\":1,\"Inclusive\":true},\"UpperBound\":{\"Value\":1001,\"Inclusive\":false}}", watermark)
+	assert.Equal(t, "{\"Key\":\"id\",\"ChunkSize\":1000,\"LowerBound\":{\"Value\": \"1\",\"Inclusive\":true},\"UpperBound\":{\"Value\": \"1001\",\"Inclusive\":false}}", watermark)
```
joycse06, line 165: maybe add a test with a large value which was causing the error?

morgo: Good point. Let me write a quick test for it.

PR #130, opened 2023-05-16
add support for prefetch into chunker

morgo:
Fixes https://github.com/squareup/spirit/issues/129

### DIFF
--- a/pkg/table/chunker_universal.go
+++ b/pkg/table/chunker_universal.go
@@ -44,7 +44,7 @@ var _ Chunker = &chunkerUniversal{}
 // nextChunkByPrefetching uses prefetching instead of feedback to determine the chunk size.
 // It is used when the chunker detects that there are very large gaps in the sequence.
 // When this mode is enabled, the chunkSize is "reset" to 1000 rows, so we know that
-// t.chunkSize is reliable. It is also expanded again based on feedback. Knowing when
+// t.chunkSize is reliable. It is also expanded again based on feedback.
 func (t *chunkerUniversal) nextChunkByPrefetching() (*Chunk, error) {
 	key := t.Ti.PrimaryKey[0]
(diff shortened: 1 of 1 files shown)


pkg/table/chunker_universal.go
```diff
@@ -31,11 +32,64 @@ type chunkerUniversal struct {
...
+
 	logger loggers.Advanced
 }
 
 var _ Chunker = &chunkerUniversal{}
 
+// nextChunkByPrefetching uses prefetching instead of feedback to determine the chunk size.
+// It is used when the chunker detects that there are very large gaps in the sequence.
+// When this mode is enabled, the chunkSize is "reset" to 1000 rows, so we know that
+// t.chunkSize is reliable. It is also expanded again based on feedback. Knowing when
```
joycse06: The last sentence in the comment seems incomplete?

pkg/table/chunker_universal.go
```diff
@@ -31,11 +32,64 @@ type chunkerUniversal struct {
...
+func (t *chunkerUniversal) nextChunkByPrefetching() (*Chunk, error) {
+	key := t.Ti.PrimaryKey[0]
+	query := fmt.Sprintf("SELECT %s FROM %s WHERE %s > ? ORDER BY %s LIMIT 1 OFFSET %d",
```
joycse06, line 50: Wondering, if we can set the `offset` to the `low_watermark`, I am not sure if that will result in any optimisation.

morgo: `OFFSET` means "skip n occurrences". So if we imagine the following ordered scenario:
```
1
2
4
6
7
8
9
10
```

If we said `WHERE id > 6 OFFSET 2` it would start at 7, and then move 2 places.. so 9. OFFSET is generally not efficient because it must count through the values exactly. If there are no gaps it's better to start at the nth already. But in the case we are using this, there are very large gaps. So we get an exact chunk from picking a strategic offset.

PR #131, opened 2023-05-24
runner.Run Should respect context cancellation

joycse06:
Fixes https://github.com/squareup/spirit/issues/69

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -114,6 +114,9 @@ func (c *Checker) checksumChunk(trxPool *dbconn.TrxPool, chunk *table.Chunk) err
 	return nil
 }
 func (c *Checker) RecentValue() string {
+	if c.recentValue == nil {
+		return "TBD"
+	}
 	return fmt.Sprintf("%v", c.recentValue)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/migration/runner_test.go
```diff
@@ -1268,3 +1268,104 @@ func TestChunkerPrefetching(t *testing.T) {
...
+		var rowCount int
+		assert.NoError(t, db.QueryRow(stmt).Scan(&rowCount))
+		if rowCount > 0 {
+			break
+		}
+	}
+	cancel()
+	// close repl client, throttler and replica
+	// we can't use runner.Close() as it will delete the checkpoint table as well, we want to test resuming
+	// TODO: Check if the following can be tied to context as well. So when context is cancelled they are cleaned up automatically.
```
morgo: I looked at the canal code, I don't think it's natively supported :( But we should be able to implement it in the wrapper and call Close() if we notice the context is canceled?

Please create a followup issue to track and then update the comment, then LGTM.

pkg/migration/runner_test.go
```diff
@@ -1322,20 +1321,11 @@ func TestResumeFromCheckpointE2E(t *testing.T) {
...
-		err := db.QueryRow(stmt).Scan(&table)
+		var rowCount int
+		err = db.QueryRow(`SELECT count(*) from _chkpresumetest_chkpnt`).Scan(&rowCount)
 		if err != nil {
-			if err == sql.ErrNoRows {
-				time.Sleep(10 * time.Millisecond)
-				continue
-			}
-			assert.NoError(t, err)
+			continue // table does not exist yet
```
joycse06, line 1327: 👍  Much cleaner.

PR #132, opened 2023-05-29
handle checksum better after type conversion

morgo:
Fixes https://github.com/squareup/spirit/issues/133

### DIFF
--- a/pkg/table/utils.go
+++ b/pkg/table/utils.go
@@ -25,26 +25,37 @@ func castableTp(tp string) string {
 	newTp := removeWidth(tp)
 	newTp = removeEnumSetOpts(newTp)
 	newTp = removeZerofill(newTp)
+	newTp = removeDecimalWidth(newTp)
 	switch newTp {
 	case "tinyint", "smallint", "mediumint", "int", "bigint":
 		return "signed"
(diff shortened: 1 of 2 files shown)


pkg/table/utils.go
```diff
@@ -14,3 +16,49 @@ func lazyFindP90(a []time.Duration) time.Duration {
...
+	case "timestamp":
+		return "datetime"
+	case "varchar", "enum", "set", "text", "mediumtext", "longtext":
+		return "char"
+	case "tinyblob", "blob", "mediumblob", "longblob", "varbinary":
+		return "binary"
+	case "float", "double": // required for MySQL 5.7
+		return "char"
+	default:
+		return removeWidth(tp) // char, binary, datetime, year, float, double, json,
```
morgo: On review, I think maybe we should change this to char. All types will eventually be converted to char in the `concat()` anyway, and this will better handle the case that I've missed a type that is not cast-able.

I'm going to push a code change now.

joycse06: I hope this wont mask data corruption during copy. I think it won't and we have pretty good test coverage.  So should be good.

morgo: It should still capture data truncation, as this is handled here https://github.com/squareup/spirit/blob/082c2d7b483b434a3de90ba5c1982e76d23f9355/pkg/dbconn/dbconn.go#L96-L121

PR #136, opened 2023-06-05
Create new chunker for composite key

YinanLiang:
Part 1 of fix https://github.com/squareup/spirit/issues/48

Use a different chunker for composite key. The new chunker currently does exactly the same thing as the universal chunker.

### DIFF
--- a/pkg/migration/runner_test.go
+++ b/pkg/migration/runner_test.go
@@ -230,7 +230,6 @@ func TestOnline(t *testing.T) {
 }
 
 func TestTableLength(t *testing.T) {
-	t.Skip("Not sure why, fails for now.")
 	runSQL(t, `DROP TABLE IF EXISTS thisisareallylongtablenamethisisareallylongtablename60charac`)
 	table := `CREATE TABLE thisisareallylongtablenamethisisareallylongtablename60charac (
 		id int(11) NOT NULL AUTO_INCREMENT,
@@ -263,7 +262,7 @@ func TestTableLength(t *testing.T) {
 		Password: cfg.Passwd,
 		Database: cfg.DBName,
 		Threads:  16,
-		Table:    "thisisareallylongtablenamethisisareallylongtablename64characters",
+		Table:    "thisisareallylongtablenamethisisareallylongtablename60charac",
 		Alter:    "ENGINE=InnoDB",
 	})
 	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


pkg/table/chunker_composite.go
```diff
@@ -0,0 +1,444 @@
...
+	// particularly for the test suite.
+	DisableDynamicChunker bool // optimization to adjust chunk size.
+
+	// This is used for restore.
+	watermark             *Chunk
+	watermarkQueuedChunks []*Chunk
+
+	// The chunk prefetching algorithm is used when the chunker detects
+	// that there are very large gaps in the sequence.
+	chunkPrefetchingEnabled bool
```
morgo:
```suggestion
```

pkg/table/chunker_composite.go
```diff
@@ -0,0 +1,444 @@
...
+
+	minVal := t.chunkPtr
+	maxVal := t.chunkPtr.Add(incrementSize)
+	t.chunkPtr = maxVal
+	return &Chunk{
+		ChunkSize:  t.chunkSize,
+		Key:        key,
+		LowerBound: &Boundary{minVal, true},
+		UpperBound: &Boundary{maxVal, false},
+	}, nil
```
morgo:
```suggestion
	return t.nextChunk()
```

pkg/table/chunker_composite.go
```diff
@@ -0,0 +1,444 @@
...
+	return uint64(math.Ceil(float64(increment) / divideBy)) // ceil to guarantee at least 1 for progress.
+}
+
+func (t *chunkerComposite) logicalRange() uint64 {
+	if !t.Ti.maxValue.IsNil() && !t.Ti.minValue.IsNil() {
+		return t.Ti.maxValue.Range(t.Ti.minValue)
+	}
+	// For binary types, we can't really do anything useful yet:
+	return math.MaxUint64
+}
```
morgo:
```suggestion
```

pkg/table/chunker_composite.go
```diff
@@ -0,0 +1,444 @@
...
+	// - This new target wants to go higher
+	// - our current p90 is only a fraction of our target time
+	if t.chunkSize == MaxDynamicRowSize && newTargetRows > MaxDynamicRowSize && (p90 < targetTime*5) {
+		t.logger.Warnf("dynamic chunking is not working as expected: target-time=%s p90-time=%s new-target-rows=%d max-dynamic-row-size=%d",
+			time.Duration(targetTime), time.Duration(p90), uint64(newTargetRows), MaxDynamicRowSize,
+		)
+		t.logger.Warn("switching to prefetch algorithm")
+		t.chunkSize = StartingChunkSize // reset
+		t.chunkPrefetchingEnabled = true
+	}
```
morgo:
```suggestion
```

YinanLiang: That makes sense. Thanks Morgan. I was planning to remove the dynamic chunking logics in the next PR to make sure this PR doesn't change the current behaviour before we introduce more tests. But thinking about it again it makes sense to start with a minimum chunker. Removed those logic as suggested :D

pkg/migration/runner_test.go
```diff
@@ -263,7 +262,7 @@ func TestTableLength(t *testing.T) {
 		Password: cfg.Passwd,
 		Database: cfg.DBName,
 		Threads:  16,
-		Table:    "thisisareallylongtablenamethisisareallylongtablename64characters",
+		Table:    "thisisareallylongtablenamethisisareallylongtablename60charac",
```
YinanLiang, line 265: I think this is the fix yeah?

morgo: In MySQL it's fixed at 4 chars, but since we need a _new table it requires an additional 5 chars.

PR #144, opened 2023-06-22
fix bugs, improve composite chunker

morgo:
The composite chunker was missing tests. After I looked at it, I saw that it didn't work correctly on varbinary keys. I've fixed it and added some tests (more needed).

I also added tests and improved handling for varbinary keys that need escaping.

(This relies on specific values from auto-inc being generated. I can fix it if 5.7 tests fail.)


pkg/checksum/checker_test.go
```diff
@@ -63,7 +63,7 @@ func TestBasicChecksum(t *testing.T) {
 	assert.Nil(t, checker.recentValue)
 	assert.Equal(t, "TBD", checker.RecentValue())
 	assert.NoError(t, checker.Run(context.Background()))
-	assert.Equal(t, "1", checker.RecentValue())
```
prudhvi, line 66: How was this passing earlier?

morgo: Previously the composite chunker had an initial minimum value that showed up and registered here. Now if there was no previous chunk pointer and its reached the end of the table, there's special handling to just return a chunk with no min/max value. I'll point it out.

pkg/table/chunker_composite.go
```diff
@@ -36,34 +36,53 @@ var _ Chunker = &chunkerComposite{}
...
 	// on the final chunk.
 	t.finalChunkSent = true
+	// If the chunkPtr is nil, it means this is our first
+	// and our last chunk (special case for small tables)
+	if t.chunkPtr.IsNil() {
+		return &Chunk{
+			ChunkSize: t.chunkSize,
+			Key:       key,
+		}, nil
+	}
```
morgo, line 85: @prudhvi this case here ^ This fixes a bug where:
- rows.Next() returns empty (there are less than chunkSize rows to the end of the table).
- There was no previous chunk sent out either.

prudhvi: Got it, unlike optimistic chunker there is no lowerbound (min value to begin with) for very first chunk so better to return `1=1` for `chunk.String()` which returns the whole table (smaller chunk) for copying. Good catch

PR #146, opened 2023-06-23
Add additional tests, small code cleanup

morgo:
This is some code cleanup post https://github.com/squareup/spirit/pull/144 and more tests (that PR was large enough)

It is function neutral

### DIFF
--- a/pkg/migration/runner_test.go
+++ b/pkg/migration/runner_test.go
@@ -1805,3 +1805,46 @@ func TestE2ERogueValues(t *testing.T) {
 	assert.Equal(t, "postChecksum", m.getCurrentState().String())
 	// All done!
 }
+
+func TestPartitionedTable(t *testing.T) {
+	runSQL(t, `DROP TABLE IF EXISTS part1, _part1_new`)
+	table := `CREATE TABLE part1 (
+			id bigint(20) NOT NULL AUTO_INCREMENT,
+			partition_id smallint(6) NOT NULL,
+			created_at timestamp(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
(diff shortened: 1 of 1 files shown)


pkg/table/chunker_composite_test.go
```diff
@@ -148,3 +149,151 @@ func TestCompositeChunkerInt(t *testing.T) {
...
+	assert.Error(t, chunker.Open()) // double open should fail
+
+	_, err = chunker.GetLowWatermark()
+	assert.Error(t, err)
+
+	assert.Equal(t, StartingChunkSize, int(chunker.chunkSize))
+	assert.True(t, chunker.KeyAboveHighWatermark(1000)) // when ptr is nil everything is above.
+	chunk, err := chunker.Next()
+	assert.NoError(t, err)
+	assert.Equal(t, "`pk` < 1008", chunk.String()) // first chunk
```
prudhvi, line 189: Why is the first chunk at 1008 with default chunksize of 1000?

prudhvi: i would have expected `pk < 1001` but i see my local mysql test that if `offset < 10` it's giving me `offset + 1` but for offset >=10 it's giving me `offset+8`

morgo: The reason why the optimistic chunker is reliable for these offsets is because it generates the numbers-client side. But in the pessimistic chunker we use the actual values from MySQL which are only guaranteed to be monotonic, not necessarily consecutive (aka `insert.. select` is not precise and loses some values).

This is what I meant by the comment from the previous PR:
> (This relies on specific values from auto-inc being generated. I can fix it if 5.7 tests fail.)

But it seems to be stable between 5.7 and 8.0 in a single thread. If it proves different in the future, we'll have to modify tests.

PR #149, opened 2023-06-27
extend number of retries to 10, cutovers to 100, fix prefetch switch

morgo:
This fixes two issues:

- Using 5 retries seems like it's not always enough since we set the MDL timeout quite low (5s) and only extend it by 1s each retry. This is carefully balanced, and backoff and retry is better than just trying to use an unlimited lock, since it will cause downtime. With 10 retries this brings the max MDL lock to 15 -- which is pretty close to downtime for most systems. After that it will fail cutover, and users can manually resume. In future, we might solve this with a `--force` flag (or similar) for those that have to push their schema changes over the line and this can be added by users on resume.  For now; I've also increased the number of cutover retries to 100.

- The logic in switching to prefetching was not totally correct. This caused it to flap back and forth between prefetching, which reduces chunker performance. It does not cause any service violations or incorrectness.

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -165,6 +165,13 @@ func (c *Checker) Run(ctx context.Context) error {
 	if err := c.feed.Flush(ctx); err != nil {
 		return err
 	}
+
+	// Assert that the change set is empty. This should always
+	// be the case because we are under a lock.
+	if c.feed.GetDeltaLen() > 0 {
+		return errors.New("the changeset is not empty, can not start checksum")
+	}
+
 	// Create a set of connections which can be used to checksum
 	// The table. They MUST be created before the lock is released
 	// with REPEATABLE-READ and a consistent snapshot (or dummy read)
@@ -173,6 +180,13 @@ func (c *Checker) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	// Assert that the change set is still empty.
+	// It should always be empty while we are still under the lock.
+	if c.feed.GetDeltaLen() > 0 {
+		return errors.New("the changeset is not empty, can not run checksum")
+	}
+
 	// We can now unlock the table before starting the checksumming.
 	if err = serverLock.Close(); err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


pkg/checksum/checker.go
```diff
@@ -165,6 +165,13 @@ func (c *Checker) Run(ctx context.Context) error {
 	if err := c.feed.Flush(ctx); err != nil {
 		return err
 	}
+
+	// Assert that the change set is empty. This should always
+	// be the case because we are under a lock.
+	if c.feed.GetDeltaLen() > 0 {
+		return errors.New("the changeset is not empty, can not start checksum")
+	}
```
morgo, line 173: This is a pedantic check. I've not been able to prove anything.

pkg/dbconn/dbconn.go
```diff
@@ -114,6 +115,12 @@ RETRYLOOP:
 				// might exist in the table and needs to be copied.
 				if code == "1062" && ignoreDupKeyWarnings {
 					continue // ignore duplicate key warnings
+				} else if code == "3170" {
+					// ER_CAPACITY_EXCEEDED
+					// "Memory capacity of 8388608 bytes for 'range_optimizer_max_mem_size' exceeded.
+					// Range optimization was not done for this query."
+					// i.e. the query still executes it just doesn't optimize perfectly
+					continue
```
morgo, line 123: I noticed this with a composite key and MySQL 8.0. AFAIK it was added post-GA 5.7 so doesn't apply to Aurora. This error is completely safe to ignore.

pkg/table/chunker_optimistic.go
```diff
@@ -384,7 +384,7 @@ func (t *chunkerOptimistic) calculateNewTargetChunkSize() uint64 {
 	// - We are already at the max chunk size
 	// - This new target wants to go higher
 	// - our current p90 is only a fraction of our target time
-	if t.chunkSize == MaxDynamicRowSize && newTargetRows > MaxDynamicRowSize && (p90 < targetTime*5) {
+	if t.chunkSize == MaxDynamicRowSize && newTargetRows > MaxDynamicRowSize && (p90*5 < targetTime) {
```
joycse06, line 387: Such a subtle thing, how did you find this?

morgo: Customer case. I'll link you from slack.

PR #162, opened 2023-07-18
Simpler fix for cutover race

morgo:
This reverts some of the complexity added in https://github.com/squareup/spirit/pull/156

The original cutover algorithm was correct with one exception: there is a race in the go-routine not starting the `RENAME` statement before the `UNLOCK TABLES` runs. This is very obvious in hindsight by inspecting this code:

https://github.com/squareup/spirit/blob/d4a5130063bbc6a737018aef9d8e2cf7e4d51546/pkg/migration/cutover.go#L100-L114

This only works because spirit only locks the original table. If it locks multiple tables (not required) then it would need to do the sentry-table trick that gh-ost uses. This has been verified by inspecting MDL locks in performance_schema.

(Hopefully this goes away and we can all start using MySQL 8.0 soon, where the cut-over is much simpler.)

Also fixes https://github.com/squareup/spirit/issues/163

### DIFF
--- a/pkg/migration/cutover.go
+++ b/pkg/migration/cutover.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/siddontang/loggers"
@@ -13,7 +14,6 @@ import (
 	"github.com/squareup/spirit/pkg/repl"
 	"github.com/squareup/spirit/pkg/table"
 	"github.com/squareup/spirit/pkg/utils"
-	"golang.org/x/sync/errgroup"
 )
 
 type CutoverAlgorithm int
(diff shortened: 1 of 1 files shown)


pkg/migration/cutover.go
```diff
@@ -223,66 +162,41 @@ func (c *CutOver) algorithmGhost(ctx context.Context) error {
...
-	// They are not known to return errors, but we check them anyway in case
-	// a change of logic is introduced.
-	if !c.feed.AllChangesFlushed() {
-		return errors.New("not all changes flushed, final flush might be broken")
-	}
-	if c.feed.GetDeltaLen() > 0 {
-		return fmt.Errorf("the changeset is not empty (%d), can not start cutover", c.feed.GetDeltaLen())
-	}
+		_, renameErr = trx.ExecContext(ctx, query)
+		wg.Done()
```
daniel-nichter, line 174: `defer wg.Done()` to ensure it's called, usually as first line in goroutine so it sits line-by-line with the `Add(1)` just above (outside the goroutine).

pkg/migration/runner.go
```diff
@@ -241,7 +241,9 @@ func (r *Runner) Run(originalCtx context.Context) error {
 	// in a racey way. If there was a delete inserted in that
 	// race, it will report an error incorrectly.
 	if err := r.postCutoverCheck(ctx); err != nil {
-		return err
+		// Don't return the error because our automation
+		// will retry the migration (but it's already happened)
+		r.logger.Errorf("post-cutover check failed: %v", err)
```
daniel-nichter, line 246: This still valid? Our automation might retry, but is valid for the open source code?

morgo: I don't have an easy way to separate the two. I think because the operation has already happened, it's reasonable in open source that users will want an `OK` back. They may make the same assumptions we are.

PR #165, opened 2023-07-20
Remove 5.7 support unless hidden flag is used

morgo:
Currently spirit supports MySQL 5.7, but with 5.7 officially reaching end of life this year, and Aurora-5.7 officially reaching end of life next year - it's dubious to release spirit with support for it. The rationale is that even though it works, it could create a bad experience where support is removed rather quickly. And we want to remove support, because some of the code that is 5.7 is non-stellar; i.e. the cutover algorithm.

This hides 5.7 behind a hidden flag, so it can still be used while we need it. But the docs only say 8.0 is supported, and we will plan on removing all the code early 2024 as planned.

### DIFF
--- a/pkg/migration/migration.go
+++ b/pkg/migration/migration.go
@@ -22,7 +22,7 @@ type Migration struct {
 	ReplicaDSN        string        `name:"replica-dsn" help:"A DSN for a replica which (if specified) will be used for lag checking." optional:""`
 	ReplicaMaxLag     time.Duration `name:"replica-max-lag" help:"The maximum lag allowed on the replica before the migration throttles." optional:"" default:"120s"`
 	LockWaitTimeout   time.Duration `name:"lock-wait-timeout" help:"The DDL lock_wait_timeout required for checksum and cutover" optional:"" default:"30s"`
-	SkipPreRunChecks  bool          `name:"skip-pre-run-checks" hidden:"" default:"false"`
+	SkipPreRunChecks  bool          `name:"i-understand-mysql57-is-not-supported" hidden:"" default:"false"`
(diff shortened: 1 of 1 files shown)


pkg/migration/migration.go
```diff
@@ -20,6 +22,7 @@ type Migration struct {
 	ReplicaDSN        string        `name:"replica-dsn" help:"A DSN for a replica which (if specified) will be used for lag checking." optional:""`
 	ReplicaMaxLag     time.Duration `name:"replica-max-lag" help:"The maximum lag allowed on the replica before the migration throttles." optional:"" default:"120s"`
 	LockWaitTimeout   time.Duration `name:"lock-wait-timeout" help:"The DDL lock_wait_timeout required for checksum and cutover" optional:"" default:"30s"`
+	SkipPreRunChecks  bool          `name:"skip-pre-run-checks" hidden:"" default:"false"`
```
daniel-nichter: This name is opaque; it could apply to anything. It's also not clear (based on its name) if I should turn it on or off for 5.7 and/or 8.0?

More explicit would be something like `--mysql57`, i.e. if user has 5.7 and want Spirit to work on 5.7, then they have to specify `--mysql57`. Also, when that flag is given, then it could (and probably should) make Spirit print a BIG banner that says "This is going to be removed!"

daniel-nichter: Maybe even painfully explicit: `--mysql57-is-not-supported` or something like that.

morgo: As discussed, I've renamed it to `--i-understand-mysql57-is-not-supported`.

PR #167, opened 2023-07-31
remove errors dependency

morgo:
This simplifies the can-retry to be included in the dbconn package instead of relying on go-mysql/errors.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.19
 require (
 	github.com/alecthomas/kong v0.7.1
 	github.com/go-mysql-org/go-mysql v1.6.1-0.20230116021724-c4a17aa5b49f
-	github.com/go-mysql/errors v0.0.0-20180603193453-03314bea68e0
 	github.com/go-sql-driver/mysql v1.7.0
 	github.com/pingcap/tidb/parser v0.0.0-20230122061550-1e0956d5ba41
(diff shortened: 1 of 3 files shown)


pkg/dbconn/dbconn.go
```diff
@@ -62,6 +66,25 @@ func standardizeTrx(ctx context.Context, trx *sql.Tx, config *DBConfig) error {
...
+// This is because it gets complicated in cases where the statement could
+// succeed but then there is a deadlock later on.
+func canRetryError(err error) bool {
+	var errNumber uint16
+	if val, ok := err.(*mysql.MySQLError); ok {
+		errNumber = val.Number
+	}
+	switch errNumber {
+	case errLockWaitTimeout, errDeadlock, errCannotConnect,
+		errConnLost, errReadOnly, errQueryKilled:
```
daniel-nichter, line 81: Will the other code work when the conn is lost/can't connect? If it does a _full_ retry of whatever it's doing, then it should be ok? It'll happen on a new conn, though, possibly without previously held locks?

morgo: Yes. The assumption (held previously and maintained here) is it's a _full retry_. See the comment just above:

> For simplicity a "retryable" error means
> rollback the transaction and start the transaction again.

Basically what counts as non retry-able is likely to be a parse error, or a duplicate key error.

PR #168, opened 2023-08-03
Chunk on all parts of the key

morgo:
Fixes https://github.com/squareup/spirit/issues/48

The composite chunker can now chunk on all parts of the key. This doesn't use row-value constructors because I noticed they don't perform as expected. So it expands the condition to be an optimizer-safe pattern.

### DIFF
--- a/pkg/table/chunker_composite.go
+++ b/pkg/table/chunker_composite.go
@@ -67,7 +67,7 @@ func (t *chunkerComposite) Next() (*Chunk, error) {
 			t.chunkSize,
 		)
 	}
-	upperDatums, err := t.nextQuery2Datums(query)
+	upperDatums, err := t.nextQueryToDatums(query)
 	if err != nil {
 		return nil, err
 	}
@@ -108,9 +108,9 @@ func (t *chunkerComposite) isFirstChunk() bool {
 	return len(t.chunkPtrs) == 0
 }
 
-// nextQuery2Datums executes the prefetch query which returns 1 row-max.
+// nextQueryToDatums executes the prefetch query which returns 1 row-max.
 // The columns in this result are then converted to Datums and returned
-func (t *chunkerComposite) nextQuery2Datums(query string) ([]Datum, error) {
+func (t *chunkerComposite) nextQueryToDatums(query string) ([]Datum, error) {
 	rows, err := t.Ti.db.Query(query)
 	if err != nil {
 		return nil, err
--- a/pkg/table/chunker_composite_test.go
+++ b/pkg/table/chunker_composite_test.go
@@ -12,6 +12,107 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestCompositeChunkerCompositeBinary(t *testing.T) {
+	runSQL(t, "DROP TABLE IF EXISTS composite_binary_t1")
+	runSQL(t, `CREATE TABLE composite_binary_t1 (
+		a varbinary(40) NOT NULL,
+		b varbinary(40) NOT NULL,
+		c int NOT NULL,
+		PRIMARY KEY (a,b)
+	)`)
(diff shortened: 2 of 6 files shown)


pkg/table/chunker_composite.go
```diff
@@ -45,65 +48,105 @@ func (t *chunkerComposite) Next() (*Chunk, error) {
...
 	}, nil
 }
 
+func (t *chunkerComposite) isFirstChunk() bool {
+	return len(t.chunkPtrs) == 0
+}
+
+// nextQuery2Datums executes the prefetch query which returns 1 row-max.
+// The columns in this result are then converted to Datums and returned
+func (t *chunkerComposite) nextQuery2Datums(query string) ([]Datum, error) {
```
prudhvi: what does `2` in this name signify?

pkg/table/chunker_composite_test.go
```diff
@@ -46,27 +46,27 @@ func TestCompositeChunkerBinary(t *testing.T) {
 	chunk, err := chunker.Next()
```
prudhvi, line 147: Can we have a test with primary key with more than 1 column?

pkg/table/chunker_composite.go
morgo: It converts "next query to datums":
- Input is a query to find the next values
- They are returned as datums.

pkg/table/chunker_composite_test.go
morgo: There are some integration level tests for this (pkg/migration/runner_test.go:TestE2EBinlogSubscribingCompositeKey, TestE2ERogueValues) but no unit test. Request makes sense, I will add one.

pkg/table/utils.go
```diff
@@ -98,3 +99,44 @@ func mysqlRealEscapeString(value string) string {
 	}
 	return sb.String()
 }
+
+// expandRowConstructorComparison is a workaround for MySQL
+// not always optimizing conditions such as (a,b,c) > (1,2,3).
+// This limitation is still current in 8.0, and was not fixed
+// by the work in https://dev.mysql.com/worklog/task/?id=7019
+func expandRowConstructorComparison(cols []string, operator Operator, vals []Datum) string {
```
prudhvi, line 107: Nice, I learnt about this today, haven't seen this in other programming languages on what (a,b) > (3,4) expands to unless you know exact behavior of how mysql treats that from https://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html

pkg/table/chunker_composite.go
prudhvi: Cool, any objections to naming it as `To` or is that a golang thing?

morgo: No objections, I'll change it.

pkg/table/tableinfo.go
```diff
@@ -277,3 +277,11 @@ func (t *TableInfo) WrapCastType(col string) string {
 	}
 	return fmt.Sprintf("CAST(`%s` AS %s)", col, castableTp(tp))
 }
```
prudhvi, line 280: In this file should `isCompatibleWithChunker` now check all key column types instead of first column to confirm that it's compatible?

morgo: Yes, you're right. Although it should be renamed from `isCompatibleWithChunker` to `isMemoryComparable` since with the composite chunker we no longer do the Key Above Watermark optimization, and use the composite chunker any time it's not a single-column auto-inc key.

We care about MemoryComparable for the Delta map, so I'll need to fix this. It could create problems if the key has a collation.

pkg/table/chunker_composite_test.go
prudhvi: Cool, thanks for pointing those, makes sense.

PR #169, opened 2023-08-07
Move memcomparable checks outside of chunker

morgo:
Previously, the chunker would refuse to allow chunking on non-memcomparable keys. It did this because (1) it didn't know how to do key-above-watermark for a non-comparable key (2) the delta map feature won't work with non mem comparable.

Because (1) has ostensibly been solved by disabling this optimization, and (2) is not a requirement of the chunker, but a requirement of migration, this changes the chunker to freely permit chunking on such keys. It is up to the migration (package checks) to verify if the key is memory comparable if it depends on such functionality.

### DIFF
--- a/pkg/migration/utils.go
+++ b/pkg/migration/utils.go
@@ -26,6 +26,10 @@ func IsCompatible(ctx context.Context, migration *Migration) bool {
 	if err := m.table.SetInfo(ctx); err != nil {
 		return false
 	}
+	// check that it's memory comparable.
+	if err := m.table.PrimaryKeyIsMemoryComparable(); err != nil {
+		return false
+	}
 	// Check that we can get a chunker.
 	if _, err := table.NewChunker(m.table, m.migration.TargetChunkTime, m.logger); err != nil {
 		return false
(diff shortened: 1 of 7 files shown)


pkg/migration/utils.go
```diff
@@ -26,6 +26,10 @@ func IsCompatible(ctx context.Context, migration *Migration) bool {
 	if err := m.table.SetInfo(ctx); err != nil {
 		return false
 	}
+	// check that it's memory comparable.
+	if err := m.table.PrimaryKeyIsMemoryComparable(); err != nil {
+		return false
+	}
 	// Check that we can get a chunker.
 	if _, err := table.NewChunker(m.table, m.migration.TargetChunkTime, m.logger); err != nil {
 		return false
```
prudhvi, line 35: Curious, why can't we replace this method with RunChecks method which checks for all preflight conditions? I also don't see it being used outside of tests.

morgo: This func was designed so that from our tern integration we can test if a migration is spirit compatible. We are not currently using it, instead we always try tern first and then fall-through to gh-ost if it doesn't work.

You are right - we probably could change it to just run all pre-flight checks. I was just fixing it for now for tests, but I'll either remove it or improve it in future.

PR #170, opened 2023-08-07
Chunk on a non primary key

morgo:
This is a draft to demo chunking on a non primary key. Fixes https://github.com/squareup/spirit/issues/159

I believe eventually I would like to change it from:
```
SetKey(keyName string, keyCols []string, additionalConditions string)
```

To:
```
SetKey(keyName string, additionalConditions string)
```

Since the keyCols should be able to be auto-determined by merging the cols from this index and the primary key.

### DIFF
--- a/pkg/table/chunker_composite.go
+++ b/pkg/table/chunker_composite.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/siddontang/loggers"
+	"golang.org/x/exp/slices"
 )
 
 type chunkerComposite struct {
@@ -406,10 +407,12 @@ func (t *chunkerComposite) SetKey(keyName string, where string) error {
 	// This is not ideal for chunking, and since we are allowed to assume InnoDB, each
 	// secondary index actually includes the PRIMARY KEY columns in it.
 	// So we can merge in the PK columns to the keyCols.
-	// For simplicity, we only do this when the columns *don't* overlap.
+	// We only do this for each non-overlapping column in the PRIMARY KEY.
 	// This is because ranging on the same column twice will create logic errors.
-	if !keysOverlap(keyCols, t.Ti.KeyColumns) {
-		keyCols = append(keyCols, t.Ti.KeyColumns...)
+	for _, pkCol := range t.Ti.KeyColumns {
+		if !slices.Contains(keyCols, pkCol) {
+			keyCols = append(keyCols, pkCol)
+		}
 	}
 	t.chunkKeys = keyCols
 	t.keyName = keyName
--- a/pkg/table/chunker_composite_test.go
+++ b/pkg/table/chunker_composite_test.go
@@ -496,3 +496,56 @@ func TestSetKey(t *testing.T) {
 		assert.NoError(t, chunker.Close())
 	}
 }
+
+// TestSetKeyCompositeKeyMerge tests our expansion when columns in the index
+// are also in the PRIMARY KEY.
+// We shouldn't include the column twice due to logic errors,
+// but we can use the other columns from the primary key for chunking.
+// Proof:
+// explain format=json SELECT * FROM setkeycomposite_t1 FORCE INDEX (dnc) WHERE dob='2023-08-10' and name=0x63643361343961382D333739392D313165652D393166352D613562616235356361653536 AND city='cd3a49ac-3799-11ee-91f5-a5bab55cae56' AND ssn=0x63643361343961392D333739392D313165652D393166352D613562616235356361653536;
+//
+//	"key": "dnc",
+//
+// "used_key_parts": [
+//
+//		"dob",
+//		"name",
+//		"city",
+//		"ssn"
+//	  ],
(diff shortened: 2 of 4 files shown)


pkg/table/chunker_composite_test.go
```diff
@@ -394,3 +394,76 @@ func TestCompositeSmallTable(t *testing.T) {
...
+		Ti:            t1,
+		ChunkerTarget: 100 * time.Millisecond,
+		logger:        logrus.New(),
+	}
+	err = chunker.SetKey("status", []string{"status", "id"}, "status = 'PENDING' AND updated_at > NOW() - INTERVAL 1 DAY")
+	assert.NoError(t, err)
+	assert.NoError(t, chunker.Open())
+	chunk, err = chunker.Next()
+	assert.NoError(t, err)
+	assert.Equal(t, "((`status` < \"PENDING\")\n OR (`status` = \"PENDING\" AND `id` < 1008)) AND (status = 'PENDING' AND updated_at > NOW() - INTERVAL 1 DAY)", chunk.String())
```
prudhvi, line 452: Interesting the user supplied `where` clause will help avoid selection of wrong chunks irrespective of rowconstructor expansion here.

morgo: Yeah, I had originally planned min/max key. But we don't technically need it. We can use a free-form `where`.

pkg/migration/utils.go
```diff
@@ -1,38 +0,0 @@
...
-
-	"github.com/squareup/spirit/pkg/table"
-)
-
-// IsCompatible checks if a migration is compatible with this program.
-// This can be used by automation to fallback to gh-ost if it's not.
-func IsCompatible(ctx context.Context, migration *Migration) bool {
-	m, err := NewRunner(migration)
-	if err != nil {
-		return false
```
morgo, line 17: This is cleanup from https://github.com/squareup/spirit/pull/174

This is no longer useful. RENAMES remain the only unsupported feature, but since these may still work in Spirit because MySQL instant operations are tried first, it's difficult to gate that in a check here.

pkg/table/chunker_composite.go
```diff
@@ -364,3 +389,30 @@ func (t *chunkerComposite) calculateNewTargetChunkSize() uint64 {
...
+	// There is a chance that if the index is something like "status" then it is low cardinality.
+	// This is not ideal for chunking, and since we are allowed to assume InnoDB, each
+	// secondary index actually includes the PRIMARY KEY columns in it.
+	// So we can merge in the PK columns to the keyCols.
+	// For simplicity, we only do this when the columns *don't* overlap.
+	// This is because ranging on the same column twice will create logic errors.
+	if !keysOverlap(keyCols, t.Ti.KeyColumns) {
+		keyCols = append(keyCols, t.Ti.KeyColumns...)
+	}
+	t.chunkKeys = keyCols
```
prudhvi, line 417: I am assuming here that the `t.chunkKeys` need not be a unique index even if it's used in ORDER BY?
let's say PK is composite key (name, ssn) and the index we selected is (dateofbirth,name). There is a overlap(name) and we don't add the PK columns making it (dob, name, ssn) as per this. Is that ok?

morgo: It does *not* need to be unique.

If the PK is `(name,ssn)` and the index selected is `(dob, name)` the chunking will be on `(dob,name)` only. The only risk of it being non-unique is that the chunks are larger than intended:

1. The prefetch query attempts to find the `(dob,name)` that is 10K rows ahead of the current point.
2. If this `(dob,name)` is pair non-unique, it might cause the actual chunk query to copy 12K rows instead.
3. The next prefetch query will start after the previous key pair of `(dob,name)`, so no data is loss.

Really the biggest risk is every user has the same `(dob,name)`, since then logically everything can only be copied in one chunk. This is why the expansion to include PK exists - consider a scenario where the index is on `(status)`. If you are finding for `status='ARCHIVED'` then without the expansion this would just be one chunk.

prudhvi: Yeah, right we discussed this earlier, there is no data loss if it's not unique . Jut that it might be a bit inefficient. So can we then change this to add all non overlapping columns of PK to the end instead of bailing out if there is a overlap of one column?

Or may be it's a bit complicated to ensure it's the right order of the columns in PK if we append at end.
Eg: user provided index has `(dob,name,city)`  and if PK is `(name,ssn)` , then not sure if we make order by columns 
`(dob,name,city, ssn)` or `(dob,name,ssn,city)`?

morgo: I was just about to write to you why it's not possible.. but I think it is actually possible.

i.e.
(dob,name,city) is actually (dob,name,city,name,ssn) internally (InnoDB's clustered index).

My concern was that you can't "skip over" `name` to get to `ssn`, but that's not actually true. Name is already known and will be specified, so there is nothing to skip over.

Let's try it. If it doesn't work, we can revert this small optimization later.

PR #173, opened 2023-08-08
Use TLS for canal connection if host is RDS

morgo:


pkg/dbconn/conn.go
```diff
@@ -58,14 +70,8 @@ func newDSN(dsn string) (string, error) {
...
 	}
-	if isRDSHost(cfg.Addr) {
-		once.Do(func() {
-			caCertPool := x509.NewCertPool()
-			caCertPool.AppendCertsFromPEM(rds2019rootCA)
-			tlsConfig := &tls.Config{RootCAs: caCertPool}
-			err = mysql.RegisterTLSConfig(rdsTLSConfigName, tlsConfig)
-		})
-		if err != nil {
+	if IsRDSHost(cfg.Addr) {
```
prudhvi, line 73: Just a thought, if we move this `IsRDSHost` check inside `InitRDSTLS()` and have that init method called inside runner setup, we don't need to call init ever again for canal or something else?

morgo: I will fix it later when we add full TLS support: https://github.com/squareup/spirit/issues/175

PR #174, opened 2023-08-09
Add support for VARCHAR primary keys

morgo:
Fixes https://github.com/squareup/spirit/issues/160

It looks like MySQL 8.0 expanded its support for `INSTANT` somewhere between 8.0.26 and 8.0.28. I've bumped the tested version because the latest Aurora release is now based on 8.0.28.


pkg/repl/client.go
```diff
@@ -338,9 +354,79 @@ func (c *Client) FlushUnderLock(ctx context.Context, lock *dbconn.TableLock) err
...
+	// Early return if there is nothing to flush.
+	if len(changesToFlush) == 0 {
+		c.SetPos(posOfFlush)
+		return nil
+	}
+
+	// Otherwise, flush the changes.
+	var stmts []string
+	var buffer []string
+	prevKeyOperation := changesToFlush[0].isDelete // for initialization
```
joycse06: `prevKeyOperation` doesn't follow the same boolean naming pattern. Maybe change it to something like `isPrevKeyOperationDelete` or `isDeletePrevKeyOperation`?

morgo: I will just call it `prevKey` and retain the full key. I agree it reads better:
```
--- a/pkg/repl/client.go
+++ b/pkg/repl/client.go
@@ -385,12 +385,12 @@ func (c *Client) flushQueue(ctx context.Context, underLock bool, lock *dbconn.Ta
        // Otherwise, flush the changes.
        var stmts []string
        var buffer []string
-       prevKeyOperation := changesToFlush[0].isDelete // for initialization
+       prevKey := changesToFlush[0] // for initialization
        for _, change := range changesToFlush {
                // We are changing from DELETE to REPLACE
                // or vice versa, *or* the buffer is getting very large.
-               if change.isDelete != prevKeyOperation || len(buffer) > DefaultBatchSize {
-                       if prevKeyOperation {
+               if change.isDelete != prevKey.isDelete || len(buffer) > DefaultBatchSize {
+                       if prevKey.isDelete {
                                stmts = append(stmts, c.createDeleteStmt(buffer))
                        } else {
                                stmts = append(stmts, c.createReplaceStmt(buffer))
@@ -398,10 +398,10 @@ func (c *Client) flushQueue(ctx context.Context, underLock bool, lock *dbconn.Ta
                        buffer = nil // reset
                }
                buffer = append(buffer, change.key)
-               prevKeyOperation = change.isDelete
+               prevKey.isDelete = change.isDelete
        }
        // Flush the buffer once more.
-       if prevKeyOperation {
+       if prevKey.isDelete {
                stmts = append(stmts, c.createDeleteStmt(buffer))
        } else {
                stmts = append(stmts, c.createReplaceStmt(buffer))
```

PR #183, opened 2023-08-22
Prudhvi.watermark map

prudhvi:
Fixes
https://github.com/squareup/spirit/issues/182
https://github.com/squareup/spirit/issues/181

### DIFF
--- a/pkg/table/chunker_composite.go
+++ b/pkg/table/chunker_composite.go
@@ -316,3 +316,17 @@ func (t *chunkerComposite) SetKey(keyName string, where string) error {
 	t.where = where
 	return nil
 }
+
+// GetLowWatermark returns the highest known value that has been safely copied,
+// which (due to parallelism) could be significantly behind the high watermark.
+// The value is discovered via Chunker Feedback(), and when retrieved from this func
+// can be used to write a checkpoint for restoration.
+func (t *chunkerComposite) GetLowWatermark() (string, error) {
+	t.Lock()
+	defer t.Unlock()
+	if t.watermark == nil || t.watermark.UpperBound == nil || t.watermark.LowerBound == nil {
+		return "", errors.New("watermark not yet ready")
+	}
+
+	return t.watermark.JSON(), nil
+}
--- a/pkg/table/core_chunker.go
+++ b/pkg/table/core_chunker.go
@@ -1,7 +1,6 @@
 package table
 
 import (
-	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -31,6 +30,13 @@ type coreChunker struct {
 	isOpen          bool
 	logger          loggers.Advanced
 }
+
+type restorableChunker interface {
+	isSpecialRestoredChunk(chunk *Chunk) bool
+	bumpWatermark(chunk *Chunk)
+}
+
+var _ restorableChunker = &coreChunker{}
 
 // isSpecialRestoredChunk is used to test for the first chunk after restore-from-checkpoint.
 // The restored chunk is a really special beast because the lowerbound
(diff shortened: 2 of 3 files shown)


pkg/table/chunk.go
```diff
@@ -99,6 +95,16 @@ func (b *Boundary) comparesTo(b2 *Boundary) bool {
 	return true
 }
 
+// ValuesString uses only values and ignores inclusive operation,
+// to get a string representation of Values in Boundary.
+func (b *Boundary) ValuesString() string {
```
morgo: If it's `valuesString()` I don't think we need to export it. But there is also no existing `String()` method for a *Boundary. You could consider adding it as `String()` since it would be more go idiomatic. The only downside is that it won't include `Inclusive` in the output.

```suggestion
// valuesString uses only values and ignores inclusive operation,
// to get a string representation of Values in Boundary.
func (b *Boundary) valuesString() string {
```

pkg/table/chunker_composite.go
```diff
@@ -4,36 +4,27 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"golang.org/x/exp/slices"
```
morgo: Causing a linter error:
```
pkg/table/chunker_composite.go:7: File is not `goimports`-ed (goimports)
```

pkg/table/chunker_composite.go
```diff
@@ -4,36 +4,27 @@ import (
...
-	Ti             *TableInfo
-	chunkSize      uint64
-	chunkPtrs      []Datum  // a list of Ptrs for each of the keys.
-	chunkKeys      []string // all the keys to chunk on (usually all the col names of the PK)
-	keyName        string   // the name of the key we are chunking on
-	where          string   // any additional WHERE conditions.
-	finalChunkSent bool
-	isOpen         bool
+	*restorableChunker
+	chunkSize uint64
```
morgo: Since you embed `restorableChunker` into both, what do you think about moving the other shared fields over? I can see:

```
chunkSize
chunkTimingInfo
Ti 
finalChunkSent 
isOpen   
```

pkg/table/restorable_chunker.go
```diff
@@ -0,0 +1,99 @@
...
+
+// isSpecialRestoredChunk is used to test for the first chunk after restore-from-checkpoint.
+// The restored chunk is a really special beast because the lowerbound
+// will be repeated by the first chunk that is applied post restore.
+// This is called under a mutex.
+func (t *restorableChunker) isSpecialRestoredChunk(chunk *Chunk) bool {
+	if chunk.LowerBound == nil || chunk.UpperBound == nil || t.watermark == nil || t.watermark.LowerBound == nil || t.watermark.UpperBound == nil {
+		return false // restored checkpoints always have both.
+	}
+	return chunk.LowerBound.Value[0] == t.watermark.LowerBound.Value[0]
```
morgo: I think this might cause a bug? It's only considering the first part of the key, versus in the composite chunker it uses a helper to compare the full key:

```
chunk.LowerBound.comparesTo(t.watermark.LowerBound)
```

prudhvi: Good catch, copied from wrong chunker. I was playing a lot with what methods to move this common one, initially wanted to move the dynamic chunking parts also , but implementation for `calculateNewTargetChunkSize` was different between chunkers and it cascades up to methods like `Feedback` etc , so couldn't pull chunking parts and only pulled restorable (watermark related) parts.

pkg/table/chunk.go
prudhvi: Thanks originally tried `String()` only and wrote specs too but realized that Boundary has a `JSON()` method which considers inclusive operation too, so didn't want String version to behave differently and made this naming explicit.  I will make it private as per suggestion.

pkg/table/chunker_composite.go
prudhvi: Sure, what about `ChunkerTarget`?

pkg/table/core_chunker.go
```diff
@@ -0,0 +1,114 @@
...
+		t.watermark = chunk
+		goto applyStoredChunks
+	}
+
+	// Validate that chunk has lower bound before moving on
+	if chunk.LowerBound == nil {
+		errMsg := fmt.Sprintf("coreChunker.bumpWatermark: nil value encountered: %v", chunk)
+		t.logger.Error(errMsg)
+		panic(errMsg)
+	}
```
prudhvi, line 78: This is the validation I moved from `applyChunks` part to beginning (before storing the chunk). I removed the upperBound check as it's redundant with first check in this method.

pkg/table/chunker_composite.go
morgo: Yes, I missed it. Please move this one too.

PR #185, opened 2023-08-24
Expose composite chunker

prudhvi:
Fixes
https://github.com/squareup/spirit/issues/184

### DIFF
--- a/pkg/table/chunker.go
+++ b/pkg/table/chunker.go
@@ -55,12 +55,23 @@ func NewChunker(t *TableInfo, chunkerTarget time.Duration, logger loggers.Advanc
 			},
 		}, nil
 	}
-	return &chunkerComposite{
+	return NewCompositeChunker(t, chunkerTarget, logger, "", "")
+}
+
+// NewCompositeChunker returns a chunkerComposite ,
+// setting its Key if keyName and where conditions are provided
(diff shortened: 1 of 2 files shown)


pkg/table/chunker.go
```diff
@@ -58,3 +58,12 @@ func NewChunker(t *TableInfo, chunkerTarget time.Duration, logger loggers.Advanc
 		logger:        logger,
 	}, nil
 }
+
```
morgo, line 60: For the linter... 
```suggestion
// NewCompositeChunker
```

pkg/table/chunker.go
```diff
@@ -55,12 +55,23 @@ func NewChunker(t *TableInfo, chunkerTarget time.Duration, logger loggers.Advanc
...
 			lowerBoundWatermarkMap: make(map[string]*Chunk, 0),
 			logger:                 logger,
 		},
-	}, nil
+	}
+	var err error
+	if keyName != "" && whereCondition != "" {
+		err = compositeChunker.SetKey(keyName, whereCondition)
+	}
+	return &compositeChunker, err
```
morgo: I think in this instance you can call it `c := chunkerComposite{}` since it's easy to follow within the func. But LGTM with these changes if you want to merge.

PR #191, opened 2023-09-05
migration: cleanup status output

morgo:
Fixes https://github.com/squareup/spirit/issues/190

The status output is changed to every 30s. Needless precision from time-values is also removed, rounding to the nearest second. i.e.

```
- INFO[0051] apply complete: instant-ddl=false inplace-ddl=false total-chunks=289 copy-rows-time=41.499063041s checksum-time=9.977830834s total-time=51.691053625s
+ INFO[0051] apply complete: instant-ddl=false inplace-ddl=false total-chunks=289 copy-rows-time=41s checksum-time=10s total-time=52s
```

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -42,7 +42,7 @@ const (
 var (
 	checkpointDumpInterval  = 50 * time.Second
 	tableStatUpdateInterval = 5 * time.Minute
-	statusInterval          = 10 * time.Second
+	statusInterval          = 30 * time.Second
 	// binlogPerodicFlushInterval is the time that the client will flush all binlog changes to disk.
 	// Longer values require more memory, but permit more merging.
 	// I expect we will change this to 1hr-24hr in the future.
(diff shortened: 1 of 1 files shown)


pkg/migration/runner.go
```diff
@@ -867,29 +867,29 @@ func (r *Runner) dumpStatus(ctx context.Context) {
...
 				)
 			case stateApplyChangeset, statePostChecksum:
 				// We've finished copying rows, and we are now trying to reduce the number of binlog deltas before
 				// proceeding to the checksum and then the final cutover.
-				r.logger.Infof("migration status: state=%s binlog-deltas=%v time-total=%v",
+				r.logger.Infof("migration status: state=%s binlog-deltas=%v total-time=%s",
 					r.getCurrentState().String(),
 					r.replClient.GetDeltaLen(),
-					time.Since(r.startTime).String(),
+					time.Since(r.startTime).Round(time.Second),
```
prudhvi, line 881: Just a thought if rounding total time to minutes instead of seconds is more readable , but I think we will lose the seconds precision?

morgo: The only part that is really unreadable is fractional seconds IMHO. The go duration formatting on a long running task will change from `41h10m11.939962388s` to `41h10m12s`... which should be fine.

PR #192, opened 2023-09-05
checksum, migration: Continue to flush binlog during long running checksum

morgo:
Fixes https://github.com/squareup/spirit/issues/189

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -21,8 +21,6 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/checksum/checker.go
```diff
@@ -21,6 +21,8 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var binlogFlushIntervalDuringChecksum = 30 * time.Second
```
prudhvi: Nit: May be this can be moved as constant to repl/client as a defaultFlushInterval

PR #193, opened 2023-09-05
Implement custom connection pool

prudhvi:
Fixes https://github.com/squareup/spirit/issues/180

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -84,7 +84,7 @@ func (c *Checker) ChecksumChunk(ctx context.Context, chunk *table.Chunk) error {
 	if err != nil {
 		return err
 	}
-	defer c.ssPool.Put(conn)
+	defer c.ssPool.Put(ctx, conn)
 	c.logger.Debugf("checksumming chunk: %s", chunk.String())
 	source := fmt.Sprintf("SELECT BIT_XOR(CRC32(CONCAT(%s))) as checksum FROM %s WHERE %s",
 		c.intersectColumns(),
--- a/pkg/dbconn/connpool.go
+++ b/pkg/dbconn/connpool.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"sync"
 
 	"github.com/siddontang/loggers"
 	"github.com/squareup/spirit/pkg/table"
@@ -14,7 +13,7 @@ import (
 
 type ConnPool struct {
 	sem    *semaphore.Weighted
-	m      sync.Mutex
+	m      *semaphore.Weighted
 	db     *sql.DB
 	config *DBConfig
 	conns  []*sql.Conn
@@ -26,6 +25,7 @@ type ConnPool struct {
 // has been opened.
 func NewPoolWithConsistentSnapshot(ctx context.Context, db *sql.DB, count int, config *DBConfig, logger loggers.Advanced) (*ConnPool, error) {
 	sem := semaphore.NewWeighted(int64(count))
+	m := semaphore.NewWeighted(1)
 	rrConns := make([]*sql.Conn, 0, count)
 	for i := 0; i < count; i++ {
 		conn, err := db.Conn(ctx)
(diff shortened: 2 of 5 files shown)


pkg/dbconn/connpool.go
```diff
@@ -0,0 +1,188 @@
...
+type ConnPool struct {
+	sync.Mutex
+	config *DBConfig
+	conns  []*sql.Conn
+}
+
+// NewRRConnPool creates a pool of transactions which have already
+// had their read-view created in REPEATABLE READ isolation.
+func NewRRConnPool(ctx context.Context, db *sql.DB, count int, config *DBConfig) (*ConnPool, error) {
+	checksumTxns := make([]*sql.Conn, 0, count)
```
morgo: We can rename this, since it is checksum agnostic.

pkg/dbconn/connpool.go
```diff
@@ -0,0 +1,188 @@
...
+		return
+	}
+	p.Lock()
+	defer p.Unlock()
+	p.conns = append(p.conns, trx)
+}
+
+// Close closes all transactions in the pool.
+func (p *ConnPool) Close() error {
+	fmt.Println("ABout to close all connections ####")
```
morgo: I don't mind if you want to Debug log this. To do so though, you would need to plumb a logger through in dbConfig.

pkg/dbconn/dbconn.go
```diff
@@ -35,6 +33,37 @@ func NewDBConfig() *DBConfig {
...
+	_, err = conn.ExecContext(ctx, "SET lock_wait_timeout=?", config.LockWaitTimeout)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 func standardizeTrx(ctx context.Context, trx *sql.Tx, config *DBConfig) error {
 	_, err := trx.ExecContext(ctx, "SET time_zone='+00:00'")
 	if err != nil {
```
morgo: I think (actually I hope) this func can also be removed? We now only have to standardize connections

pkg/dbconn/dbconn_test.go
```diff
@@ -4,12 +4,9 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"github.com/stretchr/testify/assert"
```
morgo: I think this is wrong, and probably is triggering a linter. Non built-in imports go last.

pkg/checksum/checker.go
```diff
@@ -28,7 +28,8 @@ type Checker struct {
 	concurrency int
 	feed        *repl.Client
 	db          *sql.DB
-	trxPool     *dbconn.TrxPool
+	regularPool *dbconn.ConnPool // non snapshot connection
+	pool        *dbconn.ConnPool // RR snapshot connections
```
morgo: I think we should maybe flip these around and keep the regular pool as `pool` so it's the same in all components? You can also remove `db` and use `pool.DB()`.

pkg/row/copier.go
```diff
@@ -97,7 +96,7 @@ func NewCopier(db *sql.DB, tbl, newTable *table.TableInfo, config *CopierConfig)
 }
 
 // NewCopierFromCheckpoint creates a new copier object, from a checkpoint (copyRowsAt, copyRows)
-func NewCopierFromCheckpoint(db *sql.DB, tbl, newTable *table.TableInfo, config *CopierConfig, lowWatermark string, rowsCopied uint64, rowsCopiedLogical uint64) (*Copier, error) {
+func NewCopierFromCheckpoint(db *dbconn.ConnPool, tbl, newTable *table.TableInfo, config *CopierConfig, lowWatermark string, rowsCopied uint64, rowsCopiedLogical uint64) (*Copier, error) {
 	c, err := NewCopier(db, tbl, newTable, config)
```
morgo:
```suggestion
func NewCopierFromCheckpoint(pool *dbconn.ConnPool, tbl, newTable *table.TableInfo, config *CopierConfig, lowWatermark string, rowsCopied uint64, rowsCopiedLogical uint64) (*Copier, error) {
	c, err := NewCopier(pool, tbl, newTable, config)
```

pkg/migration/runner.go
```diff
@@ -534,19 +539,20 @@ func (r *Runner) postCutoverCheck(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	checker, err := checksum.NewChecker(r.db, oldTable, cutoverTable, r.replClient, &checksum.CheckerConfig{
+	ssPoolForChecksum, err := dbconn.NewPoolWithConsistentSnapshot(ctx, r.db, 1, r.pool.DBConfig(), r.logger)
+	if err != nil {
+		return err
+	}
```
prudhvi, line 546: ssPool currently only gets initialized in checker.Run method so have to add option in CheckerConfig and supply from runner here for this step cc @morgo

PR #199, opened 2023-09-20
migration, table: disable maxValue updating after copier finishes

morgo:
Fixes https://github.com/squareup/spirit/issues/198


pkg/migration/runner.go
```diff
@@ -424,6 +442,9 @@ func (r *Runner) setup(ctx context.Context) error {
 	// Start routines in table and replication packages to
 	// Continuously update the min/max and estimated rows
 	// and to flush the binary log position periodically.
+	// These will both be stopped when the copier finishes
+	// and checkpoint starts, although the PeriodicFlush
```
prudhvi: and checksum* starts?

morgo: Yup 🤦 thx!

PR #203, opened 2023-09-25
fix recovery in varchar composite keys

morgo:
Fixes https://github.com/squareup/spirit/issues/202

### DIFF
--- a/pkg/table/chunk.go
+++ b/pkg/table/chunk.go
@@ -100,10 +100,13 @@ func (b *Boundary) comparesTo(b2 *Boundary) bool {
 func (b *Boundary) valuesString() string {
 	vals := make([]string, len(b.Value))
 	for i, v := range b.Value {
-		if v.Tp == signedType || v.Tp == unsignedType {
+		// We have to *also* quote numeric types to avoid JSON float behavior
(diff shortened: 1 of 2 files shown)


pkg/table/chunk.go
```diff
@@ -100,7 +100,11 @@ func (b *Boundary) comparesTo(b2 *Boundary) bool {
 func (b *Boundary) valuesString() string {
 	vals := make([]string, len(b.Value))
 	for i, v := range b.Value {
-		vals[i] = fmt.Sprintf(`"%s"`, v)
+		if v.Tp == signedType || v.Tp == unsignedType {
+			vals[i] = fmt.Sprintf(`"%s"`, v)
```
daniel-nichter, line 107: This if-else block is a little confusing because aren't all (numeric) types signed or unsigned? Then below the comment "// already quoted" is confusing too because the `if` checks aren't for quoted or not quoted. My guess is that this is really something like,
```go
if !quotedValue {
    // not quoted so add "" around value
} else {
    // already quoted
}
```
?

morgo: Yeah that's right. I will make it easier to read.

PR #205, opened 2023-09-27
Enable -race in testsuite

morgo:
Fixes https://github.com/squareup/spirit/issues/204

### DIFF
--- a/pkg/throttler/mysql57replica.go
+++ b/pkg/throttler/mysql57replica.go
@@ -4,13 +4,15 @@ import (
 	"database/sql"
 	"errors"
 	"strconv"
+	"sync"
 	"sync/atomic"
 	"time"
 )
 
 type MySQL57Replica struct {
+	sync.Mutex
 	Repl
-	isClosed bool
+	isClosed atomic.Bool
 }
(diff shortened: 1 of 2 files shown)


pkg/migration/runner_test.go
```diff
@@ -1514,12 +1514,6 @@ func TestTpConversion(t *testing.T) {
 }
 
 func TestResumeFromCheckpointE2E(t *testing.T) {
-	// Lower the checkpoint interval for testing.
-	checkpointDumpInterval = 100 * time.Millisecond
-	defer func() {
-		checkpointDumpInterval = 50 * time.Second
-	}()
-
```
prudhvi, line 1522: why was this not needed with --race option?

morgo: It's a race because there's multiple tests that change it or read it. I changed it in TestMain() so that it lowers the test interval and applies to all tests in the suite

prudhvi: Ah thanks, TIL about that, I guess we could reset the value back to original after m.Run() there but not necessary

PR #207, opened 2023-10-02
Move standardize connection to connection param

morgo:

Fixes https://github.com/squareup/spirit/issues/180

This uses an alternative implementation to proposed in the issue:
- It moves the standardize() functions to the driver (I've not used this prior; but tested and it works fine).
- It implements a maximum pool size for the pool of concurrency +1.

### DIFF
--- a/pkg/dbconn/conn.go
+++ b/pkg/dbconn/conn.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 
 	"github.com/go-sql-driver/mysql"
+	"github.com/squareup/spirit/pkg/utils"
 )
 
 const (
@@ -107,14 +108,28 @@ func newDSN(dsn string, config *DBConfig) (string, error) {
 	return dsn, nil
 }
 
-func New(dsn string, config *DBConfig) (*sql.DB, error) {
-	dsn, err := newDSN(dsn, config)
+// New is similar to sql.Open except we take the inputDSN and
+// append additional options to it to standardize the connection.
+// It will also ping the connection to ensure it is valid.
+func New(inputDSN string, config *DBConfig) (*sql.DB, error) {
+	dsn, err := newDSN(inputDSN, config)
 	if err != nil {
 		return nil, err
 	}
 	db, err := sql.Open("mysql", dsn)
 	if err != nil {
 		return nil, err
+	}
+	if err := db.Ping(); err != nil {
+		utils.ErrInErr(db.Close())
+		if config.Aurora20Compatible {
+			return nil, err // Already tried it, didn't work.
+		}
+		// This could be because of transaction_isolation vs. tx_isolation.
+		// Try changing to Aurora 2.0 compatible mode and retrying.
+		// because config is a pointer, it will update future calls too.
+		config.Aurora20Compatible = true
+		return New(inputDSN, config)
 	}
 	db.SetMaxOpenConns(config.MaxOpenConnections)
 	return db, nil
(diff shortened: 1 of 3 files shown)


pkg/dbconn/conn.go
```diff
@@ -76,15 +80,37 @@ func newDSN(dsn string) (string, error) {
...
+	// we need to be able to produce the same when copying.
+	// If you look at standard packages like wordpress, drupal etc.
+	// they all change the SQL mode. If you look at mysqldump, etc.
+	// they all unset the SQL mode just like this.
+	ops = append(ops, fmt.Sprintf("%s=%s", "sql_mode", url.QueryEscape(`""`)))
+	ops = append(ops, fmt.Sprintf("%s=%s", "time_zone", url.QueryEscape(`"+00:00"`)))
+	ops = append(ops, fmt.Sprintf("%s=%s", "innodb_lock_wait_timeout", url.QueryEscape(fmt.Sprint(config.InnodbLockWaitTimeout))))
+	ops = append(ops, fmt.Sprintf("%s=%s", "lock_wait_timeout", url.QueryEscape(fmt.Sprint(config.LockWaitTimeout))))
+	// go driver options, should set:
+	// character_set_client, character_set_connection, character_set_results
```
prudhvi, line 104: Are these 3 extra settings referring to more to do in future or just refer to the two options below?

pkg/migration/runner.go
```diff
@@ -153,15 +153,23 @@ func (r *Runner) Run(originalCtx context.Context) error {
...
+	r.dbConfig = dbconn.NewDBConfig()
+	r.dbConfig.LockWaitTimeout = int(r.migration.LockWaitTimeout.Seconds())
+	// The copier and checker will use Threads to limit N tasks concurrently,
+	// but we also set it at the DB pool level with +1. Because the copier and
+	// the replication applier use the same pool, it allows for some natural throttling
+	// of the copier if the replication applier is lagging. Because it's +1 it
+	// means that the replication applier can always make progress immediately,
+	// and does not need to wait for free slots from the copier *until* it needs
+	// copy in more than 1 thread.
+	r.dbConfig.MaxOpenConnections = r.migration.Threads + 1
```
prudhvi, line 165: Nice this solves the starvation issue

pkg/repl/client.go
```diff
@@ -478,12 +477,12 @@ func (c *Client) flushMap(ctx context.Context, underLock bool, lock *dbconn.Tabl
...
 		}
 	} else {
-		// Execute the statements in parallel up to 16 threads.
+		// Execute the statements in parallel
 		// They should not conflict and order should not matter
 		// because they come from a consistent view of a map,
 		// which is distinct keys.
 		g, errGrpCtx := errgroup.WithContext(ctx)
-		g.SetLimit(flushThreads)
+		g.SetLimit(c.concurrency)
```
prudhvi, line 485: May be have to remember to adjust the config in tern when we roll this out or bump the default migration threads in this PR?

pkg/dbconn/conn.go
morgo: When you set charset + collation at the driver level, it runs the statement:
```
set names binary collate binary;
```

This statement is a shortcut to update the following variables: character_set_client, character_set_connection, character_set_results.

pkg/repl/client.go
morgo: As we discussed, I think it's fine. It never previously went as high as 16 threads in any of our workloads.

PR #212, opened 2023-10-10
Re-copy chunks when a checksum fails

morgo:
Fixes https://github.com/squareup/spirit/issues/213

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -213,6 +213,12 @@ SELECT source.row_checksum as source_row_checksum, target.row_checksum as target
 	return nil // managed to inspect differences
 }
 
+// replaceChunk recopies the data from table to newTable for a given chunk.
+// Note that the chunk is dynamically sized based on the target-time that it took
+// to *read* data in the checksum. This could be substantially longer than the time
+// that it takes to copy the data. Maybe in future we could consider splitting
+// the chunk here, but this is expected to be a very rare situation, so a small
+// stall from an XL sized chunk is considered acceptable.
 func (c *Checker) replaceChunk(chunk *table.Chunk) error {
 	c.logger.Warnf("recopying chunk: %s", chunk.String())
 	deleteStmt := "DELETE FROM " + c.newTable.QuotedName + " WHERE " + chunk.String()
--- a/USAGE.md
+++ b/USAGE.md
@@ -103,10 +103,12 @@ When set to `TRUE`, Spirit will attempt to perform the schema change using MySQL
 
 When set to `TRUE`, Spirit will perform a checksum of the data in the table after the copy phase. This is a good way to ensure that the copy phase was successful, but it does add some overhead to the process. When you resume-from-checkpoint, Spirit will only run with the checksum enabled (regardless of your configuration). This is because it can not rely on duplicate-key errors to detect issues in the copy phase if the DDL included adding a new `UNIQUE` key.
 
-The checksum typically adds about 10-20% of additional time to the migration, but it is recommended to always leave it enabled. A failed checksum will cause Spirit to exit with a non-zero exit code, and means that there is either:
+The checksum typically adds about 10-20% of additional time to the migration, but it is recommended to always leave it enabled. A failed checksum means that there is either:
 - A bug in Spirit
 - A bug in MySQL
 - Hardware errors
+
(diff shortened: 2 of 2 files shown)


pkg/checksum/noop_resolver.go
```diff
@@ -0,0 +1,102 @@
...
+		} else if !sourceRowChecksum.Valid && targetRowChecksum.Valid {
+			r.logger.Warnf("row does not exist in source for pk: %s", pk)
+		} else {
+			r.logger.Warnf("row checksum mismatch for pk: %s: source %s != target %s", pk, sourceRowChecksum.String, targetRowChecksum.String)
+		}
+	}
+	return errors.New("checksum mismatch")
+}
+
+func (r *NoopResolver) intersectColumns() string {
```
prudhvi: Can't use existing utils intersectColumns method in this package?

pkg/checksum/checker.go
```diff
@@ -23,26 +24,29 @@ import (
...
+	db                 *sql.DB
+	trxPool            *dbconn.TrxPool
+	isInvalid          bool
+	chunker            table.Chunker
+	startTime          time.Time
+	ExecTime           time.Duration
+	recentValue        interface{} // used for status
+	dbConfig           *dbconn.DBConfig
+	logger             loggers.Advanced
+	resolver           Resolver
```
prudhvi: Can we take a array of resolvers, I am thinking it might be useful to always send NoOpResolver even when using ReplaceResolver so that we know which rows caused the issue

pkg/migration/runner.go
```diff
@@ -804,23 +805,37 @@ func (r *Runner) resumeFromCheckpoint(ctx context.Context) error {
...
+			// If it doesn't, we have a resolver.
+			return err
+		}
+		// If we are here, the checksum passed.
+		// But we don't know if a resolver was invoked.
+		// We want to know it passed without one.
+		if !r.checker.ResolverWasInvoked() {
+			break // success!
+		}
+		r.logger.Errorf("checksum required a resolver to pass. retrying.")
```
prudhvi: may be also log the attempt number, I assume we will still continue  the migration with checksumpassed even if all the 3 times it invoked a resolver?

pkg/checksum/replace_resolver.go
```diff
@@ -0,0 +1,50 @@
...
+	r.db = db
+	r.source = source
+	r.target = target
+	r.logger = logger
+	return nil
+}
+
+// Resolve deletes all the data in the target and recopies it from the source.
+// It should mean that all the data is now consistent, but we retry the checksum
+// again if it failed.
```
prudhvi: I assume you will add the logic here for retrying checksum again after replacing?

pkg/migration/runner.go
```diff
@@ -804,30 +805,42 @@ func (r *Runner) resumeFromCheckpoint(ctx context.Context) error {
...
+			// If it doesn't, we have a resolver.
+			return err
+		}
+		// If we are here, the checksum passed.
+		// But we don't know if differences were found and chunks were recopied.
+		// We want to know it passed without one.
+		if r.checker.DifferencesFound() == 0 {
+			break // success!
+		}
+		r.logger.Errorf("The checksum failed process failed. This likely indicates either a bug in Spirit, or manual modification to the _new table outside of Spirit. This error is not fatal; the chunks of data that mismatched have been recopied. The checksum process will be repeated until it completes without any errors. Retrying %d/%d times", i+1, 3)
```
prudhvi, line 840: IIUC, if we find differences in 3rd/last run, we will proceed as success without checksuming again right? Don't we want to fail ?

morgo: Yep, you are right :blush: PTAL again ;-)

pkg/migration/runner.go
```diff
@@ -804,30 +805,45 @@ func (r *Runner) resumeFromCheckpoint(ctx context.Context) error {
...
+			// If it doesn't, we have a resolver.
+			return err
+		}
+		// If we are here, the checksum passed.
+		// But we don't know if differences were found and chunks were recopied.
+		// We want to know it passed without one.
+		if r.checker.DifferencesFound() == 0 {
+			break // success!
+		}
+		if i >= 3 {
```
prudhvi: off by one? >= 2?

PR #216, opened 2023-10-13
Update statistics once right away and then in a loop after every inte…

prudhvi:


prudhvi:
Closing this as this change in functionality isn't needed by the dependent system.

PR #227, opened 2023-10-27
Change pingcap parser dependency location

morgo:

Fixes https://github.com/cashapp/spirit/issues/226 and https://github.com/cashapp/spirit/issues/229

This also previously attempted noving to using the PingCAP sql escape library, but I'll try that in another PR


morgo:
There's something wrong with the golang-ci check. It seems to be evaluating against an older commit in this branch. I'm going to close this PR and try creating a new one.

PR #235, opened 2023-10-31
Add support for RBR minimal row image

morgo:

Fixes https://github.com/cashapp/spirit/issues/221

(DRAFT first, checking that it breaks as expected on the pre-check, then altering the pre-check).


morgo:
Tests are failing, which means there is some work to do. I'm going to close this PR for now. Anyone interested in this work, feel free to use this as a starting point.

PR #236, opened 2023-11-01
fix test breakages

morgo:

Due to an commit a couple back (introducing a faulty TestMain) the migration package tests were not running. Thankfully only tests broke, no bugs were introduced.

### DIFF
--- a/pkg/migration/cutover.go
+++ b/pkg/migration/cutover.go
@@ -168,6 +168,7 @@ func (c *CutOver) algorithmGhost(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	defer trx.Rollback() //nolint: errcheck
 	// Start the rename operation, it's OK it will block inside
 	// of this go-routine.
 	var wg sync.WaitGroup
--- a/pkg/migration/cutover.go
+++ b/pkg/migration/cutover.go
@@ -168,6 +168,9 @@ func (c *CutOver) algorithmGhost(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	// We need to do either commit or rollback to free the connection
+	// Since the only statement was RENAME (DDL) rollback
+	// has no effect. We can just defer it to ensure it's freed.
 	defer trx.Rollback() //nolint: errcheck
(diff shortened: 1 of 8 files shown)


pkg/migration/cutover.go
```diff
@@ -168,6 +168,7 @@ func (c *CutOver) algorithmGhost(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	defer trx.Rollback() //nolint: errcheck
```
morgo, line 174: This is a leak I caught.

pkg/migration/migration_test.go
```diff
@@ -15,6 +17,7 @@ import (
 
 func TestMain(m *testing.M) {
 	checkpointDumpInterval = 100 * time.Millisecond
+	os.Exit(m.Run())
```
morgo, line 21: This was the fatal mistake. This line was missing.

prudhvi: wow , silent killer, good catch.

pkg/migration/cutover.go
prudhvi: Shouldn't you rollback only incase of error?

pkg/migration/migration_test.go
prudhvi: Should we reset the values back to original after the run() method if we want other package tests to use the original values?

prudhvi (approved):
LGTM, except have a question about `trx.rollback()`

pkg/migration/cutover.go
morgo: I will add a comment to make this clearer.

There is no `trx.Close()` (go idiomatic) you have to chose to commit or rollback. This connection is used for `RENAME` which is DDL, and is non-transactional. So either commit or rollback has the same effect. I chose rollback.

pkg/migration/migration_test.go
morgo: The other test packages are run concurrently, so they will have to live with these shorter values anyway.

PR #238, opened 2023-11-08
fix parser driver error

morgo:

Fixes https://github.com/cashapp/spirit/issues/237


prudhvi (approved):
Nice build workflow to prevent future issues.

Issue #244, opened 2023-11-16
Delay cutover feature


morgo:
I would like to propose that we add a new feature to 'delay cutover'. This is currently supported by gh-ost, but not by Spirit.

For the implementation I propose that:
- We create an additional table called the "cutover signal" (or something like that; TBD)
- When copy and checksum phases have finished, spirit blocks waiting for the cutover signal to be dropped.
- As soon as it is no longer present, the cutover phase is initiated.

The idea being that this will allow users to schedule their cutover (which requires an MDL lock) to an opportune time. Technically the checksum also requires an MDL lock, so we may need further discussion on that, but this is the basic idea.

kolbe:
I am working on this in https://github.com/kolbe/spirit/tree/defer-cutover

I'm using the term "deferred cutover", as "delay" to me sounds like it would have a specific time associated with it.

I chose to call the signal table a "sentinel".

The current behavior is to check every 1 second for 48 hours and then exit with an error if the sentinel table has not been dropped. If spirit is then restarted, it should resume from checkpoint and quickly start waiting again for the sentinel to be dropped.

The primary missing piece of functionality right now is in handling a situation where spirit is killed or times out waiting for the sentinel table to be deleted, following which an operator might remove the sentinel table before running spirit again. At present, spirit will re-create the sentinel table (assuming the `--defer-cutover` flag is used again) and block again waiting for it to be deleted.

PR #245, opened 2023-11-21
Add support for deferred cutover

kolbe:
Implements #244 

The defer occurs before checksum. Documentation is included in this PR and makes it clear.

### DIFF
--- a/pkg/migration/migration.go
+++ b/pkg/migration/migration.go
@@ -24,7 +24,7 @@ type Migration struct {
 	LockWaitTimeout      time.Duration `name:"lock-wait-timeout" help:"The DDL lock_wait_timeout required for checksum and cutover" optional:"" default:"30s"`
 	SkipPreRunChecks     bool          `name:"i-understand-mysql57-is-not-supported" hidden:"" default:"false"`
 	SkipDropAfterCutover bool          `name:"skip-drop-after-cutover" help:"Keep old table after completing cutover" optional:"" default:"false"`
-	DeferCutOver         bool          `name:"defer-cutover" help:"Defer cutover until sentinel table is dropped" optional:"" default:"false"`
+	DeferCutOver         bool          `name:"defer-cutover" help:"Defer cutover (and checksum) until sentinel table is dropped" optional:"" default:"false"`
 }
 
 func (m *Migration) Run() error {


pkg/migration/runner.go
```diff
@@ -78,6 +80,7 @@ type Runner struct {
 	table           *table.TableInfo
 	newTable        *table.TableInfo
 	checkpointTable *table.TableInfo
+	sentinelTable   *table.TableInfo
```
morgo: Because sentinel is optional, I would prefer not to add it to the struct. We can construct sentinelName in the couple of places its needed.

pkg/migration/runner.go
```diff
@@ -219,6 +222,16 @@ func (r *Runner) Run(originalCtx context.Context) error {
...
 
+	// r.handleDeferredCutOver may return an error if there is
+	// some unexpected problem checking for the existence of
+	// the sentinel table OR if sentinelWaitLimit is exceeded.
+	// This function is invoked even if DeferCutOver is false
+	// because it's possible that the sentinel table was created
+	// manually after the migration started.
+	if err := r.handleDeferredCutOver(ctx); err != nil {
+		return err
+	}
```
morgo: I prefer the naming "waitOnSentinelTable" or something of that nature so it's clear it's a block wait.

pkg/migration/runner.go
```diff
@@ -407,6 +420,17 @@ func (r *Runner) setup(ctx context.Context) error {
 		if err := r.createCheckpointTable(ctx); err != nil {
 			return err
 		}
+
+		// The sentinelTable is "registered" separately from creating it.
+		// This helps to support the case where a user manually creates
+		// the sentinel table after a migration is started.
+		r.registerSentinelTable()
```
morgo: Because of the previous suggestion about not in the struct, we can remove register.
```suggestion
		// The sentinelTable is is created if the user wants to defer cutover.
		// If they don't want to, we don't need to create it.
```

pkg/migration/runner.go
```diff
@@ -646,6 +670,34 @@ func (r *Runner) createCheckpointTable(ctx context.Context) error {
...
+	if err := dbconn.Exec(ctx, r.db, stmt, r.sentinelTable.SchemaName, r.sentinelTable.TableName); err != nil {
+		return err
+	}
+	stmt = "INSERT INTO %n.%n (alter_statement) VALUES (%?)"
+	if err := dbconn.Exec(ctx, r.db, stmt, r.sentinelTable.SchemaName, r.sentinelTable.TableName, r.migration.Alter); err != nil {
+		return err
+	}
+	return nil
+}
+
```
morgo: * Remove register
* For the dbconn.Exec, I'm trying to use the statement directly as the string argument and not via a variable. This reduces some sql injection risk because we are no longer at risk of passing around raw statements (it can be linted on in future that dbconn.Exec does not use a variable as the 3rd arg).
* For the sentinel table, we don't strictly need the alter statement. We just need the existence of the table, which you can simplify the definition of. This is because you never check the statement matches, and we more easily support the use case of manually created sentinel tables post migration start.


```suggestion
func (r *Runner) createSentinelTable(ctx context.Context) error {
	sentinelName := fmt.Sprintf("_%s_sentinel", r.table.TableName)
	if err := dbconn.Exec(ctx, r.db, "DROP TABLE IF EXISTS %n.%n",
		r.table.SchemaName,
		sentinelName); err != nil {
		return err
	}
	return dbconn.Exec(ctx, r.db, "CREATE TABLE %n.%n (id int NOT NULL AUTO_INCREMENT PRIMARY KEY, created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP)",
		r.table.SchemaName,
		sentinelName,
	)
}
```

pkg/migration/runner.go
```diff
@@ -938,3 +990,53 @@ func (r *Runner) dumpStatus(ctx context.Context) {
...
+		r.registerSentinelTable()
+	}
+	sql := "SELECT COUNT(*) FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ?"
+	var sentinelTableExists int
+	err := r.db.QueryRowContext(ctx, sql, r.sentinelTable.SchemaName, r.sentinelTable.TableName).Scan(&sentinelTableExists)
+	if err != nil {
+		return false, err
+	}
+	return sentinelTableExists > 0, nil
+}
```
morgo:
```suggestion
func (r *Runner) sentinelTableExists(ctx context.Context) (bool, error) {
	sentinelName := fmt.Sprintf("_%s_sentinel", r.table.TableName)
	sql := "SELECT COUNT(*) FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ?"
	var sentinelTableExists int
	err := r.db.QueryRowContext(ctx, sql, r.table.SchemaName, sentinelName).Scan(&sentinelTableExists)
	if err != nil {
		return false, err
	}
	return sentinelTableExists > 0, nil
}
```

pkg/migration/runner.go
```diff
@@ -938,3 +990,53 @@ func (r *Runner) dumpStatus(ctx context.Context) {
...
+	}
+
+	r.logger.Warnf("cutover deferred while sentinel table %s exists; will wait %s", r.sentinelTable.QuotedName, sentinelWaitLimit)
+
+	timer := time.NewTimer(sentinelWaitLimit)
+
+	ticker := time.NewTicker(sentinelCheckInterval)
+	defer ticker.Stop()
+	logticker := time.NewTicker(10 * sentinelCheckInterval)
+	defer logticker.Stop()
```
morgo: As well as renaming this func (wait on sentinel ..) we should use the global logger here. Which means adding a new status:

	r.setCurrentState(stateWaitingOnSentinelTable)


And the logging should be performed by `r.dumpStatus()`. This might require a little bit of work since you'll want the timer to be part of the struct so you can state the remaining time.

i.e. in dumpStatus it should have something like this:

```
			case stateWaitingOnSentinelTable:
				// We are waiting for the sentinel table to be dropped.
				// Give some really useful status info on it.
				r.logger.Infof("migration status: state=%s binlog-deltas=%v total-time=%s time-waiting=XX maximum-wait-time=XX conns-in-use=%d",
					r.getCurrentState().String(),
					r.replClient.GetDeltaLen(),
					time.Since(r.startTime).Round(time.Second),
					r.db.Stats().InUse,
				)
```

pkg/migration/runner_test.go
```diff
@@ -2202,3 +2202,109 @@ func TestDropAfterCutover(t *testing.T) {
 	assert.Equal(t, tableCount, 0)
 	assert.NoError(t, m.Close())
 }
+
+func TestDeferCutOver(t *testing.T) {
+	tableName := `deferred_cutover`
+	newName := fmt.Sprintf("_%s_new", tableName)
+
+	testutils.RunSQL(t, fmt.Sprintf(`DROP TABLE IF EXISTS %s`, tableName))
+	table := fmt.Sprintf(`CREATE TABLE %s (id bigint unsigned not null auto_increment, primary key(id))`, tableName)
```
morgo: E2E test looks good. The thing I'm not sure of (and we should test!):

- Are the binary log coordinates still being advanced while we are block waiting (I would hope so)?
- Does it work correctly with resume from checkpoint?

kolbe: I added some additional tests. One of them checks that the binlog position still advances even while the migration is in stateWaitingOnSentinelTable. Another extends the Checkpoint E2E test to manually create a sentinel table after the first migration exists and then ensures that the new migration resumes from checkpoint and correctly blocks on the sentinel, even though both migrations were started with "DeferCutOver: false".

USAGE.md
```diff
@@ -135,3 +135,13 @@ The replication throttler only affects the copy-rows operation, and does not app
...
+
+If defer-cutover is true, Spirit will create a "sentinel" table in the same schema as the table being altered; the name of the sentinel table will use the pattern `_<table>_sentinel`. Spirit will never delete the sentinel table on its own. It will block for 48 hours waiting for the sentinel table to be dropped by the operator, after which it will exit with an error.
+
+You can resume a migration from checkpoint and Spirit will start waiting again for you to drop the sentinel table. You can also choose to delete the sentinel table before restarting Spirit, which will cause it to resume from checkpoint and complete the cutover without waiting, even if you have again enabled defer-cutover for the migration.
+
+If you start a migration and realize that you forgot to set defer-cutover, worry not! You can manually create a sentinel table using the pattern `_<table>_sentinel`, and Spirit will detect the table before the cutover is completed and block as though defer-cutover had been enabled from the beginning.
```
morgo, line 147: docs LGTM, but please make it clear where the waiting occurs. Currently it is just before checksum, which has a valid use (checksum does require a lock to get started), but I would have thought just before cutover was better (I'll comment more on that inline).

If we decide defer-cutover happens before checksum:
- Document that it could still be several hours after the sentry table is dropped that the cutover actually occurs, when checksum is enabled (it is by default).

If we decide it's after checksum:
- Document that a lock is still required for checksum (could cause a brief stall)
- Cutover should be close to instant once the sentinel table is removed.

pkg/migration/migration.go
```diff
@@ -24,6 +24,7 @@ type Migration struct {
 	LockWaitTimeout      time.Duration `name:"lock-wait-timeout" help:"The DDL lock_wait_timeout required for checksum and cutover" optional:"" default:"30s"`
 	SkipPreRunChecks     bool          `name:"i-understand-mysql57-is-not-supported" hidden:"" default:"false"`
 	SkipDropAfterCutover bool          `name:"skip-drop-after-cutover" help:"Keep old table after completing cutover" optional:"" default:"false"`
+	DeferCutOver         bool          `name:"defer-cutover" help:"Defer cutover until sentinel table is dropped" optional:"" default:"false"`
```
morgo: We should be able to insert a brief comment again if it is before/after checksum.

pkg/migration/runner.go
```diff
@@ -219,6 +225,18 @@ func (r *Runner) Run(originalCtx context.Context) error {
...
+	// some unexpected problem checking for the existence of
+	// the sentinel table OR if sentinelWaitLimit is exceeded.
+	// This function is invoked even if DeferCutOver is false
+	// because it's possible that the sentinel table was created
+	// manually after the migration started.
+	r.sentinelWaitStartTime = time.Now()
+	r.setCurrentState(stateWaitingOnSentinelTable)
+	if err := r.waitOnSentinelTable(ctx); err != nil {
+		return err
+	}
```
morgo, line 238: This is implying it occurs pre-checksum. If you decide that's the better implementation (I'm open to it), we should also re-order `stateWaitingOnSentinelTable` in the statuses list so they are ordinal.

If you do make this change, we should be careful that the repl client applier is stopped in `r.prepareForCutover(ctx);` just above this, so you'll have to re-enable it so we keep applying changes while we are in the blocking wait.

kolbe: I reordered the statuses

USAGE.md
kolbe: Yes I can see both of them, but the fact we wait up to 48h for the sentinel table to be dropped, then exit and require a checkpoint resume, means that a checksum taken before the wait would have very little validity. If many users want to have the other behavior or more fine-grained control over both checksum *and* cutover, we can reassess in the future.

morgo (approved):
LGTM. I changed the timeout to 48 seconds for manual testing:

```
 go run . --host "127.0.0.1:8030" --defer-cutover
INFO[0000] Starting spirit migration: concurrency=4 target-chunk-size=500ms table=test.stock alter="engine=innodb"
INFO[0000] create BinlogSyncer with config {ServerID:1076 Flavor:mysql Host:127.0.0.1 Port:8030 User:msandbox Password: Localhost: Charset:utf8 SemiSyncEnabled:false RawModeEnabled:false TLSConfig:<nil> ParseTime:false TimestampStringLocation:UTC UseDecimal:false RecvBufferSize:0 HeartbeatPeriod:0s ReadTimeout:0s MaxReconnectAttempts:0 DisableRetrySync:false VerifyChecksum:false DumpCommandFlag:0 Option:<nil> Logger:0x14000110e60 Dialer:0x100b58580 RowsEventDecodeFunc:0x100d364a0 TableMapOptionalMetaDecodeFunc:<nil> DiscardGTIDSet:false EventCacheCount:10240}
WARN[0000] resuming from checkpoint. low-watermark: {"Key":["s_w_id","s_i_id"],"ChunkSize":56767,"LowerBound":{"Value": ["50","23955"],"Inclusive":true},"UpperBound":{"Value": ["50","80723"],"Inclusive":false}} log-file: binlog.005904 log-pos: 27728 copy-rows: 4999991
INFO[0000] skip dump, use last binlog replication pos (binlog.005904, 27728) or GTID set <nil>
INFO[0000] begin to sync binlog from position (binlog.005904, 27728)
INFO[0000] start sync binlog at binlog file (binlog.005904, 27728)
INFO[0000] copy rows complete
WARN[0000] cutover deferred while sentinel table test._stock_sentinel exists; will wait 48s
INFO[0030] migration status: state=waitingOnSentinelTable sentinel-table=test._stock_sentinel total-time=30s sentinel-wait-time=30s sentinel-max-wait-time=48s conns-in-use=0
INFO[0030] finished periodic flush of binary log: duration=4.801542ms
INFO[0048] closing canal
INFO[0048] syncer is closing...
INFO[0048] kill last connection id 62
INFO[0048] syncer is closed
spirit: error: timed out waiting for sentinel table to be dropped
exit status 1
```

It resumes fine, and responds quickly when the sentinel is dropped:
```
go run . --host "127.0.0.1:8030" --defer-cutover
INFO[0000] Starting spirit migration: concurrency=4 target-chunk-size=500ms table=test.stock alter="engine=innodb"
INFO[0000] create BinlogSyncer with config {ServerID:1392 Flavor:mysql Host:127.0.0.1 Port:8030 User:msandbox Password: Localhost: Charset:utf8 SemiSyncEnabled:false RawModeEnabled:false TLSConfig:<nil> ParseTime:false TimestampStringLocation:UTC UseDecimal:false RecvBufferSize:0 HeartbeatPeriod:0s ReadTimeout:0s MaxReconnectAttempts:0 DisableRetrySync:false VerifyChecksum:false DumpCommandFlag:0 Option:<nil> Logger:0x14000404b90 Dialer:0x100df0580 RowsEventDecodeFunc:0x100fce4a0 TableMapOptionalMetaDecodeFunc:<nil> DiscardGTIDSet:false EventCacheCount:10240}
WARN[0000] resuming from checkpoint. low-watermark: {"Key":["s_w_id","s_i_id"],"ChunkSize":56767,"LowerBound":{"Value": ["50","23955"],"Inclusive":true},"UpperBound":{"Value": ["50","80723"],"Inclusive":false}} log-file: binlog.005904 log-pos: 27728 copy-rows: 4999991
INFO[0000] skip dump, use last binlog replication pos (binlog.005904, 27728) or GTID set <nil>
INFO[0000] begin to sync binlog from position (binlog.005904, 27728)
INFO[0000] start sync binlog at binlog file (binlog.005904, 27728)
INFO[0000] copy rows complete
WARN[0000] cutover deferred while sentinel table test._stock_sentinel exists; will wait 48s
INFO[0030] migration status: state=waitingOnSentinelTable sentinel-table=test._stock_sentinel total-time=30s sentinel-wait-time=30s sentinel-max-wait-time=48s conns-in-use=2
INFO[0030] finished periodic flush of binary log: duration=986.292µs
INFO[0039] sentinel table dropped at 2023-11-27 14:01:23.270618 -0700 MST m=+39.130231834
INFO[0039] starting to flush changeset
INFO[0039] blocking until we have read all binary logs: current-pos=(binlog.005905, 511513946) target-pos=(binlog.005905, 511513730)
INFO[0039] Running ANALYZE TABLE
INFO[0039] blocking until we have read all binary logs: current-pos=(binlog.005905, 511514342) target-pos=(binlog.005905, 511514126)
INFO[0039] starting checksum operation, this will require a table lock
WARN[0039] trying to acquire table lock, timeout: 30
WARN[0039] table lock acquired
INFO[0039] blocking until we have read all binary logs: current-pos=(binlog.005905, 511514558) target-pos=(binlog.005905, 511514342)
INFO[0039] starting to flush changeset
INFO[0039] blocking until we have read all binary logs: current-pos=(binlog.005905, 511514774) target-pos=(binlog.005905, 511514558)
WARN[0039] table lock released
INFO[0039] table unlocked, starting checksum
INFO[0044] checksum passed
INFO[0044] starting to flush changeset
INFO[0044] blocking until we have read all binary logs: current-pos=(binlog.005905, 511514990) target-pos=(binlog.005905, 511514774)
INFO[0044] starting to flush changeset
INFO[0044] blocking until we have read all binary logs: current-pos=(binlog.005905, 511515419) target-pos=(binlog.005905, 511515203)
WARN[0044] Attempting final cut over operation (attempt 1/5)
INFO[0044] Using cutover algorithm: rename-under-lock
WARN[0044] trying to acquire table lock, timeout: 30
WARN[0044] table lock acquired
WARN[0044] table lock released
WARN[0044] final cut over operation complete
INFO[0044] successfully dropped old table
INFO[0044] apply complete: instant-ddl=false inplace-ddl=false total-chunks=45 copy-rows-time=0s checksum-time=5s total-time=44s conns-in-use=0
INFO[0044] closing canal
INFO[0044] syncer is closing...
INFO[0044] kill last connection id 70
INFO[0044] syncer is closed
```

PR #251, opened 2023-12-04
Changed attempt-inplace-ddl to attempt-inplace-add-index

### LINKED ISSUE #247: Spirit rebuilds table for drop/rename secondary index
DROP KEY and RENAME KEY are technically "inplace" operations, not "instant", even though they do not rebuild the table and only modify metadata. (https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html)

Unfortunately, this means that spirit will completely copy and rebuild the table even though it should be safe to simply apply the operation directly.

kolbe:
We will automatically attempt to execute inplace operations other than add index.

Fixes #247

### DIFF
--- a/pkg/migration/migration.go
+++ b/pkg/migration/migration.go
@@ -9,21 +9,21 @@ import (
 )
 
 type Migration struct {
-	Host                   string        `name:"host" help:"Hostname" optional:"" default:"127.0.0.1:3306"`
-	Username               string        `name:"username" help:"User" optional:"" default:"msandbox"`
-	Password               string        `name:"password" help:"Password" optional:"" default:"msandbox"`
-	Database               string        `name:"database" help:"Database" optional:"" default:"test"`
-	Table                  string        `name:"table" help:"Table" optional:"" default:"stock"`
-	Alter                  string        `name:"alter" help:"The alter statement to run on the table" optional:"" default:"engine=innodb"`
-	Threads                int           `name:"threads" help:"Number of concurrent threads for copy and checksum tasks" optional:"" default:"4"`
-	TargetChunkTime        time.Duration `name:"target-chunk-time" help:"The target copy time for each chunk" optional:"" default:"500ms"`
-	AttemptInplaceAddIndex bool          `name:"attempt-inplace-add-index" help:"Attempt inplace add index (only safe without replicas or with Aurora Global)" optional:"" default:"false"`
-	Checksum               bool          `name:"checksum" help:"Checksum new table before final cut-over" optional:"" default:"true"`
-	ReplicaDSN             string        `name:"replica-dsn" help:"A DSN for a replica which (if specified) will be used for lag checking." optional:""`
-	ReplicaMaxLag          time.Duration `name:"replica-max-lag" help:"The maximum lag allowed on the replica before the migration throttles." optional:"" default:"120s"`
-	LockWaitTimeout        time.Duration `name:"lock-wait-timeout" help:"The DDL lock_wait_timeout required for checksum and cutover" optional:"" default:"30s"`
-	SkipPreRunChecks       bool          `name:"i-understand-mysql57-is-not-supported" hidden:"" default:"false"`
-	SkipDropAfterCutover   bool          `name:"skip-drop-after-cutover" help:"Keep old table after completing cutover" optional:"" default:"false"`
+	Host                 string        `name:"host" help:"Hostname" optional:"" default:"127.0.0.1:3306"`
+	Username             string        `name:"username" help:"User" optional:"" default:"msandbox"`
+	Password             string        `name:"password" help:"Password" optional:"" default:"msandbox"`
+	Database             string        `name:"database" help:"Database" optional:"" default:"test"`
+	Table                string        `name:"table" help:"Table" optional:"" default:"stock"`
+	Alter                string        `name:"alter" help:"The alter statement to run on the table" optional:"" default:"engine=innodb"`
+	Threads              int           `name:"threads" help:"Number of concurrent threads for copy and checksum tasks" optional:"" default:"4"`
+	TargetChunkTime      time.Duration `name:"target-chunk-time" help:"The target copy time for each chunk" optional:"" default:"500ms"`
+	ForceInplace         bool          `name:"force-inplace" help:"Force tttempting to use inplace (only safe without replicas or with Aurora Global)" optional:"" default:"false"`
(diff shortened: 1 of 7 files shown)


pkg/migration/runner.go
```diff
@@ -367,22 +369,59 @@ func (r *Runner) attemptMySQLDDL(ctx context.Context) error {
...
+
+	// Adding an index is only online-safe to do in Aurora GLOBAL
+	// because replicas do not use the binlog.
+
+	// If the operator has specified that they want to attempt
+	// an inplace add index, we will attempt inplace regardless
+	// of the statement.
+	if r.migration.AttemptInplaceAddIndex {
+		return nil
+	}
```
morgo: I think it's easier to test this function if you move this out, so the caller then says:

```
if isASafeStatementForInplace(stmt) || r.migration.AttemptInplace {
```

In terms of the signature of the function, I think it should just accept an alter statement and return a boolean:

```
func AlgorithmInplaceConsideredSafe(stmt string) bool {}
```

I think we can put this in the utils package, so that it simplifies the migration package which is kind of large. We can then write tests for it independently.

pkg/migration/migration.go
```diff
@@ -9,21 +9,21 @@ import (
...
+	Host                   string        `name:"host" help:"Hostname" optional:"" default:"127.0.0.1:3306"`
+	Username               string        `name:"username" help:"User" optional:"" default:"msandbox"`
+	Password               string        `name:"password" help:"Password" optional:"" default:"msandbox"`
+	Database               string        `name:"database" help:"Database" optional:"" default:"test"`
+	Table                  string        `name:"table" help:"Table" optional:"" default:"stock"`
+	Alter                  string        `name:"alter" help:"The alter statement to run on the table" optional:"" default:"engine=innodb"`
+	Threads                int           `name:"threads" help:"Number of concurrent threads for copy and checksum tasks" optional:"" default:"4"`
+	TargetChunkTime        time.Duration `name:"target-chunk-time" help:"The target copy time for each chunk" optional:"" default:"500ms"`
+	AttemptInplaceAddIndex bool          `name:"attempt-inplace-add-index" help:"Attempt inplace add index (only safe without replicas or with Aurora Global)" optional:"" default:"false"`
```
morgo: I think for ease of memory we should call this ForceInplace:
```suggestion
	ForceInplace bool          `name:"force-inplace" help:"Force attempting to use inplace (only safe without replicas or with Aurora Global since index adds take considerable time)" optional:"" default:"false"`
```

pkg/migration/runner.go
kolbe: Fixed

pkg/migration/migration.go
kolbe: Fixed

pkg/utils/utils.go
```diff
@@ -70,3 +72,28 @@ func StripPort(hostname string) string {
...
+	stmt := &stmtNodes[0]
+	alterStmt, ok := (*stmt).(*ast.AlterTableStmt)
+	if !ok {
+		return false
+	}
+	for _, spec := range alterStmt.Specs {
+		if spec.Tp == ast.AlterTableAddConstraint {
+			return false
+		}
+	}
```
morgo: I would use a switch stmt here. So:
```
switch spec.Tp {
 case ast.AlterTableAddConstraint:
  return false
 case knownGoodCase:
  continue
 default:
   // optionally log, don't know how to handle.
   return false
}

File: pkg/utils/utils.go [hunk 4]
```diff
@@ -70,3 +72,42 @@ func StripPort(hostname string) string {
 	}
 	return hostname
 }
+
+func AlgorithmInplaceConsideredSafe(sql string) error {
+	// INPLACE DDL is not generally safe for online use in MySQL 8.0,
+	// because ADD INDEX can block replicas. Some INPLACE operations
+	// *are* safe, because they only modify table metadata,
+	// so we parse the ALTER and attempt to use the INPLACE
+	// algorithm for statements that in practice execute instantly.
```
morgo: Please move the comment above the `func ..` and start it with `AlgorithmInplaceConsideredSafe checks to see ...`.

pkg/migration/runner.go
```diff
@@ -385,16 +385,28 @@ func (r *Runner) attemptMySQLDDL(ctx context.Context) error {
...
+	// because replicas do not use the binlog. Some, however,
+	// only modify the table metadata and are safe.
+	//
+	// If the operator has specified that they want to attempt
+	// an inplace add index, we will attempt inplace regardless
+	// of the statement.
+	alterStmt := fmt.Sprintf("ALTER TABLE %s %s", r.migration.Table, r.migration.Alter)
+	err = utils.AlgorithmInplaceConsideredSafe(alterStmt)
+	if err != nil {
+		r.logger.Warnf("inplace DDL is not safe for this statement: %v", err)
```
morgo: Small nit, but I think when the wrapping here combines with the error it will repeat itself a lot. Maybe for the wrapping let's use a brief statement like this:
```suggestion
	alterStmt := fmt.Sprintf("ALTER TABLE %s %s", r.migration.Table, r.migration.Alter)
	err = utils.AlgorithmInplaceConsideredSafe(alterStmt)
	if err != nil {
		r.logger.Warnf("inplace DDL is not safe: %v", err)
```

pkg/utils/utils.go
```diff
@@ -70,3 +72,42 @@ func StripPort(hostname string) string {
...
+			continue
+		default:
+			unsafeClauses++
+		}
+	}
+	if unsafeClauses > 0 {
+		if len(alterStmt.Specs) > 1 {
+			return fmt.Errorf("some clause(s) of the ALTER staement are not INPLACE safe (note: combinations of INSTANT and INPLACE operations can *not* currently be detected to be safe; consider executing these as separate ALTER statements)")
+		}
+		return fmt.Errorf("ALTER TABLE statement is not safe for INPLACE")
```
morgo: As per above, the error wrapping should work with the error message here. Plus we can include reference to the force option:
```suggestion
		if len(alterStmt.Specs) > 1 {
			return fmt.Errorf("ALTER contains multiple clauses. Combinations of INSTANT and INPLACE operations can not be detected safely. Consider executing these as separate ALTER statements.  Use --force-inplace to overwrite this safety check.")
		}
		return fmt.Errorf("ALTER either does not support INPLACE or when performed as INPLACE could take considerable time. Use --force-inplace to overwrite this safety check.")
```

morgo (approved):
LGTM, just some small cleanup on error messages.

morgo (approved):

PR #253, opened 2023-12-07
Modifications necessary for running on MariaDB

bobzilladev:
Issue: https://github.com/cashapp/spirit/issues/252

I've been able to get spirit to work on a MariaDB `10.5.23` system, this a pull request demonstrating the changes I had to make. Unknown if MariaDB would be a goal of yours, but pt-osc is too slow for the size of some of our tables. Caveat that we haven't gone through large production alters with this change yet.


bobzilladev:
Closing as feature is not desired.

PR #255, opened 2023-12-07
Added check for illegal ALTER clauses

### LINKED ISSUE #254: Disallow explicit ALGORITHM and LOCK clauses
It's currently possible for an operator to include an explicit ALGORITHM or LOCK clause in the ALTER table statement they give to Spirit. This kind of steps on the toe's of Spirit's design that tries to carefully manage which algorithm and lock strategy are used. If a user wants to manually specify these things, maybe they don't want to use a tool like Spirit in the first place?

kolbe:
Fixes #254 

This PR adds a new PreRun check to make sure there are no "illegal" clauses in the ALTER statement given. Currently, this includes "ALGORITHM" and "LOCK" clauses.


morgo (commented):
LGTM, just one nit:

Please rename the files to be lowercase. It probably won't cause issues, but we currently use lower everywhere else.

morgo (approved):

PR #256, opened 2023-12-07
Add hidden --skip-checks option

kolbe:
The way the "version" check is currently skipped is pretty brute-force and means that any other PreRun checks would be skipped (if there were any other ones... such as the one I propose in #255).

This PR adds a hidden --skip-checks option that can be used to skip individual checks. Maybe people have their reasons. 🤷


morgo:
I'm not 100% against this PR, but I would prefer to set a bar that we must understand the use-case first before we allow this.

This is consistent with our [philosophy](https://github.com/cashapp/spirit/blob/main/.github/CONTRIBUTING.md#philosophy) of "decisions not options".

morgo:
I'm going to close this as obsolete. The version check is now not really skippable, because we've removed the 5.7 code.

PR #264, opened 2024-01-07
Make checksum failure errors account for unique index

morgo:

This is from https://github.com/cashapp/spirit/issues/263

(It doesn't fix it, just one of the issues from it. I believe the deadlock issue might be 5.7-Aurora related, which is unsupported).

It also bumps the included go-mysql-org/go-mysql version, which uses a newer TiDB parser.


pkg/migration/runner.go
```diff
@@ -397,7 +397,7 @@ func (r *Runner) attemptMySQLDDL(ctx context.Context) error {
 	alterStmt := fmt.Sprintf("ALTER TABLE %s %s", r.migration.Table, r.migration.Alter)
 	err = utils.AlgorithmInplaceConsideredSafe(alterStmt)
 	if err != nil {
-		r.logger.Warnf("inplace DDL is not safe: %v", err)
+		r.logger.Infof("unable to use INPLACE: %v", err)
```
morgo, line 400: I know we originally decided Warnf, but I spoke to an internal customer who was confused by this. So I am changing it slightly.

kolbe (approved):
LGTM

PR #267, opened 2024-01-12
Add support for force checksum on unique index

morgo:

Fixes https://github.com/cashapp/spirit/issues/266


pkg/migration/runner.go
```diff
@@ -209,6 +209,15 @@ func (r *Runner) Run(originalCtx context.Context) error {
 		return err
 	}
 
+	// Force enable the checksum if it's an ADD UNIQUE INDEX operation
+	// https://github.com/cashapp/spirit/issues/266
+	if !r.migration.Checksum {
+		if err := utils.AlterContainsAddUnique("ALTER TABLE unused " + r.migration.Alter); err != nil {
+			r.logger.Warnf("force enabling checksum: %v", err)
```
kolbe, line 216: Should the message contain an explanation for *why* checksum is being enabled?

morgo: Yes it will, that's in the `%v`. Specifically it will return:
> contains adding a unique index

 The idea of catching it from the `utils.AlterContainsAddUnique` func means that it can be contextual (but it's not really that contextual yet, it just contains a statement saying it contains a unique index.)

PR #268, opened 2024-02-04
initial support for RBR minimal image

morgo:

Fixes https://github.com/cashapp/spirit/issues/221

This required a change to the replication applier. Previously, on an update operation the before-image and after-image were inspected to extract the `PRIMARY KEY`. Now only the before image is used.

### DIFF
--- a/pkg/repl/client.go
+++ b/pkg/repl/client.go
@@ -129,13 +129,14 @@ func NewClientDefaultConfig() *ClientConfig {
 func (c *Client) OnRow(e *canal.RowsEvent) error {
 	var i = 0
 	for _, row := range e.Rows {
-		i++
-		if e.Action == canal.UpdateAction && i%2 == 0 {
-			// For UpdateAction there is always a before and after image (i.e. e.Rows is always in pairs.)
-			// The first row is the before image, the second is the after image.
-			// We only need to capture one of the events, and since in MINIMAL RBR row
-			// image the PK is only included in the before, we chose that one.
-			continue
+		// For UpdateAction there is always a before and after image (i.e. e.Rows is always in pairs.)
+		// We only need to capture one of the events, and since in MINIMAL RBR row
+		// image the PK is only included in the before, we chose that one.
+		if e.Action == canal.UpdateAction {
+			i++
+			if i%2 == 0 {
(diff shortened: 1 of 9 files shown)


pkg/repl/client.go
```diff
@@ -127,8 +127,20 @@ func NewClientDefaultConfig() *ClientConfig {
 // We only need to add the PK + if the operation was a delete.
 // This will be used after copy rows to apply any changes that have been made.
 func (c *Client) OnRow(e *canal.RowsEvent) error {
+	var i = 0
 	for _, row := range e.Rows {
-		key := c.table.PrimaryKeyValues(row)
+		i++
```
joycse06: I think we will have to increment this only if it's an `UpdateAction`, otherwise this will not work. Example:

```
// binlog updates
Update Row 1 // i = 0 => Correct
Update Row 2 // i = 1
Insert       // i = 2
Update Row 1 // i = 3 => This is where we will have the PK, but will be skipped with current code
Update Row 2 // i = 4 => Condition will be true here, but no PK on this row, so it will fail
```

joycse06 (approved):
One small thing, otherwise LGTM 👍

pkg/repl/client.go
morgo: No, I think it's correct as is. The for loop is for multiple rows being deleted/inserted/updated at once (it won't change statement type in a call to `OnRow`).

The scenario when Insert/Delete could have multiple loops is if it inserts/deletes multiple rows in an individual statement.

We only need to handle that `Update` e.Rows always comes in pairs.

morgo: For readability I could do this if you prefer:
```
	var i = 0
	for _, row := range e.Rows {
		if e.Action == canal.UpdateAction {
			// For UpdateAction there is always a before and after image (i.e. e.Rows is always in pairs.)
			// The first row is the before image, the second is the after image.
			// We only need to capture one of the events, and since in MINIMAL RBR row
			// image the PK is only included in the before, we chose that one.
			i++
			if i%2 == 0 {
				continue
			}
		}
	..
```

morgo: (I prefer it, I've changed it.)

PR #271, opened 2024-02-05
Fully remove 5.7

morgo:

Fixes https://github.com/cashapp/spirit/issues/157

There is a [branch kept here for 5.7 support](https://github.com/cashapp/spirit/tree/5.7-cash). We may cherry pick to it for the next couple of months if required.


prudhvi (approved):

pkg/check/version.go
```diff
@@ -21,14 +21,17 @@ func versionCheck(_ context.Context, r Resources, _ loggers.Advanced) error {
 		return err
 	}
 	defer db.Close()
-	// This ensures that we first return an error like
```
kolbe: After removing this, will the end user not ever see any errors about connectivity or authentication issues? That doesn't seem very friendly.

morgo: Good question, I think you are right. Let me fix it to restore this.

PR #272, opened 2024-02-05
Support burn-free initial table create

morgo:

Fixes https://github.com/cashapp/spirit/issues/241


pkg/migration/runner_test.go
```diff
@@ -2594,3 +2594,62 @@ func TestPreRunChecksE2E(t *testing.T) {
...
+		Threads:  1,
+		Table:    "instantburn",
+		Alter:    "add newcol2 int",
+	})
+	assert.NoError(t, err)
+	err = m.Run(context.Background())
+	assert.NoError(t, err)
+
+	assert.False(t, m.usedInstantDDL) // it would have had to apply a copy.
+	assert.Equal(t, 0, rowVersions()) // confirm we reset to zero, not 1 (no burn)
```
prudhvi, line 2658: So every time we do a non instant alter on table, `total_row_versions`, resets to zero?

pkg/migration/runner.go
```diff
@@ -545,7 +545,7 @@ func (r *Runner) createNewTable(ctx context.Context) error {
 // alterNewTable applies the ALTER to the new table.
 // It has been pre-checked it is not a rename, or modifying the PRIMARY KEY.
 func (r *Runner) alterNewTable(ctx context.Context) error {
-	if err := dbconn.Exec(ctx, r.db, "ALTER TABLE %n.%n "+r.migration.Alter,
+	if err := dbconn.Exec(ctx, r.db, "ALTER TABLE %n.%n "+utils.TrimAlter(r.migration.Alter)+", ALGORITHM=COPY",
```
prudhvi, line 557: Nice one without the need for temp table and I assume this is reasonably fast/ same speed as instant for empty table.

pkg/utils/utils.go
```diff
@@ -156,3 +156,7 @@ func AlterIsAddUnique(sql string) error {
 	}
 	return nil
 }
+
+func TrimAlter(alter string) string {
+	return strings.TrimSuffix(strings.TrimSpace(alter), ";")
+}
```
prudhvi, line 162: Should we also trim `Algorithm=xxx`, I remember one user tried to add the algorithm in the alter statement, I forgot if tern already cleans up or spirit cleans up in the beginning

prudhvi (approved):

pkg/migration/runner_test.go
morgo: That's correct.

pkg/utils/utils.go
morgo: Yeah, this is tested in https://github.com/cashapp/spirit/blob/main/pkg/check/illegalclause.go

I didn't add a check for the `;` suffix because the actual cases which broke were in the testsuite. It was only a couple, but I thought since the only affected area that needed alter trimming was in one place I could just wrap it in this function.

pkg/migration/runner.go
morgo: Yeah, it's basically the same cost because it's altering an empty table at this point.

pkg/migration/runner_test.go
```diff
@@ -2594,3 +2594,62 @@ func TestPreRunChecksE2E(t *testing.T) {
...
+		pad varbinary(1024) NOT NULL,
+		PRIMARY KEY (id)
+	)`
+	rowVersions := func() int {
+		// Check that the number of total_row_versions is Zero (i'e doesn't burn)
+		db, err := dbconn.New(testutils.DSN(), dbconn.NewDBConfig())
+		assert.NoError(t, err)
+		defer db.Close()
+		var rowVersions int
+		err = db.QueryRow(`SELECT total_row_versions FROM INFORMATION_SCHEMA.INNODB_TABLES where name='test/instantburn'`).Scan(&rowVersions)
```
prudhvi, line 2633: Just curious, I assume deleted tables don't have entry in this `INFORMATION_SCHEMA.INNODB_TABLES`  right?

morgo: That's right.

PR #274, opened 2024-03-07
reduce checksum failures

morgo:

This is another attempt to fix https://github.com/cashapp/spirit/issues/248

There seems to be a case where the checksum is failing, and inspection of the rows show that it's not garbled data but a `last_viewed_at DATETIME(3)` column which an update was missed for.

All the cases affected appear to be resume cases (typically it's gone to resume because it has high lock wait timeouts while trying to get a lock). Which leads me to believe it's:

1. The flushing under lock mechanism. This PR changes it to be the same as used in 8.0 cutover
2. The key above watermark optimization is incorrectly ignoring rows. I've added protection to make sure there's no race on resume where it could return incorrectly 
3. The binlog checkpoint is ahead of itself. We're pretending work has safely been applied, but it hasn't.

Note: (2) is more likely since (1) occurs under both resume and non-resume cases. I don't think the cause is (3).

There is also a little bit more cleanup to remove 5.7 support.

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -254,6 +254,40 @@ func (c *Checker) setInvalid(newVal bool) {
 	c.isInvalid = newVal
 }
 
+func (c *Checker) initConnPool(ctx context.Context) error {
+	// Try and catch up before we apply a table lock,
+	// since we will need to catch up again with the lock held
+	// and we want to minimize that.
+	if err := c.feed.Flush(ctx); err != nil {
+		return err
+	}
+	// Lock the source table in a trx
+	// so the connection is not used by others
+	c.logger.Info("starting checksum operation, this will require a table lock")
+	serverLock, err := dbconn.NewTableLock(ctx, c.db, c.table, true, c.dbConfig, c.logger)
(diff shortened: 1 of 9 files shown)


pkg/checksum/checker.go
```diff
@@ -273,42 +264,45 @@ func (c *Checker) Run(ctx context.Context) error {
...
 	if err != nil {
 		return err
 	}
 	defer serverLock.Close()
-
 	// With the lock held, flush one more time under the lock tables.
 	// Because we know canal is up to date this now guarantees
 	// we have everything in the new table.
-	if err := c.feed.Flush(ctx); err != nil {
+	if err := c.feed.FlushUnderLock(ctx, serverLock); err != nil {
```
prudhvi, line 275: IIUC, This is also a important change to reduce the chances the below check `AllChangesFlushed()` failing due to parallelism, but I guess not the reason the migrations under consideration failed, as they failed in the checksum phase.

prudhvi (approved):
Thanks for explaining the changes.
Removing the races also cleanup up code and makes it more explicit and readable.

PR #280, opened 2024-03-25
migration: Add progress API

morgo:

Fixes https://github.com/cashapp/spirit/issues/279

### DIFF
--- a/pkg/migration/migration.go
+++ b/pkg/migration/migration.go
@@ -22,7 +22,6 @@ type Migration struct {
 	ReplicaDSN           string        `name:"replica-dsn" help:"A DSN for a replica which (if specified) will be used for lag checking." optional:""`
 	ReplicaMaxLag        time.Duration `name:"replica-max-lag" help:"The maximum lag allowed on the replica before the migration throttles." optional:"" default:"120s"`
 	LockWaitTimeout      time.Duration `name:"lock-wait-timeout" help:"The DDL lock_wait_timeout required for checksum and cutover" optional:"" default:"30s"`
-	SkipPreRunChecks     bool          `name:"i-understand-mysql57-is-not-supported" hidden:"" default:"false"`
 	SkipDropAfterCutover bool          `name:"skip-drop-after-cutover" help:"Keep old table after completing cutover" optional:"" default:"false"`
 	DeferCutOver         bool          `name:"defer-cutover" help:"Defer cutover (and checksum) until sentinel table is dropped" optional:"" default:"false"`
 }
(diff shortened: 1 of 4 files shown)


pkg/migration/runner_test.go
```diff
@@ -2324,9 +2329,19 @@ func TestDeferCutOver(t *testing.T) {
 		DeferCutOver:         true,
 	})
 	assert.NoError(t, err)
-	err = m.Run(context.Background())
-	assert.Error(t, err)
-	assert.ErrorContains(t, err, "timed out waiting for sentinel table to be dropped")
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
daniel-nichter, line 2333: Just a test, but standard way to doing this:
```go
go f() {
    wg.Add(1)
    defer wg.Done()
}
```
In a test, no big deal (except maybe if `m.Run` panics, then `wg.Done` isn't called).

pkg/migration/runner_test.go
```diff
@@ -1224,6 +1225,7 @@ func TestE2EBinlogSubscribingCompositeKey(t *testing.T) {
 	assert.NotNil(t, chunk)
 	assert.Equal(t, "((`id1` < 1001)\n OR (`id1` = 1001 AND `id2` < 1))", chunk.String())
 	assert.NoError(t, m.copier.CopyChunk(context.TODO(), chunk))
+	assert.Equal(t, Progress{CurrentState: "copyRows", Summary: "1000/1200 83.33% copyRows ETA TBD"}, m.GetProgress())
```
daniel-nichter: In my experience, tests on raw text output are extremely fragile and become problematic. One character changes, the tests fail, but a future engineer won't know if it's just a text diff or if there's a real issue. If there's a different way to check this, it'd be worth the time/effort to do.

pkg/migration/runner.go
```diff
@@ -599,6 +607,29 @@ func (r *Runner) createCheckpointTable(ctx context.Context) error {
 	return nil
 }
 
+func (r *Runner) GetProgress() Progress {
+	var summary string
+	switch r.getCurrentState() { //nolint: exhaustive
```
daniel-nichter, line 612: Add a default to check the inevitable future bug where, somehow, it's none of the proper cases. That way, user doesn't get an empty string; they can at least get something slightly better like "Error: status unknown (report this as a bug )".

pkg/migration/migration.go
```diff
@@ -33,14 +32,8 @@ func (m *Migration) Run() error {
 		return err
 	}
 	defer migration.Close()
-	if !m.SkipPreRunChecks {
```
daniel-nichter, line 36: This change doesn't seem related to adding progress?

daniel-nichter (approved):

pkg/migration/runner_test.go
morgo: There's a race in this suggestion though. It could reach wg.Wait() before it reaches wg.Add(1).

pkg/migration/migration.go
morgo: No, it's removing 5.7 specific dead code I caught. See the comment just below.

pkg/migration/runner.go
morgo: This is just for the field `summary`, which is a text-based editorialization of the CurrentState (i.e. 99% done). The `CurrentState` will always be available for the other states, but they won't have a text-based summary (for example they can't easily report progress). I don't consider this is an error, it just means there's no second level detail.

pkg/migration/runner_test.go
morgo: There are constants for the `CurrentState`. I can always change that. There are no constants for the summary though.

pkg/migration/runner_test.go
morgo: The linter also detects the race. So I will change the defer line, but not the `wg.Add()`

PR #287, opened 2024-05-19
Add new replica throttling query

morgo:

Fixes https://github.com/cashapp/spirit/issues/286

### DIFF
--- a/pkg/throttler/mysql80replica.go
+++ b/pkg/throttler/mysql80replica.go
@@ -11,11 +11,30 @@ type MySQL80Replica struct {
 	isClosed atomic.Bool
 }
 
-const MySQL8LagQuery = `SELECT CEIL(TIMESTAMPDIFF(MICROSECOND,
-max(LAST_APPLIED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP),
-max(LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP)
-)/1000) as lagMs
-FROM performance_schema.replication_applier_status_by_worker`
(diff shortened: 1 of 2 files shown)


pkg/throttler/mysql80replica.go
```diff
@@ -11,11 +11,30 @@ type MySQL80Replica struct {
...
 
-const MySQL8LagQuery = `SELECT CEIL(TIMESTAMPDIFF(MICROSECOND,
-max(LAST_APPLIED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP),
-max(LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP)
-)/1000) as lagMs
-FROM performance_schema.replication_applier_status_by_worker`
+// MySQL8LagQuery is a query that is used to get the lag between the source and the replica.
+// The implementation is described in https://github.com/cashapp/spirit/issues/286
+// It uses performance_schema instead of a heartbeat injection or seconds_behind_source.
+const MySQL8LagQuery = `WITH applier_latency AS (
```
joycse06, line 17: This query returns no rows on instances which are not replicas (I tried it on my local MySQL instance it returns Empty result set), I think probably because none of the tables has any rows.

For my understanding, is there anyway to change the query so it always return `0` regardless of if it's replica or not? (it's probably not required for any real cluster as for a **replica** these tables will always have data).

joycse06 (approved):
LGTM

pkg/throttler/mysql80replica.go
morgo: I've updated it to return 1 row.

It should always return rows though, because there is a pre-check for replication.

PR #289, opened 2024-05-23
Add support for MySQL 8.4

kolbe:

Fixes https://github.com/cashapp/spirit/issues/269


scripts/manual_install_84.sh
```diff
@@ -0,0 +1,138 @@
...
+		mysql -u root -S "$datadir/mysql.sock" -e "SELECT 1" > /dev/null && return
+	done
+	printf %s\\n "MySQL check of $datadir failed"
+	return 1
+}
+
+
+deploy_replication() {
+	num_nodes=2
+	base_port=22334
```
morgo: This feature in dbdeployer is to use the MySQL version as the port. We don't need to follow it strictly, you can just set this to 3306.

kolbe: Doesn't dbdeployer start numbering ports from 22334 or something when you deploy a replication topology?

morgo: I forget the exact math, but it's something like 34 = 8.0.34. And the 22 = 8 + an offset of 14 or so for replica.

i.e. for version 8.4.0, the dbdeployer port is 8400. For replicas it adds an offset.

morgo (approved):
LGTM. Can you update this line in `README.md`:

> It only supports MySQL 8.0

PR #293, opened 2024-05-30
Add asserty feature

morgo:

Fixes https://github.com/cashapp/spirit/issues/292

I will leave this open for a few days at least so we can discuss it.

### DIFF
--- a/pkg/table/asserty/asserty.go
+++ b/pkg/table/asserty/asserty.go
@@ -1,0 +1,68 @@
+// Package asserty offers functionality to assert for certain DB properties.
+package asserty
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+
+	"github.com/cashapp/spirit/pkg/table"
+	_ "github.com/go-sql-driver/mysql"
+)
+
+type Table struct {
+	ti *table.TableInfo
+}
+
+func LoadTable(db *sql.DB, schema, tableName string) (*Table, error) {
+	ti := table.NewTableInfo(db, schema, tableName)
+	if err := ti.SetInfo(context.TODO()); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


pkg/table/tableinfo.go
```diff
@@ -106,6 +110,29 @@ func (t *TableInfo) setRowEstimate(ctx context.Context) error {
 	return nil
 }
 
+func (t *TableInfo) setIndexes(ctx context.Context) error {
```
joycse06, line 113: Is `Load` a better word here?, `set` (as in setter methods) to me has the connotation that you are passing in the value which will be stored in a private property of the object, here we are loading them from the DB and then `setting` them.

Not a big one, I am happy with either.

pkg/table/asserty/asserty.go
```diff
@@ -0,0 +1,68 @@
...
+
+func LoadTable(db *sql.DB, schema, tableName string) (*Table, error) {
+	ti := table.NewTableInfo(db, schema, tableName)
+	if err := ti.SetInfo(context.TODO()); err != nil {
+		return nil, err
+	}
+	return &Table{ti: ti}, nil
+}
+
+func exists(val string, collection []string) bool {
```
joycse06: Can't you [use](https://golangtutorial.dev/tips/golang-slice-contains-method/) `slices.contains(collection, val)`?

joycse06 (approved):
LGTM

pkg/table/tableinfo.go
morgo: There are existing other functions like `setRowEstimate`, `setPrimaryKey` that this follows. IIRC it was based on an earlier reviewer to use `set` here, since on some of the other features, load had ambiguous usage.

PR #295, opened 2024-06-12
reduce deadlocks in recopying chunks

morgo:

Fixes https://github.com/cashapp/spirit/issues/263

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -40,6 +40,7 @@ type Checker struct {
 	logger           loggers.Advanced
 	fixDifferences   bool
 	differencesFound atomic.Uint64
+	recopyLock       sync.Mutex
 }
 
 type CheckerConfig struct {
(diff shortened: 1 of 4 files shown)


pkg/checksum/checker.go
```diff
@@ -229,8 +230,75 @@ func (c *Checker) replaceChunk(chunk *table.Chunk) error {
...
+	// Record lock, heap no 163 PHYSICAL RECORD: n_fields 2; compact format; info bits 32
+	// 0: len 30; hex <snip>; asc <snip>; (total 62 bytes);
+	// 1: len 8; hex <snip>; asc     <snip>;;
+	//
+	// *** WE ROLL BACK TRANSACTION (1)
+	//
+	// We don't need this to be an atomic transaction. We just need to delete from the _new table
+	// first so that any since-deleted rows (which wouldn't get removed by replace) are removed first.
+	// By doing this as two transactions we should be able to remove
+	// the opportunity for deadlocks.
```
prudhvi, line 288: Can deadlock happen with multiple sequential stmts in same txn?

 Is deadlock mainly because of parallel chunk recopying(across multiple txns) instead of single chunk delete and replace in a single txn?

If we have a mutex to avoid parallel copy of chunks, do we still need to split into multiple txns for a single chunk?

prudhvi (approved):

pkg/checksum/checker.go
morgo: > Can deadlock happen with multiple sequential stmts in same txn?
No, it shouldn't be possible to deadlock with itself. In the case identified it was two independent transactions from two different re-copy tasks.

> Is deadlock mainly because of parallel chunk recopying(across multiple txns) instead of single chunk delete and replace in a single txn?

I believe it may be a good enough fix to split delete and replace. But because we have no requirement to be same transaction I implemented both fixes.

> If we have a mutex to avoid parallel copy of chunks, do we still need to split into multiple txns for a single chunk?

It's more so that there is no requirement to be one transaction, and in future we may use parallelism again, but I don't think it's likely.

Issue #291, opened 2024-05-24
Validate minimum privileges for both source and replica in CI


morgo:
Currently we document the minimum privileges required for both source and replicas in `README.md`. However, the users we use in CI test are generic (msandbox, root) and have additional privileges.

We should change this so we can more easily catch if anything is introduced that needs additional privileges.

PR #300, opened 2024-06-24
Add support for running tests with docker-compose

sahilm:
TODO:
- [x] Tests with replicas
- [x] ~Docs (holding off docs till support is complete)~

* Fixes https://github.com/cashapp/spirit/issues/187
* Fixes https://github.com/cashapp/spirit/issues/291

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM golang:1.21
+
+WORKDIR /app
+COPY ../go.mod go.sum ./
+RUN go mod download
+COPY .. .
--- a/compose/bootstrap.sql
+++ b/compose/bootstrap.sql
@@ -1,0 +1,31 @@
+use mysql;
+
+create role if not exists R_DO_IT_ALL;
+grant all on *.* to R_DO_IT_ALL;
+create user if not exists msandbox@'%' identified with caching_sha2_password by 'msandbox';
+
+grant R_DO_IT_ALL to msandbox@'%' ;
+set default role R_DO_IT_ALL to msandbox@'%';
+
+
+create role if not exists R_REPLICATION;
+grant REPLICATION SLAVE, REPLICATION CLIENT on *.* to R_REPLICATION;
+create role if not exists R_THROTTLER;
+grant SELECT on performance_schema.replication_applier_status_by_worker to R_THROTTLER;
+grant SELECT on performance_schema.replication_connection_status to R_THROTTLER;
+create user if not exists rsandbox@'%' identified with caching_sha2_password by 'rsandbox';
+grant R_REPLICATION, R_THROTTLER to rsandbox@'%';
+set default role R_REPLICATION, R_THROTTLER to rsandbox@'%';
+
+flush privileges;
+
+create database if not exists test;
+
+use test;
+
+create table t1
+(
+    id int not null primary key auto_increment,
+    b  int not null,
+    c  int not null
+);
--- a/compose/bootstrap.sql
+++ b/compose/bootstrap.sql
@@ -8,7 +8,6 @@ create role if not exists R_THROTTLER;
 grant SELECT on performance_schema.replication_applier_status_by_worker to R_THROTTLER;
 grant SELECT on performance_schema.replication_connection_status to R_THROTTLER;
 
-
 create user if not exists msandbox@'%' identified with caching_sha2_password by 'msandbox';
 grant R_MIGRATOR, R_REPLICATION to msandbox@'%' ;
 set default role R_MIGRATOR, R_REPLICATION to msandbox@'%';
@@ -16,6 +15,13 @@ set default role R_MIGRATOR, R_REPLICATION to msandbox@'%';
 create user if not exists rsandbox@'%' identified with caching_sha2_password by 'rsandbox';
 grant R_REPLICATION, R_THROTTLER to rsandbox@'%';
 set default role R_REPLICATION, R_THROTTLER to rsandbox@'%';
+
+-- unfortunately the password for tsandbox must be the same as the password for root because in tests we switch to root
+-- using the same password.
+create user if not exists tsandbox@'%' identified with caching_sha2_password by 'msandbox';
+grant R_MIGRATOR, R_REPLICATION to tsandbox@'%' ;
+grant SYSTEM_VARIABLES_ADMIN on *.* to tsandbox@'%'; -- used in tests
+set default role R_MIGRATOR, R_REPLICATION to tsandbox@'%';
 
 flush privileges;
 
--- a/compose/bootstrap.sql
+++ b/compose/bootstrap.sql
@@ -3,13 +3,13 @@ use mysql;
 create role if not exists R_MIGRATOR;
 grant alter, create, delete, drop, index, insert, lock tables, select, trigger, update on *.* to R_MIGRATOR;
 create role if not exists R_REPLICATION;
-grant REPLICATION SLAVE, REPLICATION CLIENT on *.* to R_REPLICATION;
+grant replication slave, replication client on *.* to R_REPLICATION;
 create role if not exists R_THROTTLER;
-grant SELECT on performance_schema.replication_applier_status_by_worker to R_THROTTLER;
-grant SELECT on performance_schema.replication_connection_status to R_THROTTLER;
+grant select on performance_schema.replication_applier_status_by_worker to R_THROTTLER;
+grant select on performance_schema.replication_connection_status to R_THROTTLER;
 
 create user if not exists msandbox@'%' identified with caching_sha2_password by 'msandbox';
-grant R_MIGRATOR, R_REPLICATION to msandbox@'%' ;
+grant R_MIGRATOR, R_REPLICATION to msandbox@'%';
 set default role R_MIGRATOR, R_REPLICATION to msandbox@'%';
 
 create user if not exists rsandbox@'%' identified with caching_sha2_password by 'rsandbox';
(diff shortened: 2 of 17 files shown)


Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.22
```
morgo: We are using golang 1.21 elsewhere. I'd like to bump it, but in a new issue. That way when it comes to bumping to 1.23 we can have a new contributor look at what we did previously. There are a few places that need to be changed, including docs.

morgo:
This is looking really good! When you convert the replica tests, you can delete those `dbdeployer` workflows if you like. I just want to keep some of both running for a while just in case we've made any mistakes where tests aren't running.

There is also a related FR for improving the replication tests if you get to it: https://github.com/cashapp/spirit/issues/291

morgo:
LGTM, but asking @kolbe for secondary review since he understands this stuff better.

sahilm:
> This is looking really good! When you convert the replica tests, you can delete those `dbdeployer` workflows if you like. I just want to keep some of both running for a while just in case we've made any mistakes where tests aren't running.
> 
> There is also a related FR for improving the replication tests if you get to it: #291

I have added the replica stuff to this PR because there were quite a few changes and fixing #291 is a side-effect of how I've set things up.

scripts/buildandrun.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+set -e
+
+go build ./cmd/spirit
+
+if [ -n "$REPLICA_DSN" ]; then
+  ./spirit --replica-dsn "$REPLICA_DSN" --host $HOST --username $USERNAME --password $PASSWORD --database=$DATABASE --table=$TABLE
```
sahilm: Can we consider supporting `MYSQL_DSN` and `REPLICA_DSN` environment variables as overrides to the spirit CLI?

compose/replication/compose.yml
```diff
@@ -0,0 +1,77 @@
...
+        condition: service_healthy
+      mysql_replica:
+        condition: service_healthy
+    environment:
+      HOST: mysql:3306
+      USERNAME: msandbox
+      PASSWORD: msandbox
+      DATABASE: test
+      TABLE: t1
+      REPLICA_DSN: rsandbox:rsandbox@tcp(mysql_replica)/
```
sahilm, line 73: Must leave the database name blank because `rsandbox` has no privileges to access the DB. It's only a replication client.

compose/replication/compose.yml
```diff
@@ -0,0 +1,77 @@
...
+      context: ../../.
+    command: go test -race -v ./...
+    depends_on:
+      mysql:
+        condition: service_healthy
+      mysql_replica:
+        condition: service_healthy
+    environment:
+      MYSQL_DSN: msandbox:msandbox@tcp(mysql)/test
+      REPLICA_DSN: rsandbox:rsandbox@tcp(mysql_replica)/
```
sahilm, line 56: Must leave the database name in `REPLICA_DSN` blank because rsandbox has no privileges to access the DB. It's only a replication client.

compose/compose.yml
```diff
@@ -0,0 +1,46 @@
...
+
+  test:
+    build:
+      context: ../
+    command: go test -race -v ./...
+    depends_on:
+      mysql:
+        condition: service_healthy
+    environment:
+      MYSQL_DSN: root:msandbox@tcp(mysql)/test
```
sahilm, line 29: [Tests require `root`](https://github.com/cashapp/spirit/blob/788655ae00e1e320bc75b964cea337629f13c545/pkg/check/privileges_test.go#L19). This works with the `msandbox` user too because the password happens to be the same.

compose/compose.yml
```diff
@@ -0,0 +1,46 @@
...
+  buildandrun:
+    build:
+      context: ../
+    command: scripts/buildandrun.sh
+    depends_on:
+      mysql:
+        condition: service_healthy
+    environment:
+      HOST: mysql:3306
+      USERNAME: msandbox
```
sahilm, line 40: We run `spirit` with the `msandbox` user which now has the minimum privilege level needed.

compose/bootstrap.sql
```diff
@@ -0,0 +1,37 @@
...
+
+create user if not exists rsandbox@'%' identified with caching_sha2_password by 'rsandbox';
+grant R_REPLICATION, R_THROTTLER to rsandbox@'%';
+set default role R_REPLICATION, R_THROTTLER to rsandbox@'%';
+
+-- unfortunately the password for tsandbox must be the same as the password for root because in tests we switch to root
+-- using the same password.
+create user if not exists tsandbox@'%' identified with caching_sha2_password by 'msandbox';
+grant R_MIGRATOR, R_REPLICATION to tsandbox@'%' ;
+grant SYSTEM_VARIABLES_ADMIN on *.* to tsandbox@'%'; -- used in tests
```
sahilm, line 23: Must create this user with the additional privilege of `SYSTEM_VARIABLES_ADMIN` so that https://github.com/cashapp/spirit/blob/a36118d7b3f5675aa8403fe9fa4c4e98ae035957/pkg/check/configuration_test.go#L33-L32 passes. Still much better than using `root` or the earlier `msandbox` which was effectively root.

compose/bootstrap.sql
```diff
@@ -0,0 +1,37 @@
...
+
+create user if not exists rsandbox@'%' identified with caching_sha2_password by 'rsandbox';
+grant R_REPLICATION, R_THROTTLER to rsandbox@'%';
+set default role R_REPLICATION, R_THROTTLER to rsandbox@'%';
+
+-- unfortunately the password for tsandbox must be the same as the password for root because in tests we switch to root
+-- using the same password.
+create user if not exists tsandbox@'%' identified with caching_sha2_password by 'msandbox';
+grant R_MIGRATOR, R_REPLICATION to tsandbox@'%' ;
+grant references, super, process on *.* to tsandbox@'%'; -- used in tests
```
sahilm, line 23: Must create this user with these additional privilege because they're used in tests.Still should be much better than using `root` or the earlier `msandbox` which was effectively root.

scripts/buildandrun.sh
kolbe: I'd prefer not having spirit behave specially in the presence of bizarre environment variables when we have command-line options that do the job just fine.

kolbe:
My local attempt to run some of this is failing in a pretty surprising way! Maybe just an issue with cross-platform Docker stuff?

```docker compose -f docker/compose.yml -f docker/8.0.28.yml -f docker/test.yml up --abort-on-container-exit --force-recreate```

```
mysql-1  | 2024-06-25T16:58:10.142915Z 0 [ERROR] [MY-013183] [InnoDB] Assertion failure: fil0fil.cc:7628:file->n_pending_ios > 0 thread 140737102452480
mysql-1  | InnoDB: We intentionally generate a memory trap.
mysql-1  | InnoDB: Submit a detailed bug report to http://bugs.mysql.com/.
mysql-1  | InnoDB: If you get repeated assertion failures or crashes, even
mysql-1  | InnoDB: immediately after the mysqld startup, there may be
mysql-1  | InnoDB: corruption in the InnoDB tablespace. Please refer to
mysql-1  | InnoDB: http://dev.mysql.com/doc/refman/8.0/en/forcing-innodb-recovery.html
mysql-1  | InnoDB: about forcing recovery.
mysql-1  | 16:58:10 UTC - mysqld got signal 6 ;
mysql-1  | Most likely, you have hit a bug, but this error can also be caused by malfunctioning hardware.
mysql-1  | Thread pointer: 0x0
mysql-1  | Attempting backtrace. You can use the following information to find out
mysql-1  | where mysqld died. If you see no messages after this, something went
mysql-1  | terribly wrong...
mysql-1  | stack_bottom = 0 thread_stack 0x100000
mysql-1  | /usr/sbin/mysqld(my_print_stacktrace(unsigned char const*, unsigned long)+0x3d) [0x55555767957d]
mysql-1  | /usr/sbin/mysqld(print_fatal_signal(int)+0x303) [0x5555566f6453]
mysql-1  | /usr/sbin/mysqld(my_server_abort()+0x6e) [0x5555566f656e]
mysql-1  | /usr/sbin/mysqld(my_abort()+0xa) [0x55555767370a]
mysql-1  | /usr/sbin/mysqld(ut_dbg_assertion_failed(char const*, char const*, unsigned long)+0x326) [0x55555794ace6]
mysql-1  | /usr/sbin/mysqld(+0x253d825) [0x555557a91825]
mysql-1  | /usr/sbin/mysqld(fil_aio_wait(unsigned long)+0xef) [0x555557aa034f]
mysql-1  | /usr/sbin/mysqld(+0x2390dc0) [0x5555578e4dc0]
mysql-1  | /usr/sbin/mysqld(std::thread::_State_impl<std::thread::_Invoker<std::tuple<Detached_thread, void (*)(unsigned long), unsigned long> > >::_M_run()+0xc0) [0x5555578e5030]
mysql-1  | /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xbbb2f) [0x7fffff2b5b2f]
mysql-1  | /lib/x86_64-linux-gnu/libpthread.so.0(+0x7fa3) [0x7fffff7b8fa3]
mysql-1  | /lib/x86_64-linux-gnu/libc.so.6(clone+0x3f) [0x7ffffef95eff]
mysql-1  | The manual page at http://dev.mysql.com/doc/mysql/en/crashing.html contains
mysql-1  | information that should help you find out what is causing the crash.
mysql-1  | ERROR 2013 (HY000): Lost connection to MySQL server at 'reading initial communication packet', system error: 104
```

sahilm:
> My local attempt to run some of this is failing in a pretty surprising way! Maybe just an issue with cross-platform Docker stuff?

@kolbe 8.0.28 doesn't work on arm64. Oracle doesn't provide an arm64 build. All other versions should work fine on Mac laptops.

sahilm:
@morgo @kolbe Do you need any changes? If not may I have approval to merge this.

kolbe (approved):
LGTM

PR #304, opened 2024-06-24
Add --strict option

samongyr-sq:
Fixes https://github.com/cashapp/spirit/issues/297

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -157,7 +157,8 @@ Note that the checksum, if enabled, will be computed after the sentinel table is
 - Type: Boolean
 - Default value: FALSE
 
-When set to `TRUE`, Spirit will exit if it finds checkpoints belonging to some other migration with a different --alter to avoid losing progress.
+By default, Spirit will automatically clean up these old checkpoints before starting the schema change. This allows schema changes to always be possible to proceed forward, at the risk of lost progress.
 
-By default, Spirit will automatically clean up these old checkpoints before starting the migration. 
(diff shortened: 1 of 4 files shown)


morgo:
This is looking good, I appreciate the detail you've gone to in describing the command line option for `--help` output.

A couple of quick comments:

- Make sure to add a test
- Since you are adding an option, please also update `USAGE.md` with a description in this same PR.

USAGE.md
```diff
@@ -151,3 +151,13 @@ You can resume a migration from checkpoint and Spirit will start waiting again f
...
 Note that the checksum, if enabled, will be computed after the sentinel table is dropped. Because the checksum step takes an estimated 10-20% of the migration, the cutover will not occur immediately after the sentinel table is dropped.
+
+### strict
+
+- Type: Boolean
+- Default value: FALSE
+
+When set to `TRUE`, Spirit will exit if it finds checkpoints belonging to some other migration with a different --alter to avoid losing progress.
+
+By default, Spirit will automatically clean up these old checkpoints before starting the migration.
```
morgo: I would reverse the order of these two sentences, then it's easier to follow:

```suggestion
By default, Spirit will automatically clean up these old checkpoints before starting the schema change. This allows schema changes to always be possible to proceed forward, at the risk of lost progress.

When set to `TRUE`, if Spirit encounters a checkpoint belonging to a previous migration, it will validate that the alter statement matches the `--alter` parameter. If the validation fails, spirit will exit and prevent the schema change process from proceeding.

```

morgo (approved):
LGTM, just change the docs and let's merge.

PR #307, opened 2024-06-25
Jayj/old table timestamped

jayjanssen:
* Extends the old table name to include the current timestamp, but only if  `--skip-drop-after-cutover` is used.
  * This helps distinguish old table and prevent multiple old tables from being deleted
  * This has the potential of leaving more orphaned "old" tables in some tests since their names are now dynamic
* Check the table name length early in Preflight checks.  
* Calculates the possible lengths on runtime based on the table suffix formats.

Fixes https://github.com/cashapp/spirit/issues/296

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -393,9 +393,10 @@ func (r *Runner) runChecks(ctx context.Context, scope check.ScopeFlag) error {
 		ReplicaMaxLag:   r.migration.ReplicaMaxLag,
 		// For the pre-run checks we don't have a DB connection yet.
 		// Instead we check the credentials provided.
-		Host:     r.migration.Host,
-		Username: r.migration.Username,
-		Password: r.migration.Password,
+		Host:                 r.migration.Host,
+		Username:             r.migration.Username,
+		Password:             r.migration.Password,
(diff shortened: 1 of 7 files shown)


pkg/migration/runner.go
```diff
@@ -591,8 +593,11 @@ func (r *Runner) alterNewTable(ctx context.Context) error {
 }
 
 func (r *Runner) dropOldTable(ctx context.Context) error {
-	oldName := fmt.Sprintf("_%s_old", r.table.TableName)
-	return dbconn.Exec(ctx, r.db, "DROP TABLE IF EXISTS %n.%n", r.table.SchemaName, oldName)
+	return dbconn.Exec(ctx, r.db, "DROP TABLE IF EXISTS %n.%n", r.table.SchemaName, r.oldTableName())
+}
+
+func (r *Runner) oldTableName() string {
+	return fmt.Sprintf("_%s_old_%s", r.table.TableName, r.startTime.UTC().Format("20060102_150405Z"))
```
pamelayin: For shift we're using timestamp down to milliseconds, would this be distinguishable enough?

morgo:
Change LGTM when you are ready to merge. I like the design.

morgo (approved):

PR #313, opened 2024-06-28
Reject read only connections to primary nodes

sahilm:
So that we recycle the connection if we inadvertently connect to an old primary which is now a read only replica. This behaviour has been observed during blue/green upgrades and failover on AWS Aurora. See also: https://github.com/go-sql-driver/mysql?tab=readme-ov-file#rejectreadonly

Fixes https://github.com/cashapp/spirit/issues/250

### DIFF
--- a/pkg/dbconn/conn_test.go
+++ b/pkg/dbconn/conn_test.go
@@ -13,25 +13,25 @@ func TestNewDSN(t *testing.T) {
 	dsn := "root:password@tcp(127.0.0.1:3306)/test"
 	resp, err := newDSN(dsn, NewDBConfig())
 	assert.NoError(t, err)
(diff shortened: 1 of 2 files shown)


pkg/dbconn/conn_test.go
```diff
@@ -59,3 +59,27 @@ func TestNewConn(t *testing.T) {
...
+	// See also:
+	// docs: https://github.com/go-sql-driver/mysql?tab=readme-ov-file#rejectreadonly
+	// rejectReadOnly test in `go-sql-driver/mysql`: https://github.com/go-sql-driver/mysql/blob/52c1917d99904701db2b0e4f14baffa948009cd7/driver_test.go#L2270-L2301
+	_, err = db.Exec("insert into t1 values (1, 2, 3)")
+	assert.NoError(t, err)
+
+	var count int
+	err = db.QueryRow("select count(*) from t1 where a=1").Scan(&count)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, count)
```
morgo: Nice idea for a test. You could try:
- Setting the pool size to 1
- Setting a session variable in the connection as well
 
That way when the insert fails and a new connection is created, you can ensure the session variable is no longer set and that the connection is truly recycled.

morgo (approved):

PR #316, opened 2024-06-28
add CopierEstimateHistory

jayjanssen:

Fixes https://github.com/cashapp/spirit/issues/137

### DIFF
--- a/pkg/row/copier.go
+++ b/pkg/row/copier.go
@@ -52,6 +52,7 @@ type Copier struct {
 	dbConfig             *dbconn.DBConfig
 	logger               loggers.Advanced
 	metricsSink          metrics.Sink
+	CopierEtaHistory     *CopierEtaHistory
 }
 
 type CopierConfig struct {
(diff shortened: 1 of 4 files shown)


pkg/row/copier.go
```diff
@@ -90,16 +91,17 @@ func NewCopier(db *sql.DB, tbl, newTable *table.TableInfo, config *CopierConfig)
...
+		table:            tbl,
+		newTable:         newTable,
+		concurrency:      config.Concurrency,
+		finalChecksum:    config.FinalChecksum,
+		Throttler:        config.Throttler,
+		chunker:          chunker,
+		logger:           config.Logger,
+		metricsSink:      config.MetricsSink,
+		dbConfig:         config.DBConfig,
+		CopierEtaHistory: NewCopierEtaHistory(),
```
morgo: This field and function can be made private, since they (at least currently) are not being called outside of this package.

pkg/row/copier.go
```diff
@@ -275,7 +277,13 @@ func (c *Copier) GetETA() string {
 	// c.getCopyStats() and rowsPerSecond change estimation method when the PK is auto-inc.
 	remainingRows := totalRows - copiedRows
 	remainingSeconds := math.Floor(float64(remainingRows) / float64(rowsPerSecond))
-	return time.Duration(remainingSeconds * float64(time.Second)).String()
+
+	estimate := time.Duration(remainingSeconds * float64(time.Second))
+	comparison := c.CopierEtaHistory.AddCurrentEstimateAndCompare(estimate)
```
morgo: Same comment about making functions private.

I think this is thread safe because there is a mutex on `GetETA()`, but just curious if you've put any thought into it?

pkg/row/copier_eta_history.go
```diff
@@ -0,0 +1,87 @@
...
+	diffStr := shortDuration(relativeDiff.Round(time.Minute))
+	if diffStr == "0s" {
+		diffStr = "±0m"
+	}
+
+	return diffStr + " from " + shortDuration(ago.Round(time.Minute)) + " ago"
+}
+
+func shortDuration(d time.Duration) string {
+	s := d.String()
```
morgo, line 87: You can probably use something like `d.Round(time.Hour)`  here instead https://pkg.go.dev/time#Duration.Round

jayjanssen: That yields strings like `1h0m0s` instead of `1h`.

pkg/row/copier.go
jayjanssen: I did not, but it's easy enough

pkg/row/copier.go
jayjanssen: yeah, they kind of should be isolated from copier, but I wasn't sure where to put them (util maybe?)

Anyway, changed all the funcs to private

pkg/row/copier_eta_history.go
morgo: Ahh I see, nvm.

morgo (approved):
LGTM

PR #332, opened 2024-07-24
Fix changes not synced error

morgo:

Fixes https://github.com/cashapp/spirit/issues/331 (I think)

It's a bit of a yak shaving exercise:

- The check about DeltaLen=0 is safe but the check about binlog positioning not advancing is not. It doesn't account for the fact that other tables can still be written to.
- The FlushUnderLock should run `BlockWait()` to make sure changes in the last Flush are accounted for.
- BlockWait was previously calling `injectBinlogNoise()`. We can't just call `injectBinlogNoise()` under a lock, because that table was not part of the set of tables being locked. We could instead try a different binlog advancement technique when under lock, but this seems messy. We can instead remove it. The root cause of this is that canal does not sync on all DDL statements, notably `ANALYZE TABLE` which we do right before calling this! See upstream bug fix @ https://github.com/go-mysql-org/go-mysql/pull/900
- Because the canal fix also changes `canal.CatchMasterPos()` to no longer require `FLUSH BINARY LOGS` we can change `BlockWait()` to be based on it.
- `serverLock` as a variable name has been renamed to `tableLock` so it's not confusing exactly what is locked. This is also the root cause of the issue, since changes to other tables were not locked (nor expected to be), but the pedantic check did not allow for them to be.

This also fixes https://github.com/cashapp/spirit/issues/330 at the same time.


morgo:
There are data races now because the table setXXX functions can now be called any time. I will have to fix them, but it will take some refactoring.

pkg/repl/client.go
```diff
@@ -381,7 +369,8 @@ func (c *Client) startCanal() {
 			// but since canal is now closed we can safely return
 			return
 		}
-		c.logger.Errorf("canal has failed. error: %v", err)
+
+		c.logger.Errorf("canal has failed. error: %v, table: %s", err, c.table.TableName)
```
morgo, line 380: This is helpful for when tests fail

pkg/migration/migration_test.go
```diff
@@ -181,7 +181,7 @@ func TestRenameInMySQL80(t *testing.T) {
 	migration.Database = cfg.DBName
 	migration.Threads = 16
 	migration.Checksum = true
-	migration.Table = "t1"
+	migration.Table = "renamet1"
```
morgo, line 184: This was technically an existing bug in the test.

pkg/repl/client.go
```diff
@@ -239,12 +229,19 @@ func (c *Client) SetPos(pos mysql.Position) {
...
+
+	// We check if the position canal is up to is the same position
+	// as what we've made changes for. If this is zero it's a good
+	// indicator that we are up to date. However, because the
+	// "server lock" is not a global lock, it's possible that the synced
+	// position could still advance.
+	if c.canal.SyncedPosition().Compare(c.binlogPosSynced) != 0 {
+		c.logger.Warnf("Binlog reader info canal-position=%v synced-position=%v. Discrepancies could be due to modifications on other tables.", c.canal.SyncedPosition(), c.binlogPosSynced)
+	}
+	return deltaLen == 0
```
morgo, line 244: This is the main bug. The check here was too strict. If there were any other changes from other tables, it would advance the `SyncedPosition` and not compare as 0.

This can only be guaranteed on idle systems, or systems where there is one dominant table, which is being modified.

pkg/repl/client.go
```diff
@@ -395,18 +391,23 @@ func (c *Client) Close() {
...
+//     This is required to satisfy the binlog position is updated for the c.AllChangesFlushed() check.
 func (c *Client) FlushUnderLock(ctx context.Context, lock *dbconn.TableLock) error {
+	if err := c.flush(ctx, true, lock); err != nil {
+		return err
+	}
+	// Wait for the changes flushed to be received.
+	if err := c.BlockWait(ctx); err != nil {
+		return err
+	}
+	// Do a final flush
```
morgo: This catches another bug that the flush generates binlog activities which then causes the compare (`c.canal.SyncedPosition().Compare(c.binlogPosSynced)`) to show changes.

The compare could still show changes from other tables, but it will be less noisy because its guaranteed all changes here are read through.

kolbe (approved):

PR #342, opened 2024-08-28
Use quoted table name as lock name

prudhvi:
Fixes https://github.com/cashapp/spirit/issues/341


jayjanssen (approved):
makes sense to me.

PR #344, opened 2024-09-02
Add resume from checksum

morgo:

This fixes https://github.com/cashapp/spirit/issues/317

In our case we support yield via automation (pod cycling every 24hrs), so I am going to call this fixed.. even though it only completes part of the described FR.

Also fixes https://github.com/cashapp/spirit/issues/320 and fixes https://github.com/cashapp/spirit/issues/315 - the root cause is depending when the context cancel is detected, the error message can vary.

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -42,7 +42,7 @@ const (
 
 // These are really consts, but set to var for testing.
 var (
-	checkpointDumpInterval  = 2 * time.Second
+	checkpointDumpInterval  = 50 * time.Second
 	tableStatUpdateInterval = 5 * time.Minute
 	statusInterval          = 30 * time.Second
 	sentinelCheckInterval   = 1 * time.Second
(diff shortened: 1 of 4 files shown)


pkg/migration/runner_test.go
```diff
@@ -944,6 +955,77 @@ func TestCheckpointRestore(t *testing.T) {
...
+	}
+
+	assert.NoError(t, r.checksum(context.TODO()))       // run the checksum, the original Run is blocked on sentinel.
+	assert.NoError(t, r.dumpCheckpoint(context.TODO())) // dump a checkpoint with the watermark.
+	cancel()                                            // unblock the original waiting on sentinel.
+	assert.NoError(t, r.Close())                        // close the run.
+
+	// drop the sentinel table.
+	testutils.RunSQL(t, `DROP TABLE _cptresume_sentinel`)
+
```
prudhvi, line 1009: Can we test by inserting some new rows before restarting, even if it has checksum watermark it would still resume from copy phase that way and go to checksum phase right?

prudhvi (approved):

pkg/migration/runner.go
```diff
@@ -821,20 +835,24 @@ func (r *Runner) checksum(ctx context.Context) error {
...
+	// Handle a case just in the tests not having a dbConfig
 	if r.dbConfig == nil {
 		r.dbConfig = dbconn.NewDBConfig()
 	}
 	r.db.SetMaxOpenConns(r.dbConfig.MaxOpenConnections + 2)
 	var err error
 	for i := range 3 { // try the checksum up to 3 times.
+		if i > 0 {
+			r.checksumWatermark = "" // reset the watermark if we are retrying.
+		}
```
prudhvi, line 847: Is this intentional? so resume from checksum watermark only works if first resume is successful, otherwise it starts checksum from scratch again?

morgo: Yes that's right. If there is any failure we recheck everything.

pkg/migration/runner_test.go
morgo: Good idea, done.

PR #350, opened 2024-09-27
Use tablename for metadatalock name as tablename respsects both the m…

prudhvi:
Fixes https://github.com/cashapp/spirit/issues/347


morgo (approved):
Nice! I like the added test in `TestResumeFromCheckpointE2EWithManualSentinel`.

PR #357, opened 2024-12-03
use a hash for metadata lock name

kolbe:
Previously, metadata locks would conflict across same-named tables in different schemas. This PR fixes that issue by using a combination of schema name, table name, and a hash value for the metadata lock name.

### DIFF
--- a/pkg/dbconn/metadatalock.go
+++ b/pkg/dbconn/metadatalock.go
@@ -57,6 +57,11 @@ func NewMetadataLock(ctx context.Context, dsn string, table *table.TableInfo, lo
 		// The hash is truncated to 8 characters to avoid the maximum lock length.
 		// bizarrely_long_schema_name.thisisareallylongtablenamethisisareallylongtablename60charac ==>
 		// bizarrely_long_schem.thisisareallylongtablenamethisis-66fec116
+		//
+		// The computation of the hash is done server-side to simplify the whole process,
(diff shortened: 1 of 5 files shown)


pkg/dbconn/metadatalock.go
```diff
@@ -52,28 +51,34 @@ func NewMetadataLock(ctx context.Context, dsn string, lockName string, logger lo
...
 
 	// Acquire the lock or return an error immediately
 	// We only Infof the initial acquisition.
-	logger.Infof("attempting to acquire metadata lock: %s", lockName)
+	logger.Infof("attempting to acquire metadata lock for %s.%s", table.SchemaName, table.TableName)
 	if err = getLock(); err != nil {
 		return nil, err
 	}
-	logger.Infof("acquired metadata lock: %s", lockName)
+	logger.Infof("acquired metadata lock: %s.%s", table.SchemaName, table.TableName)
```
prudhvi, line 86: May be worth logging `lockName` too in one of the logs somewhere

prudhvi (approved):

pkg/dbconn/metadatalock.go
kolbe: It's not easy to do that, because I'm computing the lock name server-side inside the GET_LOCK function call. So for now let's leave it like this and if in the future we have a real need to log the lock name, I think we could just add another MySQL round-trip to compute the lock name and then use it in GET_LOCK.

PR #374, opened 2025-02-16
compose lock name on spirit side - clients acquiring locks can get to know their exact lock name

Kiran01bm:
## Change Summary
Compose lockName on the spirit side - improved logging and also exporting the lockName so clients acquiring locks can get to know their exact lockName. Migration Runner test is one such client. 

Local tests have passed ✅.


kolbe (approved):
LGTM!

PR #375, opened 2025-02-17
Add support for --statement

morgo:

Fixes https://github.com/cashapp/spirit/issues/364

It refactors quite a bit to add an AbstractStatement type for tracking what change is being made.


pkg/migration/migration.go
```diff
@@ -47,3 +52,52 @@ func (m *Migration) Run() error {
...
+		m.ReplicaMaxLag = 120 * time.Second
+	}
+	if m.Host == "" {
+		return errors.New("host is required")
+	}
+	if !strings.Contains(m.Host, ":") {
+		m.Host = fmt.Sprintf("%s:%d", m.Host, 3306)
+	}
+	if m.Database == "" {
+		return errors.New("schema name is required")
```
kolbe: Is it possibly a bit confusing for the option to be DB/Database and the error to refer only to "schema"?

pkg/migration/migration.go
```diff
@@ -47,3 +52,52 @@ func (m *Migration) Run() error {
...
+	}
+	if m.Statement != "" { // statement is specified
+		if m.Table != "" || m.Alter != "" {
+			return errors.New("only --statement or --table and --alter can be specified")
+		}
+		// extract the table and alter from the statement.
+		// if it is a CREATE INDEX statement, we rewrite it to an alter statement.
+		// This also returns the StmtNode, which we do nothing with so far.
+		// TODO: if it's a CREATE INDEX statement, do we store the rewritten
+		// form in m.Statement?
```
kolbe: It's probably a good idea to preserve the original statement submitted by the user, so that if there are errors or multiple statements in their workflow it'll be obvious what's happening.

pkg/statement/statement.go
```diff
@@ -0,0 +1,256 @@
...
+		}, nil
+	case *ast.CreateIndexStmt:
+		// Need to rewrite to a corresponding ALTER TABLE statement
+		return convertCreateIndexToAlterTable(stmtNodes[0])
+	// returning an empty alter means we are allowed to run it
+	// but it's not a spirit migration. But the table should be specified.
+	case *ast.CreateTableStmt:
+		stmt := stmtNodes[0].(*ast.CreateTableStmt)
+		if stmt.Table.Schema.String() != "" {
+			return nil, ErrSchemaNameIncluded
```
kolbe: Did you somehow validate that there aren't any existing cases of explicit schema names in CREATE INDEX statements in any of our zillion existing migration files?

kolbe (approved):

pkg/statement/statement.go
kolbe: See https://github.com/morgo/spirit/pull/1 for a possible solution

morgo: I like it, let's do it.

pkg/migration/migration.go
morgo: I originally had abstractStatement contain `statement` and `normalizedStatement`, but simplified it to reduce some of the complexity. I think a workaround is to just add a SQL comment saying "/* rewritten to ALTER TABLE */". I will do that.

PR #382, opened 2025-03-25
Fix non printable characters in checkpoint

morgo:

This fixes https://github.com/cashapp/spirit/issues/381  

- Checkpoints with non-printable values are hex encoded (and quoted) and if it looks like a hex value on restore, it's decoded. Special care is taken on encoding to force-encode strings that start with the hex prefix.
- Non printable values are hex encoded in log output.

### DIFF
--- a/pkg/table/datum.go
+++ b/pkg/table/datum.go
@@ -1,9 +1,11 @@
 package table
 
 import (
+	"encoding/hex"
 	"fmt"
 	"math"
 	"strconv"
+	"unicode/utf8"
 
 	"github.com/cashapp/spirit/pkg/dbconn/sqlescape"
 )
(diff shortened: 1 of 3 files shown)


pkg/table/datum.go
```diff
@@ -82,6 +84,16 @@ func datumValFromString(val string, tp datumTp) (interface{}, error) {
 	} else if tp == unsignedType {
 		return strconv.ParseUint(val, 10, 64)
 	}
+	// If it starts with 0x then it's a binary string. If it was actually
+	// a string that contained 0x, then we have encoded it as hex anyway.
+	// see pkg/table/chunk.go:valuesString()
+	if len(val) > 2 && val[0:2] == "0x" {
```
kolbe: Maybe just use strings.HasPrefix instead?

kolbe (approved):

pkg/table/datum.go
morgo: Good call.

PR #403, opened 2025-04-10
Support add partition. Fix syntax errors for partition statements.

JashLal:
This follows https://github.com/cashapp/spirit/pull/387 and fixes https://github.com/cashapp/spirit/issues/383. 

`DROP PARTITION` and `TRUNCATE PARTITION` statements require that `algorithm` and `lock` statements are placed before the partition statement and after the alter statement. Otherwise, the syntax is incorrect and the operation will fail. This PR also allows `ADD PARTITION` since it can be performed safely inplace without a lock for range/list partitioned tables. The operation fails with a mysql syntax error when tried on a hash/key partitioned table preventing an unsafe DDL.

### DIFF
--- a/pkg/migration/runner_test.go
+++ b/pkg/migration/runner_test.go
@@ -311,6 +311,60 @@ func TestOnline(t *testing.T) {
 	assert.False(t, m.usedInstantDDL) // unfortunately false in 8.0, see https://bugs.mysql.com/bug.php?id=113355
 	assert.False(t, m.usedInplaceDDL) // unfortunately false, since it combines INSTANT and INPLACE operations
 	assert.NoError(t, m.Close())
+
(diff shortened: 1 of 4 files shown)


pkg/migration/runner_test.go
```diff
@@ -311,6 +311,60 @@ func TestOnline(t *testing.T) {
...
+		Threads:      16,
+		Table:        "testonline6",
+		Alter:        "add partition partitions 4",
+		ForceInplace: false,
+	})
+	assert.NoError(t, err)
+	err = m.Run(t.Context())
+	assert.NoError(t, err)
+	assert.False(t, m.usedInstantDDL)
+	assert.False(t, m.usedInplaceDDL) // false, hash/key partitioned tables require a lock
```
morgo, line 338: ~nit: Duplicate line.~

nvm, my bad.

morgo (approved):

PR #409, opened 2025-04-22
Fix the type conversion issue when resuming from checkpoint for PK ty…

prudhvi:
Fixes one case for https://github.com/cashapp/spirit/issues/360


pkg/row/copier.go
```diff
@@ -112,7 +112,9 @@ func NewCopierFromCheckpoint(db *sql.DB, tbl, newTable *table.TableInfo, config
 		return c, err
 	}
 	// Overwrite the previously attached chunker with one at a specific watermark.
-	if err := c.chunker.OpenAtWatermark(lowWatermark, newTable.MaxValue()); err != nil {
+	// For high watermark, use the type of old table, not the new one.
+	highPtr := table.NewDatum(newTable.MaxValue().Val, tbl.MaxValue().Tp)
+	if err := c.chunker.OpenAtWatermark(lowWatermark, highPtr); err != nil {
```
prudhvi, line 117: Not sure if I should similar type setup for high watermark in checker.go

morgo: No I think this is fine. Good fix!

morgo (approved):

PR #415, opened 2025-05-09
Add logging and increase timeout of BlockWait calls

morgo:

Fixes https://github.com/cashapp/spirit/issues/414 -- or at least I believe so.

The timeout of 10s does not appear to be well suited for this wait, as it can take significant time to catch up. In future we may want to make this timeout configurable - but for now I've increased it to 30seconds and added logging.

If we do make it configurable:
- From within Flush() calls, we can wait virtually unlimited time when calling BlockWait() because this is not performed under a lock.
- Form within FlushUnderLock() calls, we will have to keep it short, and bail if we are not able to catch up. This keeps inline with our defaults being moderately safe.

### DIFF
--- a/pkg/repl/client.go
+++ b/pkg/repl/client.go
@@ -41,7 +41,7 @@ const (
 	// I expect we will change this to 1hr-24hr in the future.
 	DefaultFlushInterval = 30 * time.Second
 	// DefaultTimeout is how long BlockWait is supposed to wait before returning errors.
-	DefaultTimeout = 10 * time.Second
+	DefaultTimeout = 30 * time.Second
 )
 
 type Client struct {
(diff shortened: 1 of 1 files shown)


pkg/repl/client.go
```diff
@@ -562,23 +564,23 @@ func (c *Client) BlockWait(ctx context.Context) error {
...
+			return fmt.Errorf("timed out waiting to catch up to source position: %v, current position is: %v", targetPos, c.getBufferedPos())
+		default:
+			if err := dbconn.Exec(ctx, c.db, "FLUSH BINARY LOGS"); err != nil {
+				break // error flushing binary logs
+			}
+			if c.getBufferedPos().Compare(targetPos) >= 0 {
+				return nil // we are up to date!
+			}
+			// We are not caught up yet, so we need to wait.
+			time.Sleep(100 * time.Millisecond)
```
morgo, line 581: This refactors the code to be the same as canal, which we used previously.

I don't think this was the cause of the timeouts, but when I was trying to debug this - this struck me as sensitive code, and the canal implementation is well tested.

prudhvi (approved):

Issue #421, opened 2025-07-25
[feature request] Support reading db credentials from ini file


brirams:
It'd be great to be able to pass spirit a path to an ini file instead of needing plain text credentials in a CLI, which has the downside of having those credentials visible on a host's processlist. 

Happy to lend a hand in implementation if it's something worthwhile for the project at large. Thanks!

morgo:
We're interested in this feature (although we wouldn't use it ourselves due to how we automate spirit migrations).

I don't think it would be complicated to implement. We use [kong](https://github.com/alecthomas/kong) as the option parser, which can be extended to support config files.

brirams:
found some downtime to give it a shot: https://github.com/block/spirit/pull/422

PR #422, opened 2025-07-31
[migration] Add support for loading credentials from file provided in command line

brirams:
## Summary

Fixes https://github.com/block/spirit/issues/421. This PR adds support for loading host, port, username, password,  database, tls mode and tls certificate path from a `--conf` file argument to spirit. It expects the path to exist, as enforced by kong's `existingfile` type annotation and use `go-ini` to load credentials from the `client` section. Adds test to verify basic functionality.

## Changes

- `Migration` now has a `Conf` field, which is populated by kong using the `--conf` argument. I chose a name that seemed natural to me but open to any better suggestions. I
- `normalizeOptions` loads the file using go-ini and checks for a `client` section and uses the mysql standard `host`, `user`, `password`, `database` and `port` names for these arguments . Using `tls-*` instead of MySQL's `ssl-*` for tls settings. Args for `--host`, `--username`, `--password`, `--database`, `--tls-ca` and `--tls-mode` take precedence over what's in file
- updated USAGE to explain this functionality

## Testing

- new and existing unit tests pass
- built and run with expected results:
```
❯ ./spirit --help
Usage: spirit

Flags:
  -h, --help                                                           Show context-sensitive help.
      --host=STRING                                                    Hostname
      --username=STRING                                                User
      --password=PASSWORD                                              Password
      --database=STRING                                                Database
      --conf=STRING                                                    MySQL conf file
      --table=STRING                                                   Table
      --alter=STRING                                                   The alter statement to run on the table
      --threads=4                                                      Number of concurrent threads for copy and checksum tasks
      --target-chunk-time=500ms                                        The target copy time for each chunk
      --replica-dsn=STRING                                             A DSN for a replica which (if specified) will be used for lag checking.
      --replica-max-lag=120s                                           The maximum lag allowed on the replica before the migration throttles.
      --lock-wait-timeout=30s                                          The DDL lock_wait_timeout required for checksum and cutover
      --skip-drop-after-cutover                                        Keep old table after completing cutover
      --defer-cutover                                                  Defer cutover (and checksum) until sentinel table is dropped
      --force-kill                                                     Kill long-running transactions in order to acquire metadata lock (MDL) at checksum and cutover time
      --strict                                                         Exit on --alter mismatch when incomplete migration is detected
      --statement=""                                                   The SQL statement to run (replaces --table and --alter)
      --tls-mode=STRING                                                TLS connection mode (case insensitive): DISABLED, PREFERRED (default), REQUIRED, VERIFY_CA, VERIFY_IDENTITY
      --tls-ca=STRING                                                  Path to custom TLS CA certificate file
      --enable-experimental-buffered-copy                              Use the experimental buffered copier/repl applier based on the DBLog algorithm
      --enable-experimental-linting                                    Enable experimental linting checks before running migration
      --enable-experimental-linters=ENABLE-EXPERIMENTAL-LINTERS,...    Experimental linters to enable (default "all")
      --experimental-linter-config=EXPERIMENTAL-LINTER-CONFIG,...      Configuration options for experimental linters in the form linter_name.key=value
      --experimental-lint-only                                         Exit after executing linters
❯ cat my.cnf
[client]
user=some-user
password=super-secret
database=bramos_test
host=localhost
port=3306
❯ ./spirit --conf my.cnf --table testing --alter  'DROP  KEY idx_name_unique'
INFO[0000] Starting spirit migration: concurrency=4 target-chunk-size=500ms table='bramos_test.testing' alter=DROP  KEY idx_name_unique
INFO[0000] attempting to acquire metadata lock
INFO[0000] acquired metadata lock: bramos_test.testing-46cd71b3
INFO[0000] apply complete: instant-ddl=false inplace-ddl=true
INFO[0000] releasing metadata lock: bramos_test.testing-46cd71b3
❯ ./spirit --host localhost:3306 --database test --conf /var/log/nonexistent --table test_table --alter "ADD COLUMN other_json_column_again_5 varchar(100)"
spirit: error: --conf: stat /var/log/nonexistent: no such file or directory
```

### DIFF
--- a/pkg/migration/migration.go
+++ b/pkg/migration/migration.go
@@ -181,8 +181,8 @@ func (m *Migration) normalizeConnectionOptions() error {
 // nil or parameter is not defined.
 type confParams struct {
 	host, database, user, tlsMode, tlsCA string
-	password *string
-	port                           int
+	password                             *string
+	port                                 int
 }
 
 func (c *confParams) GetHost() string {
@@ -257,12 +257,12 @@ func newConfParams(confFilePath string) (*confParams, error) {
 
 	if creds.HasSection("client") {
 		clientSection := creds.Section("client")
-		confParams.host     = clientSection.Key("host").String()
+		confParams.host = clientSection.Key("host").String()
 		confParams.database = clientSection.Key("database").String()
-		confParams.user     = clientSection.Key("user").String()
-		confParams.tlsMode  = clientSection.Key("tls-mode").String()
-		confParams.tlsCA    = clientSection.Key("tls-ca").String()
-		confParams.port     = clientSection.Key("port").MustInt()
+		confParams.user = clientSection.Key("user").String()
+		confParams.tlsMode = clientSection.Key("tls-mode").String()
+		confParams.tlsCA = clientSection.Key("tls-ca").String()
+		confParams.port = clientSection.Key("port").MustInt()
 
 		if clientSection.HasKey("password") {
 			pw := clientSection.Key("password").String()
--- a/pkg/migration/migration_test.go
+++ b/pkg/migration/migration_test.go
@@ -628,13 +628,13 @@ func TestMigrationParamsDefaultsUsed(t *testing.T) {
 
 func TestMigrationParamsCLIUsed(t *testing.T) {
 	migration := &Migration{
-		Host:     "cli-host:3306",
-		Username: "cli-user",
-		Password: mkPtr("cli-password"),
-		Database: "cli-db",
-		Table:    "testtable",
-		Alter:    "ENGINE=InnoDB",
-		TLSMode:  "VERIFY_CA",
+		Host:               "cli-host:3306",
+		Username:           "cli-user",
+		Password:           mkPtr("cli-password"),
+		Database:           "cli-db",
+		Table:              "testtable",
+		Alter:              "ENGINE=InnoDB",
+		TLSMode:            "VERIFY_CA",
 		TLSCertificatePath: "/path/to/ca",
 	}
 
@@ -652,15 +652,15 @@ func TestMigrationParamsCLIUsed(t *testing.T) {
 func TestMigrationParamsEmptyPasswordUsedIfProvided(t *testing.T) {
 	migration := &Migration{
 		Password: mkPtr(""),
-		Table: "test_table",
-		Alter: "ENGINE=INNODB",
+		Table:    "test_table",
+		Alter:    "ENGINE=INNODB",
 	}
 
 	_, err := migration.normalizeOptions()
 	require.NoError(t, err)
 
 	assert.Equal(t, defaultUsername, migration.Username)
-	assert.Equal(t, "", *migration.Password)
+	assert.Empty(t, *migration.Password)
 	assert.Equal(t, fmt.Sprintf("%s:%d", defaultHost, defaultPort), migration.Host)
 	assert.Equal(t, defaultDatabase, migration.Database)
 	assert.Equal(t, defaultTLSMode, migration.TLSMode)
@@ -697,14 +697,14 @@ tls-ca = /path/from/file
 	require.NoError(t, tmpFile.Close())
 
 	migration := &Migration{
-		Host:     "cli-host:1234",
-		Username: "cli-user",
-		Password: mkPtr("cli-password"),
-		Database: "cli-db",
-		Table:    "testtable",
-		Alter:    "ENGINE=InnoDB",
-		ConfFile: tmpFile.Name(),
-		TLSMode:  "REQUIRED",
+		Host:               "cli-host:1234",
+		Username:           "cli-user",
+		Password:           mkPtr("cli-password"),
+		Database:           "cli-db",
+		Table:              "testtable",
+		Alter:              "ENGINE=InnoDB",
+		ConfFile:           tmpFile.Name(),
+		TLSMode:            "REQUIRED",
 		TLSCertificatePath: "/path/to/cert",
 	}
 
@@ -773,7 +773,7 @@ tls-ca = /path/to/another/ca
 	assert.Equal(t, "fileuser", migration.Username)
 	assert.Equal(t, "filepass", *migration.Password)
 	assert.Equal(t, "filehost:3306", migration.Host)
-	assert.Equal(t, "filedb",  migration.Database)
+	assert.Equal(t, "filedb", migration.Database)
 	assert.Equal(t, "VERIFY_IDENTITY", migration.TLSMode)
 	assert.Equal(t, "/path/to/another/ca", migration.TLSCertificatePath)
 }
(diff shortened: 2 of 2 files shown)


morgo:
Thank you for working on this. I see from the tests that this parses a MySQL-like ini file (rather than kong file etc). I think that's a great level of integration - it wasn't what I had intended but I think works fine.

But I have the following suggestions:
- Parse `[client]` only (I think you are doing this), but document as such and give an example.
- It's not just username + password, but we should also consider reading out host + database name. Thus the only arguments required to spirit will be either a --table + --alter or a --statement.
- In terms of precedence, this might be tricky because I'm not sure I've tried to differentiate between a default kong has sent me, or a value. Ideally we can differentiate and then arguments specified take precedence if both exist. This would also make it possible to auto-parse `~/.my.cnf` if it exists.

brirams:
> I see from the tests that this parses a MySQL-like ini file (rather than kong file etc). I think that's a great level of integration - it wasn't what I had intended but I think works fine.

ha! i completely missed that kong had the ability to parse an arbitrary config file but yeah -- keeping in line with MySQL ini files seems like a good thing to do for consistency with other tooling in the ecosystem

> Parse [client] only (I think you are doing this), but document as such and give an example.

yup -- we're only parsing the client section for these params

> It's not just username + password, but we should also consider reading out host + database name. Thus the only arguments required to spirit will be either a --table + --alter or a --statement.

host and database make sense: would it be reasonable, then, to also parse port from the config file, as it seems standard to have it defined in its own field in things like the MySQL CLI. I'd assume yes but want to confirm before going too far upfield

> In terms of precedence, this might be tricky because I'm not sure I've tried to differentiate between a default kong has sent me, or a value. Ideally we can differentiate and then arguments specified take precedence if both exist. This would also make it possible to auto-parse ~/.my.cnf if it exists.

so the easy answer here is to make `Host`,  `Username`, `Password` and `Database` be string pointers in the `Migration` struct and apply defaults for these fields in code instead of relying on kong's `default` annotation. The downside here is that, without the annotation, you lose the pretty printing of those values in the help output:

*current behavior*
```
❯ spirit -h
Usage: spirit

Flags:
  -h, --help                       Show context-sensitive help.
      --host="127.0.0.1:3306"      Hostname
      --username="msandbox"        User
      --password="msandbox"        Password
      --database="test"            Database
```

*without the default annotation*
```
❯ ./spirit -h
Usage: spirit

Flags:
  -h, --help                       Show context-sensitive help.
      --host=HOST                  Hostname
      --username=USERNAME          User
      --password=PASSWORD          Password
      --database=DATABASE          Database
```

I could also just shove the default in the help text but that also begs the question: do ya'll care for defaults for these four fields? we could  just remove them for these fields and fail when they're not provided in the command line or conf file. I _believe_ that's what part of the code actually checks for [here](https://github.com/block/spirit/blob/43b408215002b831738f4099d45f79d07219dab6/pkg/migration/migration.go#L69-L71) and [here](https://github.com/block/spirit/blob/43b408215002b831738f4099d45f79d07219dab6/pkg/migration/migration.go#L75-L77) but these were never true since the defaults injected by kong would also be non-empty.

let me know what you think and i can take another crack 🙇🏽

morgo:
> host and database make sense: would it be reasonable, then, to also parse port from the config file, as it seems standard to have it defined in its own field in things like the MySQL CLI. I'd assume yes but want to confirm before going too far upfield

Yes that's fine.

> I could also just shove the default in the help text but that also begs the question: do ya'll care for defaults for these four fields? we could just remove them for these fields and fail when they're not provided in the command line or conf file.

Yes, I agree that's the tradeoff here. The most straight forward way to implement this is to remove kong defaults.

We can later parse and substitute in defaults, but as you said it means they won't appear in help output. But I think that's fine.

pkg/migration/migration.go
```diff
@@ -69,15 +77,11 @@ func (m *Migration) normalizeOptions() (stmts []*statement.AbstractStatement, er
 	if m.ReplicaMaxLag == 0 {
 		m.ReplicaMaxLag = 120 * time.Second
 	}
-	if m.Host == "" {
```
brirams, line 85: this would have never been true as a normal course of operation since kong would inject a default at runtime

pkg/migration/migration.go
```diff
@@ -69,15 +77,11 @@ func (m *Migration) normalizeOptions() (stmts []*statement.AbstractStatement, er
...
-		return nil, errors.New("host is required")
-	}
-	if !strings.Contains(m.Host, ":") {
-		m.Host = fmt.Sprintf("%s:%d", m.Host, 3306)
-	}
-	if m.Database == "" {
```
brirams, line 91: same goes here: kong would inject the default of `test`.

pkg/migration/migration.go
morgo: This covers a use case we have internally. We use spirit as a library/package and don't rely on kong to inject values.

I think it's fine to remove though, we can find another way to handle this.

pkg/migration/migration.go
```diff
@@ -11,18 +11,26 @@ import (
...
 
 type Migration struct {
-	Host                 string        `name:"host" help:"Hostname" optional:"" default:"127.0.0.1:3306"`
-	Username             string        `name:"username" help:"User" optional:"" default:"spirit"`
-	Password             string        `name:"password" help:"Password" optional:"" default:"spirit"`
-	Database             string        `name:"database" help:"Database" optional:"" default:"test"`
+	Host                 *string       `name:"host" help:"Hostname" optional:""`
+	Username             *string       `name:"username" help:"User" optional:""`
+	Password             *string       `name:"password" help:"Password" optional:""`
+	Database             *string       `name:"database" help:"Database" optional:""`
```
morgo: I'm sure there's a reason, but can you explain the `*string` change? Is it to differentiate between not set, and set to empty?

If we can avoid it, that would be ideal of course.

brirams: ah -- so this was a carry over of trying to keep the default values within kong's purview which ended up not working out and then eventually convincing myself that it was cleaner to represent a value not being passed in as a nil pointer than an empty string. Happy to keep these as strings to make the PR smaller. lmk!

morgo: Yes, I think it's better as strings. If you can make this change back, I think this PR is much smaller/we don't have to consider the side effects here from `*string`.

brirams: question: do you still want to support passing the literal string "" as a password on the CLI and file? I could see that being a legitimate use-case for password test databases. if so, that would necessitate switching just that field to be a pointer since we'd need to make the distinction between "" and not being provided at all.

brirams:
> If you get a chance to add support for the new TLS options from https://github.com/block/spirit/pull/443 , I think this further supports the use case of why config files are handy.

i'll take a look!

brirams:
an interesting thing to note: the parameters for `tls-ca` and `tls-mode` are [`ssl-ca` and `ssl-mode` in mysql8](https://dev.mysql.com/doc/refman/8.4/en/using-encrypted-connections.html#using-encrypted-connections-client-side-configuration) so this won't be fully compatible with conf files used by plain vanilla mysql clients

pkg/migration/migration.go
morgo: Good point. Yes, I guess we need to support passing an empty string for the password.

brirams: cool -- i'll update to support that

PR #428, opened 2025-08-20
Add force kill functionality

### LINKED ISSUE #425: Add force-kill option during cutover for other connections
We've historically resisted this feature, because it indicates problems introduced elsewhere..

But we've seen some use cases where we have to complete a schema change, and there is no way the metadata lock will be acquired unless other connections holding it are killed first.

This is a FR to implement `force-kill` as an option. I am open to either a boolean (and basing it on some percentage of lock wait timeout) or a timeout setting. I haven't put too much thought into which is better, and will leave it to the implementer to propose.

If it is a timeout, please validate options to make sure that it makes sense in combination with [`lock-wait-timeout`](https://github.com/block/spirit/blob/main/USAGE.md#lock-wait-timeout). There will be some combinations which just don't make sense.

kolbe:

Closes #425.

* Adds a `force-kill` command-line boolean option (default false)
* Spirit aborts if it detects that another connection holds an explicit (`LOCK TABLES`) lock on any table Spirit is working with
* Spirit kills *connections* with transactions that have locks on tables Spirit is working with
* Transactions running longer than Spirit's configured locki-wait-time are eligible to be killed
* If any open transaction has a sufficiently high "weight" (currently set as >1m rows modified), Spirit will abort rather than kill the connection and risk an expensive rollback

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -80,7 +80,7 @@ When set to TRUE, Spirit will aggressively try to kill connections that are bloc
 
 Enabling the `force-kill` option requires spirit to be granted additional privileges:
 
-GRANT XX ON performance_schema.* TO spirituser;
+GRANT SELECT ON performance_schema.* TO spirituser;
 GRANT CONNECTION_ADMIN, PROCESS ON *.* TO spirituser;
 
 
--- a/pkg/dbconn/tablelock.go
+++ b/pkg/dbconn/tablelock.go
@@ -62,7 +62,6 @@ func NewTableLock(ctx context.Context, db *sql.DB, tables []*table.TableInfo, co
 					err := KillLockingTransactions(ctx, db, tables, config, logger, []int{pid})
 					if err != nil {
 						logger.Errorf("failed to kill locking transactions: %v", err)
-						isFatal = false
 					}
 				})
 				defer timer.Stop()


USAGE.md
```diff
@@ -63,7 +64,14 @@ Note that the checksum, if enabled, will be computed after the sentinel table is
...
+Even when force-inplace is `FALSE`, Spirit automatically detects "safe" operations that use the `INPLACE` algorithm. These include operations that modify only metadata, specifically `ALTER INDEX .. VISIBLE/INVISIBLE`, `DROP KEY/INDEX` and `RENAME KEY/INDEX`. Consult <https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html> for more details.
+
+### force-kill
+
+- Type: Boolean
+- Default value: FALSE
+
+When set to TRUE, Spirit will find other connections that hold open locks that block Spirit from acquiring a table lock during checksum and cutover. If any other connection holds an explicit table lock (LOCK TABLES) on a table Spirit needs to lock, Spirit will abort, because tools that acquire this kind of lock will probably not handle a killed connection very well. If another connection holds a transactional lock on a table Spirit needs to lock, Spirit will kill the connection only if the lock has been held for longer than 90% of our value of `lock-wait-timeout` (for example, 27 seconds if `lock-wait-timeout` is set to 30 seconds) _and_ the transaction's "weight" (the number of rows modified) is less than 1 million.
```
morgo: I think the nuances are easier to read in bullet points. Let me take a shot:

```suggestion
When set to TRUE, Spirit will aggressively try to kill connections that are blocking the checksum or cutover process from starting. It does this in a semi-intelligent way:

- It will read `performance_schema` to find only connections that are blocking a meta data lock being acquired on the migrating table.
- It refuses to kill connections if they have a transaction open that has modified a large number of rows (>1 million).
- It refuses to kill connections that hold an explicit `LOCK TABLE`, since unlike transactions these are not always retryable.
- It only starts killing transactions as it approaches the `lock-wait-timeout`. For example, if the `lock-wait-timeout` is 30 seconds, it will start killing transactions after 27 seconds.

Enabling the `force-kill` option requires spirit to be granted additional privileges:

GRANT XX ON performance_schema.* TO spirituser;
GRANT CONNECTION_ADMIN, PROCESS ON *.* TO spirituser;

```

pkg/dbconn/tablelock.go
```diff
@@ -67,6 +67,16 @@ func NewTableLock(ctx context.Context, db *sql.DB, tables []*table.TableInfo, co
...
+		if config.ForceKill {
+			// If we are using hostile MDL, we can try to acquire the lock again
+			// without rolling back the transaction.
+			logger.Warnf("failed to get lock (%v); trying to kill long running transactions", err)
+			err := KillLongRunningTransactions(ctx, db, tables, config, logger)
+			if err != nil {
+				logger.Errorf("failed to kill long-running transactions: %v", err)
+				return nil, err
+			}
+		}
```
morgo: Where's the @ 90% timer kicking in here? It looks like it's just killing @ 100% after the lock tables statement has timed out.

I am expecting line 50 to be in a goroutine now, and a timer is set to kill if the go routine hasn't succeeded at 27s.

kolbe: Sorry, I badly misunderstood your earlier descriptions of how you wanted this to work. Please take another look at my subsequent commits.

morgo: I can see that it now does the @ 90%:
```
		if config.ForceKill {
				// If ForceKill is true, we will wait for 90% of the configured LockWaitTimeout
				threshold := time.Duration(float64(config.LockWaitTimeout)*lockWaitTimeoutForceKillMultiplier) * time.Second
				timer := time.AfterFunc(threshold, func() {
					logger.Warnf("waited for %v; trying to kill locking transactions", threshold)
					err := KillLockingTransactions(ctx, db, tables, config, logger, []int{pid})
					if err != nil {
						logger.Errorf("failed to kill locking transactions: %v", err)
					}
				})
				defer timer.Stop()
			}
```

But what happens in the case that it was successful in acquiring the lock in less than this? It doesn't look like it skips killing. That is unless I misunderstand how `AfterFunc` works.

I think it needs a boolean or something for `success`, and then in the `afterFunc` check for !success.

kolbe: If the table locks are acquired successfully, the function will exit and the deferred `timer.Stop()` will execute, which means the timer will never fire and will not execute the function.

morgo: I see! LGTM then.

Issue #430, opened 2025-08-22
Refactor chunkers to be embeddable in a multi chunker


morgo:
This is an issue to refactor the chunker interface so that:

- When you create a copier, you specify the chunker as input (not table, newTable).
- Progress() gets moved from the copier to the chunker interface.
- The copier gets table, newTable directly from the chunk. It becomes dumber in that its role is to just call Next() until done and manage parallelism.

Thus, we will then be able to create a new MultiChunker which implements the chunker interface and have many underlying chunkers. Next() is routed to the chunker which has made the least progress.

PR #431, opened 2025-08-22
Refactor copier/chunkers for multi chunker

morgo:

Fixes https://github.com/block/spirit/issues/430

**Note:** the checksum process still knows about newTable and table. I will fix it in a followup PR, since this still needs more refactoring (introducing a multiChunker).


pkg/testutils/testing.go
```diff
@@ -13,7 +13,7 @@ import (
 func DSN() string {
 	dsn := os.Getenv("MYSQL_DSN")
 	if dsn == "" {
-		return "msandbox:msandbox@tcp(127.0.0.1:8030)/test"
+		return "msandbox:msandbox@tcp(127.0.0.1:3306)/test"
```
morgo: This is unrelated, but I think we can just change this, it was tied to an old sandbox version.

pkg/row/copier_test.go
```diff
@@ -300,78 +325,11 @@ func TestCopierValidation(t *testing.T) {
...
 }
 
-func TestETA(t *testing.T) {
-	testutils.RunSQL(t, "DROP TABLE IF EXISTS testeta1, testeta2, _testeta1_new, _testeta2_new")
-	testutils.RunSQL(t, "CREATE TABLE testeta1 (a INT NOT NULL, b INT, c INT, PRIMARY KEY (a))")
-	testutils.RunSQL(t, "CREATE TABLE testeta2 (a INT NOT NULL auto_increment, b INT, c INT, PRIMARY KEY (a))")
-	testutils.RunSQL(t, "CREATE TABLE _testeta1_new (a INT NOT NULL, b INT, c INT, PRIMARY KEY (a))")
-	testutils.RunSQL(t, "CREATE TABLE _testeta2_new (a INT NOT NULL auto_increment, b INT, c INT, PRIMARY KEY (a))")
-	// high max value
-	testutils.RunSQL(t, "INSERT IGNORE INTO testeta2 VALUES (10000, 2, 3)")
```
morgo, line 317: The copier is now no longer responsible for this. I've removed the test for now, but I may need to add more tests to replace it once I add the multi chunker.

pkg/testutils/testing.go
kolbe: Do you still use sandbox for your testing? Can we also just change the default to use `root`?

morgo: I do still use it for testing, but yes we can change it.

pkg/table/chunker_optimistic.go
```diff
@@ -205,11 +226,20 @@ func (t *chunkerOptimistic) Close() error {
...
 	t.Lock()
 	defer t.Unlock()
 	t.bumpWatermark(chunk)
 
+	// It is up to the chunker implementation to decide how to track "rows copied"
+	// In the optimistic chunker, since it is really designed around auto_increment
+	// tables, we add the ChunkSize to the rowsCopied counter, and ignore
+	// the actualRows. This differs from the composite chunker, which doesn't have an
+	// auto_inc max so it takes the table estimate and compares it to the actual
+	// rows copied.
```
prudhvi, line 239: Ah for progress tracking we just estimate based on chunk boundaries and max value instead of actual number of rows being copied?

pkg/table/chunker.go
```diff
@@ -30,16 +30,21 @@ const (
...
+	Feedback(chunk *Chunk, duration time.Duration, actualRows uint64)
 	GetLowWatermark() (string, error)
 	KeyAboveHighWatermark(key any) bool
+	Progress() (uint64, uint64, uint64) // Returns (rowsRead, chunksCopied, totalRowsExpected)
 }
 
-func NewChunker(t *TableInfo, chunkerTarget time.Duration, logger loggers.Advanced) (Chunker, error) {
+func newChunker(t *TableInfo, chunkerTarget time.Duration, logger loggers.Advanced) (Chunker, error) {
+	return NewChunker(t, nil, chunkerTarget, logger)
+}
```
prudhvi, line 45: I assume this is mainly for tests where we test all other chunker functionality that doesn't rely on destination?

pkg/table/chunker_optimistic.go
morgo: For the optimistic chunker yes:
- Highest auto-inc number
- Progress is based on how close we are to that number.

pkg/table/chunker.go
morgo: Yes that's right, the tests require it. I may remove this in the future, but just trying to keep the lines down for now.

Issue #451, opened 2025-09-19
Experiment with 'buffered' specific algorithm


morgo:
Currently spirit works like this:

- It starts a replication client to subscribe to table changes (tracking the PK only).
- It then progressively copies from the table using `INSERT IGNORE .. SELECT` ([src](https://github.com/block/spirit/blob/551860682b1eb31930e048e796133f22b449cd15/pkg/row/copier.go#L102-L109))
- Progressively the replication client changes are flushed using `REPLACE INTO .. SELECT` and `DELETE` statements ([src](https://github.com/block/spirit/blob/551860682b1eb31930e048e796133f22b449cd15/pkg/repl/subscription.go#L66-L97))

I propose we experiment with a modified algorithm, which I am calling _the Buffered_ algorithm. I don't know if it will be faster, but I think it's worth creating a proof of concept to try it. Here's how it works:

- The replication client starts, but rather than tracking just the PK, it buffers the whole row.
- The copy reads from the table using much larger chunks (10-20s instead of 2s) and fills a buffer.
- The copy buffer is then flushed in 1-2s writes, but at a higher parallel factor.
- The replication client flushes conditionally. It has to check that the key is below the low watermark of the copier. This ensures that there is no race from the copy being split into two parts and there being a lost update (the race is very brief: as an alternative we could rely on the checksum to repair issues, but I think it's better to use it as a back-stop).

Why I think this has merit:

- (for Aurora specifically) The longer reads trigger Aurora's read ahead which helps a lot when the data is not in cache. We can also use longer reads, because they won't be locking.
- The increased parallelism writes make good use of group commit, which Aurora also really benefits from. This is particularly beneficial because the writes are basically guaranteed to be disjoint sets.
- (for MySQL) the medium sized batches are more likely to be optimal for multi-threaded replicas. I believe (if I am paraphrasing JFGs research correctly) that disjoint small batches work best.
- This algorithm can also be used in other contexts, such as ShardMerge/MoveTables. These `INSERT.. SELECT` / `REPLACE .. SELECT` statements are the only thing in spirit tying it to a single server.

I call it the 'buffered' algorithm because this is the most distinctive difference from the existing algorithms. I previously considered calling this the 'Aurora' algorithm, but upon reflection I think that's not entirely true.

It's not too much of a code change to implement:
- Because we technically already have 2 methods to track replication changes (deltaMap and deltaQueue) we should extract this to an interface.
[...]

PR #452, opened 2025-09-22
repl: refactor subscription to interface

morgo:

Part of https://github.com/block/spirit/issues/451

This is a function neutral refactor. It looks like a lot of code is changed, but it really hasn't, it just moves around so that deltaMap/deltaQueue are split and we install an interface for subscription.


pkg/repl/client.go
```diff
@@ -145,10 +145,23 @@ func (c *Client) AddSubscription(currentTable, newTable *table.TableInfo, keyAbo
 		return fmt.Errorf("subscription already exists for table %s.%s", currentTable.SchemaName, currentTable.TableName)
 	}
 
-	c.subscriptions[subKey] = &subscription{
+	// Decide which subscription type to use. We always prefer deltaMap
+	// But will fall back to deltaQueue if the PK is not memory comparable.
+	if err := currentTable.PrimaryKeyIsMemoryComparable(); err == nil { // note: checking err is nil!
```
kolbe: Not a big deal, but why not just swap the handling so that you do a more idiomatic `err != nil` ?

pkg/repl/subscription.go
```diff
@@ -2,243 +2,22 @@ package repl
...
-	}
-	return s.flushDeltaMap(ctx, underLock, lock)
-}
-
-// flushDeltaQueue flushes the FIFO queue that is used when the PRIMARY KEY
-// is not memory comparable. It needs to be single threaded,
-// so it might not scale as well as the Delta Map, but offering
-// it at least helps improve compatibility.
+// Subscription is now an interface! It defines how the replication changes are tracked.
+// We ostensibly have two strategies:
```
kolbe: ostensibly: "apparently or purportedly, but perhaps not actually."

PR #453, opened 2025-09-22
row/copier: rename and refactor to interface

morgo:

Part of https://github.com/block/spirit/issues/451 (But useful regardless since it improves code-quality)

This is the same change as https://github.com/block/spirit/pull/452 but for the copier.

### DIFF
--- a/pkg/copier/copier.go
+++ b/pkg/copier/copier.go
@@ -1,0 +1,94 @@
+package copier
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"time"
+
+	"github.com/block/spirit/pkg/dbconn"
+	"github.com/block/spirit/pkg/metrics"
+	"github.com/block/spirit/pkg/table"
+	"github.com/block/spirit/pkg/throttler"
+	"github.com/siddontang/go-log/loggers"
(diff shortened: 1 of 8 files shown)


pkg/copier/copier.go
```diff
@@ -0,0 +1,94 @@
...
+type Copier interface { //nolint: interfacebloat
+	Run(ctx context.Context) error
+	GetETA() string
+	GetChunker() table.Chunker
+	SetThrottler(throttler throttler.Throttler)
+	GetThrottler() throttler.Throttler
+	StartTime() time.Time
+	GetProgress() string
+
+	// The following are for testing purposes only
```
kolbe, line 39: I think these could be unexported if they're only used for tests in this package *and* if you don't ever intend to implement this interface in another package.

kolbe (approved):

pkg/copier/copier.go
morgo: They are used for tests in the migration package. But what I could do is export them in one of the copiers but not make them part of the interface. I'm not sure if that's a good idea or not, I think it will be clearer once we add another copier.

PR #456, opened 2025-09-25
adding toc to the usage doc

Kiran01bm:
## What and Why ?
Adding TOC to the usage doc - makes it a bit easy to read.


morgo (approved):
LGTM, lets do it.

Issue #463, opened 2025-09-29
Implement Move command based off buffered algorithm


morgo:
This will have some initial restrictions:

- No Checkpoint / resume
- No Defer secondary index creation
- No Pluggable cutover
- Does not support schema changes / new tables introduced mid flight

PR #467, opened 2025-09-29
Add experimental move command

morgo:

Fixes https://github.com/block/spirit/issues/463

What's missing:
- Checkpoint / resume
- Defer secondary index creation
- Pluggable cutover
- Does not support schema changes / new tables introduced mid flight

### DIFF
--- a/pkg/check/visibility_change.go
+++ b/pkg/check/visibility_change.go
@@ -2,22 +2,71 @@ package check
 
 import (
 	"context"
+	"errors"
 
+	"github.com/block/spirit/pkg/statement"
+	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/parser/mysql"
 	"github.com/siddontang/loggers"
 )
+
+// ErrVisibilityMixedWithOtherChanges is returned when index visibility changes are mixed with table-rebuilding operations
+var ErrVisibilityMixedWithOtherChanges = errors.New("the ALTER operation contains a change to index visibility mixed with table-rebuilding operations. This creates semantic issues for experiments. Please split the ALTER statement into separate statements for changing the invisible index and other operations")
 
 func init() {
 	registerCheck("visibility", visibilityCheck, ScopePostSetup)
 }
 
-// We don't want to allow visibility changes
-// This is because we've already attempted MySQL DDL as INPLACE, and it didn't work.
+// visibilityCheck validates index visibility changes in ALTER statements.
+// It allows index visibility changes when mixed with other metadata-only operations,
+// but blocks them when mixed with table-rebuilding operations to avoid semantic issues.
 // It likely means the user is combining this operation with other unsafe operations,
 // which is not a good idea. We need to protect them by not allowing it.
 // https://github.com/block/spirit/issues/283
-func visibilityCheck(_ context.Context, r Resources, _ loggers.Advanced) error {
-	if err := r.Statement.AlterContainsIndexVisibility(); err != nil {
-		return err
+func visibilityCheck(ctx context.Context, r Resources, logger loggers.Advanced) error {
+	return AlterContainsIndexVisibility(r.Statement)
+}
+
+// AlterContainsIndexVisibility checks to see if there are any clauses of an ALTER to change index visibility.
+// Allows index visibility changes when mixed with other metadata-only operations,
+// but blocks them when mixed with table-rebuilding operations to avoid semantic issues.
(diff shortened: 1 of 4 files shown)


pkg/move/runner.go
```diff
@@ -0,0 +1,280 @@
...
+func NewRunner(m *Move) (*Runner, error) {
+	r := &Runner{
+		move: m,
+	}
+	return r, nil
+}
+
+func (r *Runner) Close() error {
+	return nil
+}
```
aparajon, line 47: Is there any clean-up or logging we need to do here?

pkg/move/runner.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Start the replication client.
+	if err := r.replClient.Run(ctx); err != nil {
+		return err
+	}
+
+	// Run the copier.
+	if err := r.copier.Run(ctx); err != nil {
+		return err
+	}
```
aparajon, line 215: Do we log status / progress per table?

pkg/move/runner.go
```diff
@@ -0,0 +1,280 @@
...
+	return r, nil
+}
+
+func (r *Runner) Close() error {
+	return nil
+}
+
+// getSourceTables connects to r.source and fetches the list of tables
+// to populate r.sourceTables.
+func (r *Runner) getSourceTables(ctx context.Context) error {
```
aparajon, line 51: I suppose a future enhancement could be to provide an explicit list of tables to operate on from the `source`. This idea is inspired by the Vitess MoveTables `--tables` option, which I've seen used in practice to omit some tables from being moved

pkg/move/runner.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	// When the copier has finished, catch up the replication client
+	// This is in a non-blocking way first.
+	if err := r.replClient.Flush(ctx); err != nil {
+		return err
+	}
+
+	r.logger.Infof("All tables copied successfully.")
+
```
aparajon, line 224: If any of the table moves were to fail mid-flight, would the entire command / process stop? Or can there be a partial success state where some tables succeed and others fail?

pkg/move/move_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"database/sql"
+	"testing"
+	"time"
+
+	"github.com/block/spirit/pkg/testutils"
+	"github.com/go-sql-driver/mysql"
+	"github.com/stretchr/testify/assert"
+)
+
```
aparajon, line 12: Are there any failure scenarios that are worth testing here?

pkg/move/move_test.go
```diff
@@ -0,0 +1,50 @@
...
+	assert.NoError(t, err)
+	defer db.Close()
+	// test
+	move := &Move{
+		SourceDSN:       sourceDSN,
+		TargetDSN:       targetDSN,
+		TargetChunkTime: 5 * time.Second,
+		Threads:         4,
+	}
+	assert.NoError(t, move.Run())
```
aparajon, line 49: Should we also assert that the rows were copied to the target?

pkg/move/runner.go
morgo: Not yet. The copier does have a logger in its struct, but (comparing to table `migration`) the 30s log prints are external. This has not been implemented for move yet.

pkg/move/runner.go
morgo: Yes, for future improvement. Could we use two regexes? (include and ignore list).

pkg/move/runner.go
morgo: There is no checkpointing, and it doesn't currently support that there could be a schema change mid-flight either (I will add this one to the list of restrictions).

pkg/move/runner.go
morgo: Likely yes in future. Currently because most code is in Run() I am able to do cleanups in `defer`s.

pkg/move/move_test.go
morgo: Move.Run() does have a checksum, so it is validating in a sense the rows have made it there.

pkg/move/move_test.go
morgo: Yes (in future). This is experimental to start with, so I can start work planning how to integrate this with our orchestrator (tern) in parallel.

pkg/move/runner.go
aparajon: Yeah that would be nice

Issue #448, opened 2025-09-16
Acquire metadatalock for multi migrations


morgo:
Currently we are not acquiring an MDL lock (getLock) with multi migrations. This is mostly because in the mdl API it requires a table name.

~Maybe by convention we lock `_spirit_migration` or something?~ We can just lock each of the affected tables, it works fine.

Issue #441, opened 2025-08-27
Write extensive tests for multi-chunker


morgo:
The multi-chunker is pretty basic.

I think we can write some tests to check its behavior, particularly around Next() with multiple underlying chunkers. What likely makes the most sense is to implement another chunker, which is a mock implantation fo a chunker. This can make the tests entirely deterministic.

Issue #436, opened 2025-08-27
Implement checkpointing/resume for multi-chunker


morgo:
This is a feature we can implement after we get the basics right.

Checkpointing needs to be implemented slightly differently for multi-chunkers, since it doesn't make sense to have a bunch of individual checkpoints with different binary log positions for recovery.

It should be one checkpoint which is centralized (_spirit_chkpnt). We may as well do the same with the sentinel table.

PR #469, opened 2025-10-01
Add support for resume from checksum in multi chunker

morgo:

Fixes https://github.com/block/spirit/issues/448
Fixes https://github.com/block/spirit/issues/436
Fixes https://github.com/block/spirit/issues/441

This does change the checkpoint format, but that seems like the most logical way to implement this. Unfortunately this means that when upgrading from an earlier version, checkpoints will not be resumable.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -10,7 +10,6 @@
     - [checksum](#checksum)
     - [database](#database)
     - [defer-cutover](#defer-cutover)
-    - [force-inplace](#force-inplace)
     - [force-kill](#force-kill)
     - [host](#host)
     - [lock-wait-timeout](#lock-wait-timeout)
@@ -29,6 +28,8 @@
       - [VERIFY\_CA](#verify_ca)
       - [VERIFY\_IDENTITY](#verify_identity)
   - [Experimental Features](#experimental-features)
+    - [enable-experimental-multi-table-support](#enable-experimental-multi-table-support)
+    - [enable-experimental-buffered-copy](#enable-experimental-buffered-copy)
 
 ## Getting Started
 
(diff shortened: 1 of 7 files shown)


pkg/copier/copier_test.go
```diff
@@ -353,8 +353,7 @@ func TestCopierFromCheckpoint(t *testing.T) {
 	assert.NoError(t, err)
 
 	// Open chunker at the specified watermark
-	highPtr := table.NewDatum(t1new.MaxValue().Val, t1.MaxValue().Tp)
-	err = chunker.OpenAtWatermark(lowWatermark, highPtr, 0)
+	err = chunker.OpenAtWatermark(lowWatermark)
```
morgo, line 356: This has to be pushed down into the chunker, since the high pointer needs to be calculated per table.

pkg/migration/runner.go
```diff
@@ -670,8 +678,7 @@ func (r *Runner) createCheckpointTable(ctx context.Context) error {
 	checksum_watermark TEXT,
 	binlog_name VARCHAR(255),
 	binlog_pos INT,
-	rows_copied BIGINT,
-	alter_statement TEXT
```
morgo, line 674: The rows_copied can not apply to multi-migrations because we need to know a number per chunker, so this is pushed into the watermark.

The optimistic chunker does not need this number, since it can use the value based on the current chunkPtr. The composite chunker does need this, so it now writes that into its checkpoint. The checkpoints are technically different for all 3 chunkers as a result.

pkg/migration/runner.go
```diff
@@ -154,9 +161,14 @@ func (r *Runner) Run(originalCtx context.Context) error {
...
-	if !r.migration.EnableExperimentalMultiTableSupport {
-		// If it's not an alter table, that means it is a CREATE TABLE, DROP TABLE, or RENAME table.
-		// We should execute it immediately before acquiring TableInfo.
+	if r.migration.EnableExperimentalMultiTableSupport {
+		// We don't (yet) support a lot of features in multi-schema changes, and
+		// we never attempt instant/inplace DDL. So for now all we need to do
+		// is setup and call SetInfo on each of the tables.
+		r.logger.Warn("Enabling the experimental option: enable-experimental-multi-table-support")
+	} else {
+		// We only allow non-ALTERs (i.e. CREATE TABLE, DROP TABLE, RENAME TABLE)
```
kolbe, line 170: This is confusing. I guess you mean that we only directly execute non-ALTERs for single-table mode ("only allow" is really confusing to me). And these are handled elsewhere (where? why?) when multi-table is enabled?

pkg/migration/runner.go
```diff
@@ -117,6 +117,13 @@ func (r *Runner) SetLogger(logger loggers.Advanced) {
 	r.logger = logger
 }
 
+func (r *Runner) attemptMySQLDDL(ctx context.Context) error {
+	if r.migration.EnableExperimentalMultiTableSupport {
```
kolbe, line 121: could this instead check the len(r.changes) so that the checking is a bit more robust and doesn't rely on this option being around forever?

pkg/migration/runner.go
kolbe: Actually it seems like non-ALTERs just aren't supported at all in multi-table mode, is that right? I think we should urgently add some better messaging about that, but it doesn't need to block this PR.

morgo: Yes that's correct. The problem is we can only guarantee atomicity if they are all ALTER.

pkg/migration/runner.go
morgo: It could. I did it this way intentionally though, since I want to be able to test the mutli code path, even when there is just one change.

I was considering _proposing_ we enable it for staging. This has the downside that there are no instant changes, which may be considered a high cost.

PR #470, opened 2025-10-02
remove force-inplace and changes to automatic DDL algorithm detection

Kiran01bm:
## What and Why ?
This PR removes the manual `--force-inplace` flag from Spirit and replaces it with intelligent automatic DDL algorithm detection.

### **Safety Improvements**
- ✅ Prevents unsafe replica-blocking INPLACE operations
- ✅ Eliminates dangerous manual overrides  
- ✅ Provides clear error messages for unsafe combinations

### **Removed Manual Configuration**
- **Before**: Users had to manually decide when to use INPLACE with `--force-inplace` flag
- **After**: System automatically determines the safest approach with zero configuration

### **Enhanced Mixed Operations Support**
- **Index visibility + metadata operations**: Now automatically allowed and optimized
- **Index visibility + table-rebuilding**: Blocked with clear error messages for safety

### **Improved Safety**
- **ADD INDEX operations**: Use copy process to prevent replica blocking
- **DROP INDEX operations**: Use copy process for enhanced safety while allowing logical mixing with index visibility
- **Automatic algorithm selection**: Prevents unsafe combinations without user intervention

### **Classification of Operations**
```sql
-- ✅ Now Allowed (Enhanced Capabilities)
ALTER INDEX idx VISIBLE, DROP INDEX old_idx;        -- Uses copy (safe)
ALTER INDEX idx VISIBLE, RENAME INDEX old TO new;   -- Uses INPLACE  
ALTER INDEX idx VISIBLE, ADD PARTITION p1 VALUES LESS THAN (100);

-- ❌ Blocked (Enhanced Safety)
ALTER INDEX idx VISIBLE, ADD INDEX (col);           -- Semantic conflicts
ALTER INDEX idx VISIBLE, ADD COLUMN col INT;        -- Table rebuilding
```

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -5,12 +5,16 @@
 - [How to use Spirit](#how-to-use-spirit)
   - [Table of Contents](#table-of-contents)
   - [Getting Started](#getting-started)
+  - [Automatic DDL Algorithm Detection](#automatic-ddl-algorithm-detection)
+    - [How it Works](#how-it-works)
+    - [Safe INPLACE Operations](#safe-inplace-operations)
+    - [Enhanced Index Visibility Support](#enhanced-index-visibility-support)
+    - [Safety Guarantees](#safety-guarantees)
   - [Configuration](#configuration)
     - [alter](#alter)
     - [checksum](#checksum)
     - [database](#database)
     - [defer-cutover](#defer-cutover)
-    - [force-inplace](#force-inplace)
     - [force-kill](#force-kill)
     - [host](#host)
     - [lock-wait-timeout](#lock-wait-timeout)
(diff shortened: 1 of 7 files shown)


USAGE.md
```diff
@@ -89,14 +93,42 @@ If you start a migration and realize that you forgot to set defer-cutover, worry
...
-- Default value: FALSE
+Spirit automatically detects and uses the most appropriate DDL algorithm for your schema changes without requiring manual configuration. This intelligent detection system prioritizes safety while maximizing performance.
+
+### How it Works
+
+Spirit analyzes your ALTER statement and automatically chooses between three approaches:
+
+1. **INSTANT DDL**: For operations that can be completed instantly (e.g., adding columns with default values in MySQL 8.0+)
+2. **INPLACE DDL**: For metadata-only operations that are safe and non-blocking
+3. **Copy Process**: For complex operations that require data migration
```
morgo: I think this is not really correct. It doesn't analyze the statement fro INSTANT DDL, it just tries it, and falls back if it fails. The only one it analyzes is INPLACE.

I'm happy/prefer not to document this. The behavior does not affect usage (you use it all the same), and these are docs that will likely fail to be maintained as we make tweaks to what is a safe inplace.

pkg/migration/runner_test.go
```diff
@@ -240,20 +240,19 @@ func TestOnline(t *testing.T) {
...
+		Threads:  16,
+		Table:    "testonline3",
+		Alter:    "ADD INDEX(b)",
 	})
 	assert.NoError(t, err)
 	err = m.Run(t.Context())
 	assert.NoError(t, err)
 	assert.False(t, m.usedInstantDDL) // not possible
-	assert.True(t, m.usedInplaceDDL)  // as
+	assert.False(t, m.usedInplaceDDL) // ADD INDEX operations now use copy process for replica safety
```
morgo:
```suggestion
	assert.False(t, m.usedInplaceDDL) // ADD INDEX operations now always require copy
```

The replica safety thing is true (when non Aurora), but neither here nor there. The behavior here is also strongly correlated to `ForceInplace`.

pkg/migration/runner_test.go
```diff
@@ -267,20 +266,19 @@ func TestOnline(t *testing.T) {
...
+		Threads:  16,
+		Table:    "testonline4",
+		Alter:    "drop index name, drop index b",
 	})
 	assert.NoError(t, err)
 	err = m.Run(t.Context())
 	assert.NoError(t, err)
 	assert.False(t, m.usedInstantDDL) // unfortunately false in 8.0, see https://bugs.mysql.com/bug.php?id=113355
-	assert.True(t, m.usedInplaceDDL)
+	assert.False(t, m.usedInplaceDDL) // DROP INDEX operations now use copy process for replica safety
```
morgo: This is a bug. It looks like below we've removed `DROP INDEX` from the safe inplaces. It's the most subjective of the safe inplaces (can take slightly longer than the others), but its still usually fast, and worth including in them.

pkg/statement/statement.go
```diff
@@ -158,8 +158,7 @@ func (a *AbstractStatement) AlgorithmInplaceConsideredSafe() error {
 	unsafeClauses := 0
 	for _, spec := range alterStmt.Specs {
 		switch spec.Tp {
-		case ast.AlterTableDropIndex,
```
morgo, line 161: Here is where it looks like DROP INDEX was dropped.

USAGE.md
Kiran01bm: sure thank you., wrong choice of words from my side (wrt analyze).. i'll remove them and yes i agree this is a moving target and will likely go stale..

pkg/migration/runner_test.go
Kiran01bm: thats a nice catch.. and fair pint - i've added `DROP INDEX` to safe in-place..

pkg/statement/statement.go
```diff
@@ -158,12 +158,12 @@ func (a *AbstractStatement) AlgorithmInplaceConsideredSafe() error {
...
-			ast.AlterTableRenameIndex,
+		case ast.AlterTableRenameIndex,
 			ast.AlterTableIndexInvisible,
 			ast.AlterTableDropPartition,
 			ast.AlterTableTruncatePartition,
-			ast.AlterTableAddPartitions:
+			ast.AlterTableAddPartitions,
+			ast.AlterTableDropIndex:
```
morgo, line 166: This looks like no change (except the order). You might want to change it back to simplify the diff.

pkg/statement/statement.go
```diff
@@ -172,15 +172,16 @@ func (a *AbstractStatement) AlgorithmInplaceConsideredSafe() error {
 			if spec.NewColumns[0].Tp != nil && spec.NewColumns[0].Tp.GetType() == mysql.TypeVarchar {
 				continue
 			}
+			unsafeClauses++
```
morgo, line 175: Was this a bug? It might be, I'm not quite understanding from the diff if we missed a branch where we were supposed to increment `unsafeClauses`.

pkg/statement/statement.go
```diff
@@ -227,20 +228,48 @@ func (a *AbstractStatement) AlterContainsAddUnique() error {
...
+			// All other operations are considered non-metadata (table rebuilding)
+			hasNonMetadataOperation = true
 		}
 	}
+
+	// Only fail if index visibility is mixed with non-metadata operations
+	if hasIndexVisibility && hasNonMetadataOperation {
+		return errors.New("the ALTER operation contains a change to index visibility mixed with table-rebuilding operations. This creates semantic issues for experiments. Please split the ALTER statement into separate statements for changing the invisible index and other operations")
+	}
+
```
morgo, line 272: This semantic change makes sense to me. I wonder if we can move it into the checks/ system though so we keep the flow of a migration simple.

We can handle it as a followup though.

morgo (approved):
LGTM, just a small question and a nit. Otherwise happy to merge.

pkg/statement/statement.go
morgo: I read through it, looks like a bug. Resolving this.

morgo:
On second thoughts, I'm happy with this. Let's merge.

pkg/statement/statement.go
Kiran01bm: sorry, i missed this comment when responding..

pkg/statement/statement.go
Kiran01bm: yep..

pkg/statement/statement.go
Kiran01bm: and a few other tests required fixing - all done

Issue #471, opened 2025-10-02
Ensure correct handling for JSON, RBR minimal, non memory comparable keys


morgo:
There are some edge cases we need to think about:

- JSON in MySQL 8.0 does not contain the full image in the binary log, thus if we encounter a JSON object we (likely) need to either handle the patching (we can do this, since we flush after passing the low watermark) or read the row from the source.
- Minimal RBR will not contain the full row images, we either need to refuse support or fallback to reads from the source.
- Non-memory comparable keys do not work safely with high/low watermarks. I think this algorithm really depends on watermarks, and this is quite rare, so my suggestion would be we require the checksum be enabled and catch any discrepancies there.

morgo:
See also https://github.com/block/spirit/issues/464

PR #473, opened 2025-10-06
Add boundary checking, improve buffered copier

morgo:

Fixes https://github.com/block/spirit/issues/471
Fixes https://github.com/block/spirit/issues/464

### DIFF
--- a/pkg/repl/client_test.go
+++ b/pkg/repl/client_test.go
@@ -661,8 +661,17 @@ func TestCompositePKUpdate(t *testing.T) {
 }
 
 func TestAllChangesFlushed(t *testing.T) {
-	srcTable, dstTable := setupTestTables(t)
-
+	t1 := `CREATE TABLE subscription_test (
+		id INT NOT NULL,
+		name VARCHAR(255) NOT NULL,
+		PRIMARY KEY (id)
+	)`
+	t2 := `CREATE TABLE _subscription_test_new (
+		id INT NOT NULL,
+		name VARCHAR(255) NOT NULL,
+		PRIMARY KEY (id)
+	)`
+	srcTable, dstTable := setupTestTables(t, t1, t2)
 	client := &Client{
 		db:              nil,
 		logger:          logrus.New(),
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -24,6 +24,7 @@ func HashKey(key []any) string {
 }
 
 // IntersectNonGeneratedColumns returns a string of columns that are in both tables
+// The column names are in backticks and comma separated.
 func IntersectNonGeneratedColumns(t1, t2 *table.TableInfo) string {
 	var intersection []string
 	for _, col := range t1.NonGeneratedColumns {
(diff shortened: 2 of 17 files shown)


pkg/utils/utils_test.go
```diff
@@ -53,3 +71,67 @@ func TestStripPort(t *testing.T) {
...
+
+	// Test JSON types
+	jsonBytes := []byte(`[1, 2, 3]`)
+	assert.Equal(t, "CONVERT('[1, 2, 3]' USING utf8mb4)", EscapeMySQLType("JSON", jsonBytes))
+	assert.Equal(t, "CONVERT('test' USING utf8mb4)", EscapeMySQLType("json", "test"))
+
+	// Test VARBINARY types
+	binaryData := []byte{0x01, 0x02, 0x03}
+	result := EscapeMySQLType("VARBINARY(255)", binaryData)
+	assert.Contains(t, result, "_binary")
```
kolbe, line 89: Isn't this redundant with the following assertion?

pkg/utils/utils.go
```diff
@@ -48,15 +62,91 @@ func UnhashKey(key string) string {
...
+
+	default:
+		// For other types, convert []byte to string and use default escaping
+		if b, ok := value.([]byte); ok {
+			// Convert to string first, then escape
+			return "'" + sqlescape.EscapeString(string(b)) + "'"
+		}
+		escaped, err := sqlescape.EscapeSQL("%?", value)
+		if err != nil {
+			return "NULL"
```
kolbe: Should this possibly do something more dramatic than just return NULL? Is there a danger of data loss? Do you just rely on checksum to work it out? Since this utility function might be used elsewhere, I think this should maybe be a panic.

pkg/repl/client_test.go
```diff
@@ -474,7 +474,26 @@ func TestDDLNotification(t *testing.T) {
...
-	tableModified := <-ddlNotifications
+	// Because tests across packages are run in parallel, this test
+	// is actually racey because another notification may be received instead.
+	// To counteract this we just expect *this* notification within 2 seconds
+	var tableModified string
+	timeout := time.After(2 * time.Second)
+
+	for {
+		select {
+		case notification := <-ddlNotifications:
```
kolbe, line 485: I guess if there's some other test watching for notifications, this will consume the notification and make the _other_ test fail?

pkg/utils/utils_test.go
morgo: No, what this does is checks that the collation hint is specified ahead of the string. So it doesn't use the connection's properties.

pkg/utils/utils.go
morgo: Yeah I agree, I'll fix it.

pkg/repl/client_test.go
morgo: Not the case. The replication client sends all DDL changes to the notification channel. Because the binary logs are shared, there is a race where it may occasionally get the DDL of another test. But the channels are not shared.

An alternative fix is that there could be filtering in the replication client to only send notifications for tables for which there are subscriptions. I didn't do this originally, but I could switch to it in future.

Issue #472, opened 2025-10-06
Untenably slow checksum and large number of mismatches


mcfadden:
We are migrating a table with has approximately 2B rows in it and is ~600GB of data (plus indices). The PK on the table is a compound key with three `bigint` keys.

The table has pretty high write volume, reaching between 500 and 1000 queries per second (cumulative across INSERT, UPDATE, and DELETE)

The migration we are performing is changing the character set on 3 fields - two are `varchar` and one is a `mediumtext` field. 

When running the migration, the copy phase completes successfully (in 16h18m) but we run into problems with the checksum. Immediately it started finding checksum mismatches:

```
inspection revealed row checksum mismatch for pk: 3,1557041100,3551579362: source 1256209406 != target 867087309
inspection revealed row checksum mismatch for pk: 3,1559149667,3556875582: source 488007175 != target 3404618948
inspection revealed row checksum mismatch for pk: 3,1559461267,3557876427: source 3316947326 != target 4191483632
inspection revealed row checksum mismatch for pk: 3,1559467322,3557902209: source 3533309006 != target 227376112
inspection revealed row checksum mismatch for pk: 3,1559467322,3557903036: source 3807221302 != target 2298420767
inspection revealed row checksum mismatch for pk: 3,1559469264,3557904301: source 1131887794 != target 4204702317
inspection revealed row checksum mismatch for pk: 3,1559558257,3558138009: source 1256415619 != target 2754777990
inspection revealed row checksum mismatch for pk: 3,1559643070,3558412353: source 4272772669 != target 4004342247
inspection revealed row checksum mismatch for pk: 3,1559651341,3558455736: source 3194258644 != target 1620073712
```

followed by recopying chunks of data.

There is also a large number of rows missing, ex:
`inspection revealed row does not exist in target for pk: 5,1559694394,3558567424`

The biggest problem here, however, is that it is _so slow_. We let the last iteration run in the checksum process for 11 hours, and it completed approximately 1.5% of table (based on `checksum-progress`, which in this case is at least _close enough_ of an estimate based on our ids). At that rate we'd be doing the checksum for a month - and it would have to run a second time I believe since it found mismatches.

I am not sure if it is that slow because it is having to recopy so much of the table, or if there is another reason for that?

I wondered if something strange happened the first run, so I reattempted the migration (a total of 3 times, actually) and every run through has had the same problem in the checksum phase.

I'm not sure where to look next or what make the most sense to try, but I am happy to put in some work here if there is any data that would be helpful.

[...]

Issue #475, opened 2025-10-09
Enable force-enable-buffered-map by default; remove option


morgo:
Currently if a key is non-memory comparable, we force the replication subscription to be a queue, and not the buffered map. This is considerably worse for performance for what is an obscure case (the character order does not match the binary comparable order).

We have added an option which is entirely safe *because* of the checksum: https://github.com/block/spirit/blob/main/docs/migrate.md#force-enable-buffered-map

There is no technical we can't enable this by default, and remove the option. The only reason I added the option is to reduce some of the massive changes being made at once. Once we have some confidence in stability, we should proceed forward on this path.

morgo:
~There's one problem I don't know a good answer for:~

~- Yes we can correct issues with the checksum~
~- But after the checksum runs there is a race where the last few modifications are not technically checksummed.~

~One possible solution is to switch back to a queue after the checksum, but I'm not sure that's very elegant.~

This is exactly what we've added: we switch to a queue automatically when the watermark optimizations are disabled, and it looks to be working fine.

PR #478, opened 2025-10-10
Fix flush correctness issue

morgo:

Fixes https://github.com/block/spirit/issues/472

This fixes a correctness issue where if after retries are exhausted the replication flush fails, pending changes will be discarded. This manifests in checksum errors, or in the absence of a checksum, lost updates (i.e. corruption).

This fix may have a performance impact, since the replication subscription mutex is now held for the duration of the Flush operation. I have future plans for how to lessen this impact:

1. We deprecate using the queue subscription mechanism, since it performs horribly https://github.com/block/spirit/issues/475
2. We implement keyAboveWatermark for the composite chunker (provided key[0] is memory comparable). https://github.com/block/spirit/issues/479
3. We implement wait for the change to be below the copier low watermark before attempting to flush a change https://github.com/block/spirit/issues/465
4. We adjust the flushing chunk size if contention is detected. https://github.com/block/spirit/issues/480

In the user's bug report, (1) is not applicable.

I had previously considered (4) as the ideal fix, but I am now leaning towards 2 & 3 more immediately, since I believe they will work sufficiently. We can re-evaluate (4) if they do not.

### DIFF
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -38,11 +38,13 @@ type Resources struct {
 	ForceKill            bool
 	// The following resources are only used by the
 	// pre-run checks
-	Host               string
-	Username           string
-	Password           string
-	TLSMode            string
-	TLSCertificatePath string
+	Host                     string
+	Username                 string
+	Password                 string
+	TLSMode                  string
+	TLSCertificatePath       string
+	ExperimentalBufferedCopy bool
+	Checksum                 bool
 }
 
 type check struct {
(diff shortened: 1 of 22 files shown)


pkg/repl/subscription_map.go
```diff
@@ -149,6 +147,10 @@ func (s *deltaMap) Flush(ctx context.Context, underLock bool, lock *dbconn.Table
 			return err
 		}
 	}
+	// If it's successful, we can clear the map
```
kolbe, line 150: What's the flow if it's *not* successful? The changes are executed in parallel, so if one of them fails the others will have finished, but they are all kept in the map. I guess executing them should be idempotent because we're either doing DELETE or REPLACE?

morgo: Yes that's right. Because its idempotent we can allow replay if any single one fails.

PR #484, opened 2025-10-12
refactor alter visibility to the checks sub-system

Kiran01bm:
## What and Why ?
Changes in this PR is to address a follow-up suggestion linked here - https://github.com/block/spirit/pull/470#discussion_r2401970759

**NOTE:**
1. All changes are mainly re-factoring/re-org type in nature (no-functional changes).
2. Sequential tests are now modified to a table-driven test suite (easier to read, expand and maintain).

### DIFF
--- a/pkg/check/visibility_change_test.go
+++ b/pkg/check/visibility_change_test.go
@@ -73,6 +73,10 @@ func TestVisibilityChange(t *testing.T) {
 		{"ALTER INDEX b VISIBLE, modify `a` varchar(100)", false, nil, "visibility + varchar modify"},
 		{"ALTER INDEX b INVISIBLE, change column `a` `a` varchar(150)", false, nil, "visibility + varchar change"},
 
+		// Complex mixed operations - multiple metadata operations with visibility
+		{"ALTER INDEX a INVISIBLE, rename index `b` to `new_b`, modify `col` varchar(100)", false, nil, "visibility + rename + varchar modify"},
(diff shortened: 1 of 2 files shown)


pkg/check/visibility_change.go
```diff
@@ -3,21 +3,33 @@ package check
...
+// AlterContainsIndexVisibility checks to see if there are any clauses of an ALTER to change index visibility.
+// Allows index visibility changes when mixed with other metadata-only operations,
+// but blocks them when mixed with table-rebuilding operations to avoid semantic issues.
+func AlterContainsIndexVisibility(stmt *statement.AbstractStatement) error {
+	alterStmt, ok := (*stmt.StmtNode).(*ast.AlterTableStmt)
+	if !ok {
+		return statement.ErrNotAlterTable
 	}
-	return nil
+	return validation.AlterContainsIndexVisibilityLogic(alterStmt)
```
morgo: Am I missing something? I don't think we need the third package:

- The logic can be moved here.
- The pkg/statement functionality can be removed, rather than redirected to `validation`.
- This allows cleanup in `migration/runner.go` since we don't need to call into `statement. AlterContainsIndexVisibility()` anymore. The check will automatically be done by the checks system.

Kiran01bm: okay, so my thinking was the validation package could potentially also house the other validation logic for example - `AlgorithmInplaceConsideredSafe` or `AlterContainsAddUnique` ...

if you prefer otherwise - i'm onboard with moving the logic to here and remove the validation package..

morgo: For now: yes, I prefer to remove the validation package.

Kiran01bm: i've made the necessary changes now..

Issue #446, opened 2025-09-12
Implement native Spirit linting


kolbe:
Spirit's current approach to linting has a couple important drawbacks:

1. linting requires connecting to a running database and executing DDL there to create a temporary table
2. unsafe diff detection isn't configurable (we can't easily add a new feature such as the invisible index check described below)

An alternative approach is hereby proposed: use the PingCAP TiDB parser to _parse_ the `CREATE TABLE` statement rather than creating a temporary table by _executing_ the `CREATE TABLE` statement. This means that we wouldn't have to connect to a MySQL server to be able to check an `ALTER TABLE` statement for adherence to standards (linting), and we could easily implement additional checks (unsafe diff) that relate the content of an `ALTER TABLE` statement to the current state of the table as described by the output of `SHOW CREATE TABLE`.

The prototypical use case for this is to require that an index be made invisible before it can be dropped. The current approach to linting offers no support for this, whereas the new approach would make it trivial to detect an alter with a `DROP INDEX` clause and check whether the index is currently marked invisible.

PR #486, opened 2025-10-14
add support for parsing CREATE TABLE

kolbe:
This is the first step toward implementing #446

This PR adds support for parsing CREATE TABLE statements using the TiDB parser. This support is primarily targeted toward parsing "canonical" CREATE TABLE statements, the ones returned by `SHOW CREATE TABLE`. The new support is not currently used, but will form the basis for future work that implements native linting.

It should be possible to parse _any_ legal MySQL CREATE TABLE statement using this functionality, but it's also possible to parse many _invalid_ statements (multiple primary keys, non-unique column names, indexes on columns that don't exist, etc.).

Many column/index/constraint attributes are expressed as pointers, following the general pattern in go structs when it should be possible to determine whether a field has *any* value set rather than just some default value.

This library provides convenient access to most aspects of a table, as well as access to the "raw" AST that comes back from the TiDB parser for more in-depth analysis depending on the situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,12 @@ This table does [include some secondary indexes](https://github.com/square/finch
 
 For a non-idle table, the performance delta is even greater. Consider the following microbench performed on a m1 mac with 10 cores and MySQL 8.0.31 using defaults:
 
-| Table/Scenario                               | Gh-ost   | spirit  | spirit (checksum disabled) |
-| -------------------------------------------- | -------- | ------- | -------------------------- |
-| finch.balances (800MB/1M rows), idle load    | 28.720s  | 11.197s | 9.278s                     |
-| finch.balances (800MB/1M rows), during bench | 2:50m+   | ~15-18s | ~15-18s                    |
+| Table/Scenario                               | Gh-ost   | spirit  |
+| -------------------------------------------- | -------- | ------- |
+| finch.balances (800MB/1M rows), idle load    | 28.720s  | 11.197s |
+| finch.balances (800MB/1M rows), during bench | 2:50m+   | ~15-18s |
 
-This scenario is kind of a worse case for gh-ost since it prioritizes replication over row-copying and the benchmark never lets up. The checksum feature is not present in gh-ost, and adds about 10-20% to migration time. We typically do not recommend disabling it.
+This scenario is kind of a worse case for gh-ost since it prioritizes replication over row-copying and the benchmark never lets up. The spirit time also includes a checksum.
 
 ## Unsupported Features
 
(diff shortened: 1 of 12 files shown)


pkg/statement/parse_create_table.go
```diff
@@ -0,0 +1,969 @@
...
+	for _, idx := range indexes {
+		if idx.Invisible != nil && *idx.Invisible {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (constraints Constraints) AnyForeignKeys() bool {
```
morgo: I prefer `HasForeignKeys`.

pkg/statement/parse_create_table.go
```diff
@@ -0,0 +1,969 @@
...
+	return partDef
+}
+
+// parsePartitionClause converts a partition clause to PartitionValues
+func (ct *CreateTable) parsePartitionClause(clause ast.PartitionDefinitionClause) *PartitionValues {
+	switch c := clause.(type) {
+	case *ast.PartitionDefinitionClauseLessThan:
+		values := &PartitionValues{
+			Type:   "LESS_THAN",
+			Values: make([]interface{}, 0, len(c.Exprs)),
```
morgo: Prefer `any` over interface{}` in new code

morgo (approved):
LGTM

pkg/statement/parse_create_table_test.go
```diff
@@ -0,0 +1,1251 @@
...
+
+	// Test first column (id)
+	idCol := columns[0]
+	assert.Equal(t, "id", idCol.Name)
+	assert.Contains(t, idCol.Type, "int") // TiDB returns "int(11)" not just "int"
+	assert.True(t, idCol.AutoInc)
+	assert.False(t, idCol.Nullable)
+
+	// Test second column (name)
+	nameCol := columns[1]
```
Kiran01bm, line 40: qq - is the order guaranteed ?

Kiran01bm: i.e is accessing columns by index safe/deterministic, or if we need to use name-based lookups instead ?

pkg/statement/parse_create_table_test.go
```diff
@@ -0,0 +1,1251 @@
...
+type ComprehensiveTestCase struct {
+	Name        string
+	SQL         string
+	ShouldParse bool
+	Validate    func(t *testing.T, createTable *CreateTable)
+}
+
+// TestComprehensiveParsingFromTiDBTestSuite tests our parsing library against patterns
+// derived from the TiDB parser's comprehensive test suite
+func TestComprehensiveParsingFromTiDBTestSuite(t *testing.T) {
```
Kiran01bm, line 712: should we add tests for generated columns as well ?

pkg/statement/parse_create_table_test.go
kolbe: Yes, columns need to be in a consistent order for a lot of reasons, so testing that is important

pkg/statement/parse_create_table_test.go
kolbe: Yes I would love that. After I merge this PR I'd love to see a *lot* of additional tests.

PR #490, opened 2025-10-16
Add support for keyBelowLowWatermark

morgo:

Fixes https://github.com/block/spirit/issues/465

This adds support to 'wait' before flushing changes if they are currently in progress by the copier. This is required by the buffered copier strategy because it reads changes locally, then writes them and there could be a lost update (although it will usually be discovered by the checksum).

However, I've implemented it for buffered and unbuffered strategies, since I think it will help reduce a potential contention/deadlock scenario between the copier and the replication flush.

It has a limitation: it prevents the checkpoint from advancing, since it doesn't safely know that up until that point is actually safely checkpointed. I believe this is acceptable, we'll have to look at one of the workarounds if that proves incorrect.

### DIFF
--- a/pkg/repl/subscription_map.go
+++ b/pkg/repl/subscription_map.go
@@ -108,7 +108,7 @@ func (s *deltaMap) Flush(ctx context.Context, underLock bool, lock *dbconn.Table
 	for key, isDelete := range s.changes {
 		unhashedKey := utils.UnhashKey(key)
 		if s.chunker != nil && !s.chunker.KeyBelowLowWatermark(unhashedKey[0]) {
-			s.c.logger.Debugf("key below watermark: %v", unhashedKey[0])
+			s.c.logger.Debugf("key not below watermark: %v", unhashedKey[0])
 			allChangesFlushed = false
 			continue
 		}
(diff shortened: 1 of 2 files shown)


pkg/repl/subscription_map.go
```diff
@@ -89,21 +89,31 @@ func (s *deltaMap) createReplaceStmt(replaceKeys []string) statement {
...
 	var replaceKeys []string
 	var stmts []statement
+	var keysFlushed []string // keys we need to remove from the map at the end.
 	var i int64
+	allChangesFlushed = true // assume all changes are flushed unless we find some that are not.
 	target := atomic.LoadInt64(&s.c.targetBatchSize)
 	for key, isDelete := range s.changes {
+		unhashedKey := utils.UnhashKey(key)
+		if s.chunker != nil && !s.chunker.KeyBelowLowWatermark(unhashedKey[0]) {
+			s.c.logger.Debugf("key below watermark: %v", unhashedKey[0])
```
kolbe: This appears to check that the key is *not* below the watermark (`!s.chunker.KeyBelowLowWatermark`), but the log line says the opposite?

pkg/table/datum.go
```diff
@@ -222,3 +222,13 @@ func (d Datum) GreaterThanOrEqual(d2 Datum) bool {
 	}
 	return d.Val.(uint64) >= d2.Val.(uint64)
 }
+
+func (d Datum) GreaterThan(d2 Datum) bool {
```
kolbe, line 229: Should this function assert that both Datums have the same type of underlying value?

kolbe (approved):

PR #492, opened 2025-10-17
add linter framework

kolbe:
This PR implements the linting framework for #446 

See the README.md in pkg/lint for more details

### DIFF
--- a/pkg/lint/README.md
+++ b/pkg/lint/README.md
@@ -85,18 +85,39 @@ func (l *MyCustomLinter) Lint(createTables []*statement.CreateTable, alterStatem
 
 ### Configuring Linters
 
+#### Enabling/Disabling Linters
+
 ```go
 // Disable specific linters
 violations, err := lint.RunLinters(tables, stmts, lint.Config{
     Enabled: map[string]bool{
-        "table_name_length": false,
-        "duplicate_column":  true,
+        "invisible_index_before_drop": false,
+        "primary_key_type":            true,
     },
 })
 if err != nil {
     // Handle configuration errors
 }
 ```
+
+#### Configurable Linters
+
+Some linters support additional configuration options via the `Settings` field. Linters that implement the `ConfigurableLinter` interface accept settings as `map[string]string`:
+
+```go
+violations, err := lint.RunLinters(tables, stmts, lint.Config{
+    Settings: map[string]map[string]string{
+        "invisible_index_before_drop": {
+            "raiseError": "true",  // Make violations errors instead of warnings
+        },
+    },
+})
+if err != nil {
+    // Handle configuration errors (e.g., invalid settings)
+}
+```
+
+Each configurable linter defines its own settings keys and values. See the individual linter documentation below for available options.
 
 ## Core Types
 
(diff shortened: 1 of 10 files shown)


pkg/lint/example/example.go
```diff
@@ -0,0 +1,132 @@
...
+
+// TableNameLengthConfig holds configuration for the table name length linter.
+type TableNameLengthConfig struct {
+	MaxLength int `json:"max_length"`
+}
+
+// NewTableNameLengthLinter creates a new table name length linter with default configuration.
+func NewTableNameLengthLinter() *TableNameLengthLinter {
+	return &TableNameLengthLinter{
+		maxLength: 64, // MySQL's limit
```
morgo: Nit: because we need to spirit migrate, we should limit at ~58 characters. There is an existing check in checks/ which checks this too.

pkg/lint/example/example.go
```diff
@@ -0,0 +1,132 @@
...
+	return l.Name()
+}
+
+func (l *TableNameLengthLinter) Name() string {
+	return "table_name_length"
+}
+
+func (l *TableNameLengthLinter) Category() string {
+	return "naming"
+}
```
morgo: Can we simplify some of these boiler plate funcs. It seems like the easiest is to remove `Name()`?

pkg/lint/README.md
```diff
@@ -0,0 +1,257 @@
...
+// Register your linter in init()
+func init() {
+    Register(&MyCustomLinter{})
+}
+
+// MyCustomLinter checks custom rules
+type MyCustomLinter struct{}
+
+func (l *MyCustomLinter) Name() string        { return "my_custom" }
+func (l *MyCustomLinter) Category() string    { return "naming" }
```
morgo, line 61: I'm not sure I see category really being used - or being difficult to manage since there is a lot of overlap.

We can go ahead with it if you have a plan, but if it doesn't work out, we should remove it.

pkg/lint/README.md
```diff
@@ -0,0 +1,257 @@
...
+pkg/lint/
+├── lint.go                      # Core API
+├── linter.go                    # Interface definition
+├── registry.go                  # Registration system
+├── violation.go                 # Violation types
+├── lint_invisible_index.go      # Built-in linter
+├── lint_multiple_alter.go       # Built-in linter
+├── lint_primary_key_type.go     # Built-in linter
+├── lint_my_new_rule.go          # Your new linter
+└── lint_my_new_rule_test.go     # Your tests
```
morgo, line 302: This LGTM

pkg/lint/README.md
kolbe: I agree, I don't have a good sense of the categories we'd want or how we'd use them, so I'll remove this for now.

pkg/lint/example/example.go
kolbe: We want each linter to have a name so that it can be referred to in output and configuration, and in golang making that a part of the interface makes sense to me. Would you rather the name just be given as a string when the linter registers itself? That drawback of that is that we then have to hardcode it into violations I think.

pkg/lint/example/example.go
kolbe: This is just an example linter, but there's no harm in changing this default value.

PR #493, opened 2025-10-18
fix data races in tests

kolbe:
I've seen persistent [data races in tests lately](https://github.com/block/spirit/actions/runs/18603224304/job/53046474106?pr=492#step:3:11472) and I asked goose to go figure it out. This is the PR it proposed.

@morgo you have better familiarity with this part of the code than I do. Does it look like the proposed change to execute `r.replClient.SetDDLNotificationChannel(nil)` before `close(r.ddlNotification)` is safe or do you think this might be problematic?

### DIFF
--- a/pkg/repl/client.go
+++ b/pkg/repl/client.go
@@ -101,7 +101,7 @@ func NewClient(db *sql.DB, host string, username, password string, config *Clien
 	if config.WriteDB == nil {
 		config.WriteDB = db // default to using the read DB for writes
 	}
-	client := &Client{
+	return &Client{
 		db:                         db,
 		dbConfig:                   dbconn.NewDBConfig(),
 		host:                       host,
@@ -109,6 +109,7 @@ func NewClient(db *sql.DB, host string, username, password string, config *Clien
 		password:                   password,
 		logger:                     config.Logger,
 		targetBatchTime:            config.TargetBatchTime,
+		targetBatchSize:            DefaultBatchSize, // initial starting value.
 		concurrency:                config.Concurrency,
 		subscriptions:              make(map[string]Subscription),
 		onDDL:                      config.OnDDL,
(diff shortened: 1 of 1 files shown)


pkg/repl/client.go
```diff
@@ -101,22 +101,25 @@ func NewClient(db *sql.DB, host string, username, password string, config *Clien
...
 		subscriptions:              make(map[string]Subscription),
 		onDDL:                      config.OnDDL,
 		serverID:                   config.ServerID,
 		useExperimentalBufferedMap: config.UseExperimentalBufferedMap,
 		writeDB:                    config.WriteDB,
 	}
+	// Initialize targetBatchSize atomically to avoid data race
+	// since it's accessed atomically elsewhere in the code
+	atomic.StoreInt64(&client.targetBatchSize, DefaultBatchSize)
+	return client
```
morgo: I don't think this will fix this race.

The `DefaultBatchSize` is a pseudo const. We make it a var though so it can be changed in the tests. I think to fix the race we'll need to change the `DefaultBatchSize` in the TestMain before tests start executing.

pkg/migration/runner.go
```diff
@@ -711,6 +711,14 @@ func (r *Runner) Close() error {
...
 		}
 	}
+	// Set the DDL notification channel to nil before closing it
+	// to prevent race conditions where another goroutine might try to send to it
+	if r.replClient != nil {
+		r.replClient.SetDDLNotificationChannel(nil)
+	}
+	if r.ddlNotification != nil {
+		close(r.ddlNotification)
+	}
```
morgo, line 721: This looks fine to me. I've seen this race too and it probably fixes it.

pkg/repl/client.go
kolbe: I don't think this is an actual race at all, this is just something goose found while rummaging around. If it doesn't cause any problems, maybe we should use its suggestion so it doesn't get sidetracked again in future?

morgo: I'd prefer we revert this part if we don't see a benefit. Otherwise I'll be in a position in the future where I am trying to refactor and I won't understand the atomic requirement.

kolbe: Sure thing. I hope we don't have this come up every time we ask an LLM to help us find data races.

PR #496, opened 2025-10-21
Improve MoveTables command

morgo:

Fixes https://github.com/block/spirit/issues/495

It refactors `Status` so it can be shared between Move and Migration. There is still a lot of code duplication though, with future opportunities to cleanup such as the dump checkpoint / resume from checkpoint code paths.

### DIFF
--- a/pkg/dbconn/dbconn.go
+++ b/pkg/dbconn/dbconn.go
@@ -10,7 +10,9 @@ import (
 	"time"
 
 	"github.com/block/spirit/pkg/dbconn/sqlescape"
+	"github.com/block/spirit/pkg/table"
 	"github.com/go-sql-driver/mysql"
+	"github.com/siddontang/loggers"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


pkg/move/runner.go
```diff
@@ -5,68 +5,101 @@ import (
...
+	gomysql "github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-sql-driver/mysql"
+	"github.com/siddontang/loggers"
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	sentinelCheckInterval   = 1 * time.Second
+	tableStatUpdateInterval = 5 * time.Minute
+	sentinelWaitLimit       = 48 * time.Hour
```
aparajon, line 27: Should this be configurable?

pkg/move/runner.go
```diff
@@ -276,5 +486,170 @@ func (r *Runner) prepareForCutover(ctx context.Context) error {
...
+// It writes the current state of the migration to the checkpoint table,
+// which can be used in recovery. Previously resuming from checkpoint
+// would always restart at the copier, but it can now also resume at
+// the checksum phase.
+func (r *Runner) dumpCheckpoint(ctx context.Context) error {
+	// Retrieve the binlog position first and under a mutex.
+	binlog := r.replClient.GetBinlogApplyPosition()
+	copierWatermark, err := r.copyChunker.GetLowWatermark()
+	if err != nil {
+		return status.ErrWatermarkNotReady // it might not be ready, we can try again.
```
aparajon, line 592: Do we care about logging these sort of errors?

pkg/move/runner.go
```diff
@@ -155,9 +234,44 @@ func (r *Runner) Run(ctx context.Context) error {
...
+	} else {
+		return r.newCopy(ctx)
+	}
+	return nil
+}
+
+func (r *Runner) newCopy(ctx context.Context) error {
+	// We are starting fresh:
+	// For each table, fetch the CREATE TABLE statement from the source and run it on the target.
+	if err := r.createTargetTables(); err != nil {
```
aparajon, line 257: Should we add a timer log to see how long this took?

pkg/move/cutover.go
```diff
@@ -0,0 +1,108 @@
...
+			continue
+		}
+		c.logger.Warn("final cut over operation complete")
+		return nil
+	}
+	c.logger.Error("cutover failed, and retries exhausted")
+	return err
+}
+
+func (c *CutOver) algorithmRenameUnderLock(ctx context.Context) error {
```
aparajon, line 74: Is it worth adding any info logs here for each of the phases? E.g.

```
  c.logger.Info("Starting cutover: acquiring table locks")
  tableLock, err := dbconn.NewTableLock(ctx, c.db, c.tables, c.dbConfig, c.logger)
 
  ...

  c.logger.Info("Cutover: flushing changes under lock")
  if err := c.feed.FlushUnderTableLock(ctx, tableLock); err != nil {
      return err
  }

  ...
  c.logger.Info("Cutover: performing table rename operations")
  return tableLock.ExecUnderLock(ctx, renameStatement)
```

pkg/move/runner.go
morgo: If we understand the use-cases maybe. This timeout is pre-existing and just copied to the move package from migration.

For now, we limit it to 48 hours per "decisions, not options" in https://github.com/block/spirit/blob/main/.github/CONTRIBUTING.md

pkg/move/cutover.go
morgo: The flush phase does record its time and print it out to the logger.

The table lock will do similar - it prints before it executes statements and we can deduce time from there.

Exec under lock I don't think we will have time on, but since it has pre-acquired the lock it should be instant.

pkg/move/runner.go
morgo: We can roughly tell from the `dumpStatus()` as things move to different stages. Unfortunately, I've not yet implemented dumpStatus() for the move runner.

pkg/move/runner.go
morgo: It will log it. It returns the error to dumpCheckpointContinuously, which on 999 will log it.

Issue #502, opened 2025-10-23
force-kill does not free transactions


morgo:
The force-kill consumes a transaction, and doesn't free it back to the pool.

It also dumps queries to the log.

This is a problem in multi-use cases because it runs a check for-each ALTER, which can DoS all the connections that are available.

PR #503, opened 2025-10-23
fix kill issues

morgo:

Fixes https://github.com/block/spirit/issues/502

### DIFF
--- a/pkg/check/privileges_test.go
+++ b/pkg/check/privileges_test.go
@@ -55,14 +55,12 @@ func TestPrivileges(t *testing.T) {
 	r.ForceKill = true
 	err = privilegesCheck(t.Context(), r, logrus.New())
 	assert.Error(t, err)
-	t.Log(err)
 
 	_, err = db.Exec("GRANT SELECT on `performance_schema`.* TO testprivsuser")
 	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


pkg/dbconn/kill.go
```diff
@@ -225,8 +220,9 @@ func GetLockingTransactions(ctx context.Context, db *sql.DB, tables []*table.Tab
 func GetTableLocks(ctx context.Context, db *sql.DB, tables []*table.TableInfo, logger loggers.Advanced, ignorePIDs []int) ([]*LockDetail, error) {
 	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
-		logger.Errorf("failed to begin transaction: %v", err)
+		return nil, err
```
morgo, line 223: This is likely a bug. If it can't get a transaction it should return an error. I haven't reproduced it yet.

pkg/dbconn/kill.go
```diff
@@ -225,8 +220,9 @@ func GetLockingTransactions(ctx context.Context, db *sql.DB, tables []*table.Tab
...
 	}
+	defer tx.Rollback()
```
morgo: This is the main bug.

Issue #504, opened 2025-10-23
support force-kill in INSTANT / INPLACE DDL code paths


morgo:
Currently INSTANT + INPLACE code paths could be starved for a metadata lock.

We should use the same logic we use for force-kill when acquiring a lock for cutover, and kill any connections preventing the main operation after 90% of timeout time has passed.

PR #506, opened 2025-10-25
Add force-kill support into INSTANT/INPLACE attempts

morgo:

Fixes https://github.com/block/spirit/issues/504

### DIFF
--- a/pkg/migration/migration_test.go
+++ b/pkg/migration/migration_test.go
@@ -561,6 +561,7 @@ func TestSecondaryEngineAttribute(t *testing.T) {
 		Database:  cfg.DBName,
 		Threads:   1,
 		Statement: `ALTER TABLE t1secondary ADD KEY (title) SECONDARY_ENGINE_ATTRIBUTE='{"type":"spann", "distance":"l2", "product_quantization":{"dimensions":96}}'`,
+		ForceKill: true, // this happens to be instant, tests this code path too.
 	}
 	err = migration.Run()
 	assert.NoError(t, err)
--- a/pkg/migration/runner_test.go
+++ b/pkg/migration/runner_test.go
@@ -3049,13 +3049,14 @@ func TestIndexVisibility(t *testing.T) {
 
 	// Test again with visible
 	m, err = NewRunner(&Migration{
-		Host:     cfg.Addr,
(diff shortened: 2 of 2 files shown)


pkg/dbconn/dbconn.go
```diff
@@ -184,6 +186,40 @@ func backoff(i int) {
...
+	timer := time.AfterFunc(threshold, func() {
+		logger.Warnf("waited for %v; trying to kill locking transactions", threshold)
+		err := KillLockingTransactions(ctx, db, tables, dbConfig, logger, []int{connId})
+		if err != nil {
+			logger.Errorf("failed to kill locking transactions: %v", err)
+		}
+	})
+	defer timer.Stop()
+	_, err = trx.ExecContext(ctx, sqlescape.MustEscapeSQL(stmt, args...))
+	return err
```
aparajon, line 220: Do we need to add context about the statement that failed and/or the success state?

E.g.

```
if err != nil {
    return fmt.Errorf("failed to execute statement with force-kill: %w", err)
}
logger.Infof("force-kill DDL operation completed successfully")
return nil
```

pkg/dbconn/dbconn.go
```diff
@@ -184,6 +186,40 @@ func backoff(i int) {
...
+	}
+	defer func() {
+		// We need to ensure we always clean up the transaction.
+		// In the typically case we are using this for non-transactional
+		// statements (and could rollback either way), but just to be safe
+		// we check the error and commit on-nil.
+		if err != nil {
+			_ = trx.Rollback()
+		} else {
+			_ = trx.Commit()
```
kolbe, line 205: I'd rather commit only explicitly after ExecContext if it doesn't return an error and then simply defer a call to Rollback. I think that's a more common pattern?

pkg/dbconn/dbconn.go
morgo: I think it's fine - the logger is passed into the force-kill handler, and it will log some context about what actions it takes.

An error in this context that is non-kill related usually means that it was not suitable as instant/inplace. Both of these functions are called optimistically, and expected to fail. This is handled by the caller that will mark some flags for usedInplace/usedInstant which are logged in final output.

PR #507, opened 2025-10-27
move checksum retries into checksum package

morgo:

Fixes https://github.com/block/spirit/issues/513

This previously made some test changes, but I've moved them to https://github.com/block/spirit/pull/514

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -391,7 +391,7 @@ func (c *Checker) Run(ctx context.Context) error {
 	// Try the checksum up to n times if differences are found and we can fix them
 	for attempt := 1; attempt <= c.maxRetries; attempt++ {
 		if attempt > 1 {
-			c.logger.Errorf("checksum failed, retrying %d/%d times", attempt-1, c.maxRetries)
+			c.logger.Errorf("checksum failed, retrying %d/%d times", attempt, c.maxRetries)
 			// Reset the chunker to start from the beginning
 			if err := c.chunker.Reset(); err != nil {
 				return fmt.Errorf("failed to reset chunker for retry: %w", err)


pkg/checksum/checker.go
```diff
@@ -385,6 +388,44 @@ func (c *Checker) Run(ctx context.Context) error {
 		c.ExecTime = time.Since(startTime)
 	}()
 
+	// Try the checksum up to n times if differences are found and we can fix them
+	for attempt := 1; attempt <= c.maxRetries; attempt++ {
+		if attempt > 1 {
+			c.logger.Errorf("checksum failed, retrying %d/%d times", attempt-1, c.maxRetries)
```
kolbe: This will print "1/3 times" on the first retry, "2/3 times" on the second retry, and then bail, right? That's kind of confusing.

pkg/checksum/checker.go
```diff
@@ -385,6 +388,44 @@ func (c *Checker) Run(ctx context.Context) error {
...
+		}
+	}
+
+	// Retries exhausted:
+	// This used to say "checksum failed, this should never happen" but that's not entirely true.
+	// If the user attempts a lossy schema change such as adding a UNIQUE INDEX to non-unique data,
+	// then the checksum will fail. This is entirely expected, and not considered a bug. We should
+	// do our best-case to differentiate that we believe this ALTER statement is lossy, and
+	// customize the returned error based on it.
+	return fmt.Errorf("checksum failed after %d attempts. This likely indicates either a bug in Spirit, or a manual modification to the _new table outside of Spirit. Please report @ github.com/block/spirit", c.maxRetries)
```
kolbe, line 425: The error you return doesn't capture the lossy/UNIQUE case you explain in the comments. Should it?

pkg/migration/runner.go
```diff
@@ -872,52 +867,15 @@ func (r *Runner) checksum(ctx context.Context) error {
...
-			if r.addsUniqueIndex() {
-				return errors.New("checksum failed after 3 attempts. Check that the ALTER statement is not adding a UNIQUE INDEX to non-unique data")
-			}
-			return errors.New("checksum failed after 3 attempts. This likely indicates either a bug in Spirit, or a manual modification to the _new table outside of Spirit. Please report @ github.com/block/spirit")
+
+	// Run the checksum with internal retry logic
+	if err := r.checker.Run(ctx); err != nil {
+		if r.addsUniqueIndex() {
+			// Overwrite the error if we think it's because of a unique index addition
+			return errors.New("checksum failed after several attempts. This is likely related to your statement adding a UNIQUE index on non-unique data")
```
kolbe, line 875: Hmmm I see that this might be the solution to the prior review comment.

pkg/checksum/checker.go
morgo: It overwrites the error in the caller if it detects a unique index, so it should be fine.

pkg/checksum/checker.go
morgo: Yeah good point, I'll fix it.

PR #510, opened 2025-10-28
tls support for replication throttler and binlog replica stream - implement issue #175 (part 2)

### LINKED ISSUE #175: Add full TLS support
Currently we auto-select the `rds-ca-2019-root.pem` certificate if the hostname matches an RDS hostname. There's 2 problems with this:

1. Users may select a newer RDS certificate.
2. Users (outside of RDS) may want to use TLS and have their own certificate.

This is a FR to add full TLS support. We are open to external contributors.

Kiran01bm:
## What and Why ?
The following changes are part of this PR:
1. TLS support for the following (Fixes #175 - part 2_):
- Replication throttler (connections to replica db).
- Binlog sync'er (replica stream from main db).
- TLS Inheritance for replica db and override support via explicit parameter preservation.
2. FallBack DSN generation is now re-useable  (applies to the `PREFERRED` mode and enables graceful degradation from TLS to plain text).
3. Explicit tests for custom CA and CA overrides.
4. Majority of the tests are written such that the migrations end-up using table copy method as a forcing function to test replica db connections.
5. Tests for mixed envs - main db is rds but replica is not and vice-versa.

### TLS Inheritance
- **Replica throttler** automatically inherits TLS configuration from main database connection.
- **Inheritance only applies** when main TLS mode is not `DISABLED` and replica DSN has no explicit TLS settings.
- **All TLS modes supported**: `PREFERRED`, `REQUIRED`, `VERIFY_CA`, `VERIFY_IDENTITY` inherited correctly.

### Explicit TLS Parameter Preservation  
- **Explicit override precedence over inheritance** - `tls=false`, `tls=required`, etc. in replica DSN are now respected and not overridden. 
- **Complete preservation** - both query string and structured TLS configs preserved and case-insensitive handling i.e handles `tls=`, `TLS=`, `Tls=`.


pkg/dbconn/conn.go
```diff
@@ -292,6 +305,11 @@ func newDSN(dsn string, config *DBConfig) (string, error) {
 // append additional options to it to standardize the connection.
 // It will also ping the connection to ensure it is valid.
 func New(inputDSN string, config *DBConfig) (db *sql.DB, err error) {
+	return NewWithContext(inputDSN, config, "main database")
+}
+
+// NewWithContext is like New but includes context about the connection type for better error messages
+func NewWithContext(inputDSN string, config *DBConfig, connectionType string) (db *sql.DB, err error) {
```
morgo: To me "Context" means something specific in golang. Can we use a name like `NewWithConnectionType` instead?

pkg/migration/runner.go
```diff
@@ -146,7 +146,7 @@ func (r *Runner) Run(ctx context.Context) error {
 	}
 	r.db, err = dbconn.New(r.dsn(), r.dbConfig)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to connect to main database (DSN: %s): %w", r.dsn(), err)
```
morgo: I think this will leak the password? dsn() is a string so it will be included.

pkg/repl/client.go
```diff
@@ -113,6 +116,7 @@ func NewClient(db *sql.DB, host string, username, password string, config *Clien
 	return &Client{
 		db:                         db,
 		dbConfig:                   dbconn.NewDBConfig(),
+		tlsConfig:                  config.TLSConfig,
```
morgo: You can remove `tlsConfig` from the struct:

- Add dbConfig to the clientConfig as a field and set line 118 to config.DBConfig.
- When we create a new client config, we pass in DBConfig not TLSConfig.

morgo (commented):
A couple of small things to cleanup

pkg/migration/runner.go
Kiran01bm: thats a good catch.. updated it now..

pkg/repl/client.go
Kiran01bm: sure, sounds good... i've made the changes - pls take a look.

pkg/migration/runner.go
```diff
@@ -404,6 +405,30 @@ func (r *Runner) dsn() string {
...
+	// Find the : that separates username from password
+	colonIndex := strings.Index(dsn, ":")
+	if colonIndex == -1 || colonIndex > atIndex {
+		return dsn // No password in DSN
+	}
+
+	// Replace password with ***
+	return dsn[:colonIndex+1] + "***" + dsn[atIndex:]
+}
+
```
morgo, line 437: This is probably correct, but to limit having to test it, why not use ParseDSN:
https://github.com/go-sql-driver/mysql/blob/76c00e35a8d48f8f70f0e7dffe584692bd3fa612/dsn.go#L396-L399

I haven't checked if there is a stringer on Config which blanks out the password, but after it's parsed you could also just blank it out.

Kiran01bm: good point.. will take a look..

Kiran01bm: the changes are now in-place .. will merge it per your message and it there are any other concerns i'll deal with as follow-up..

Issue #498, opened 2025-10-21
Make MoveTables check the table structure is consistent


morgo:
Currently on the resume path it doesn't check reliably that there have been no changes.

Issue #466, opened 2025-09-29
data race: TestChangeDatatypeLossyFailEarly


morgo:
```
025-09-29T16:25:24.5808070Z test-1           | ==================
2025-09-29T16:25:24.5808148Z test-1           | WARNING: DATA RACE
2025-09-29T16:25:24.5808269Z test-1           | Read at 0x00c0005b84f8 by goroutine 10727:
2025-09-29T16:25:24.5808419Z test-1           |   github.com/go-sql-driver/mysql.(*mysqlConn).cleanup()
2025-09-29T16:25:24.5808629Z test-1           |       /go/pkg/mod/github.com/go-sql-driver/mysql@v1.7.1/connection.go:146 +0x7b
2025-09-29T16:25:24.5808765Z test-1           |   github.com/go-sql-driver/mysql.(*mysqlConn).cancel()
2025-09-29T16:25:24.5808961Z test-1           |       /go/pkg/mod/github.com/go-sql-driver/mysql@v1.7.1/connection.go:436 +0x72
2025-09-29T16:25:24.5809136Z test-1           |   github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1()
2025-09-29T16:25:24.5809334Z test-1           |       /go/pkg/mod/github.com/go-sql-driver/mysql@v1.7.1/connection.go:622 +0x21e
2025-09-29T16:25:24.5809401Z test-1           | 
2025-09-29T16:25:24.5809646Z test-1           | Previous write at 0x00c0005b84f8 by goroutine 10698:
2025-09-29T16:25:24.5810009Z test-1           |   github.com/go-sql-driver/mysql.(*mysqlConn).writeHandshakeResponsePacket()
2025-09-29T16:25:24.5810389Z test-1           |       /go/pkg/mod/github.com/go-sql-driver/mysql@v1.7.1/packets.go:371 +0xc90
2025-09-29T16:25:24.5810574Z test-1           |   github.com/go-sql-driver/mysql.(*connector).Connect()
2025-09-29T16:25:24.5810767Z test-1           |       /go/pkg/mod/github.com/go-sql-driver/mysql@v1.7.1/connector.go:103 +0xcef
2025-09-29T16:25:24.5810860Z test-1           |   database/sql.(*DB).conn()
2025-09-29T16:25:24.5810992Z test-1           |       /usr/local/go/src/database/sql/sql.go:1431 +0xed0
[...]

PR #515, opened 2025-10-28
Add table structure test to resume

morgo:

Fixes https://github.com/block/spirit/issues/498

Fixes https://github.com/block/spirit/issues/466

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.24.0
 require (
 	github.com/alecthomas/kong v0.7.1
 	github.com/go-mysql-org/go-mysql v1.9.1
-	github.com/go-sql-driver/mysql v1.7.1
+	github.com/go-sql-driver/mysql v1.9.3
 	github.com/pingcap/errors v0.11.5-0.20250523034308-74f78ae071ee
 	github.com/pingcap/tidb/pkg/parser v0.0.0-20250811102254-4230cf349b01
 	github.com/siddontang/go-log v0.0.0-20190221022429-1e957dd83bed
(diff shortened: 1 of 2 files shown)


pkg/move/runner.go
```diff
@@ -641,6 +651,10 @@ func (r *Runner) dumpCheckpointContinuously(ctx context.Context) {
 					r.logger.Warnf("could not write checkpoint yet, watermark not ready")
 					continue
 				}
+				// If the error is context canceled, that's fine too.
+				if errors.Is(err, context.Canceled) {
+					return
```
aparajon, line 656: Do we need to log the cancellation here?

pkg/move/move_test.go
```diff
@@ -49,3 +61,104 @@ func TestBasicMove(t *testing.T) {
...
+	assert.NoError(t, r.source.Close())
+	assert.NoError(t, r.target.Close())
+	r.Close()
+
+	// Alter the definition of target.t1 just to be difficult.
+	// This will prevent resume.
+	testutils.RunSQL(t, `ALTER TABLE dest_resume.t1 ADD COLUMN extra_col INT DEFAULT 0`)
+	r, err = NewRunner(move)
+	assert.NoError(t, err)
+	assert.Error(t, r.Run(t.Context()))
```
aparajon, line 153: Should we assert on a specific error message here, which I'm guessing happens when resume gets interrupted?

pkg/move/runner.go
morgo: No. This means it was canceled elsewhere, this is just handling the cleanup inside this go-routine.

pkg/move/move_test.go
morgo: We could. Because I'm not returning a specific fixed error though on failure, I would need to string match (not saying I don't do this other places, but because the retry passes with the one change of dropping the col, I don't need to.)

Issue #522, opened 2025-10-30
Flaky test: TestAddUniqueIndexChecksumEnabled


morgo:
This is a recently added test:

```
2025-10-29T22:51:44.9771432Z test-1   | [31m--- FAIL: TestAddUniqueIndexChecksumEnabled (17.86s)[0m
2025-10-29T22:51:44.9771761Z test-1   | 
2025-10-29T22:51:44.9772149Z test-1   |     runner_test.go:697: 
2025-10-29T22:51:44.9772708Z test-1   |         	Error Trace:	/app/pkg/migration/runner_test.go:697
2025-10-29T22:51:44.9773120Z test-1   |         	Error:      	Received unexpected error:
2025-10-29T22:51:44.9773426Z test-1   |         	            	context canceled
2025-10-29T22:51:44.9773902Z test-1   |         	Test:       	TestAddUniqueIndexChecksumEnabled
```

Issue #499, opened 2025-10-21
Add support for checksum fixDifferences for moveTables


morgo:
Currently it can only detect issues, it won't fix them.

PR #523, opened 2025-10-30
MoveTables: Add status, checksum repair

morgo:

Fixes https://github.com/block/spirit/issues/499
Fixes https://github.com/block/spirit/issues/522

Also makes the following changes:
- Implements `status` for moveTables and refactors status and checkpointing to be part of status package, so there is some consistency in the frequencies these are called between tools.
- Some small cleanup post https://github.com/block/spirit/pull/507 because we no longer need to check if the checker is nil.
- Removes the testlogger (no longer required with refactoring such that status returns a string)

### DIFF
--- a/pkg/checksum/checker.go
+++ b/pkg/checksum/checker.go
@@ -23,6 +23,11 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var (
+	// Query template for row checksums
+	queryTemplate = "SELECT CRC32(CONCAT(%s)) as row_checksum, CONCAT_WS(',', %s) as pk FROM %s WHERE %s"
+)
+
 type Checker struct {
 	sync.Mutex
 	concurrency      int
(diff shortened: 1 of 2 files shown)


pkg/checksum/checker.go
```diff
@@ -155,169 +152,288 @@ func (c *Checker) ChecksumChunk(ctx context.Context, trxPool *dbconn.TrxPool, ch
...
+			return fmt.Errorf("failed to scan source row: %w", err)
+		}
+		sourceChecksums[pk] = checksum
+	}
+	if err := sourceRows.Err(); err != nil {
+		return fmt.Errorf("error iterating source rows: %w", err)
+	}
+
+	// Query target database for row checksums
+	targetQuery := fmt.Sprintf("SELECT CRC32(CONCAT(%s)) as row_checksum, CONCAT_WS(',', %s) as pk FROM %s WHERE %s",
```
kolbe: sourceQuery and targetQuery have the same template text so maybe that could be reused

pkg/move/runner.go
```diff
@@ -460,8 +461,53 @@ func (r *Runner) startBackgroundRoutines(ctx context.Context) {
...
+		// This could take a while if it's a large table.
+		return fmt.Sprintf("migration status: state=%s checksum-progress=%s binlog-deltas=%v total-time=%s checksum-time=%s",
+			r.status.Get().String(),
+			r.checker.GetProgress(),
+			r.replClient.GetDeltaLen(),
+			time.Since(r.startTime).Round(time.Second),
+			time.Since(r.checker.StartTime()).Round(time.Second),
+		)
+	}
+	return ""
```
kolbe: If you put the return statement inside a `default:` section of the case I think you wouldn't need to set //nolint:exhaustive.

kolbe (approved):

Issue #524, opened 2025-10-31
Make Metadatalock accept tables (plural) instead of table


morgo:
Currently we need one connection per metadata lock. It should be fairly straight forward to make things use one connection:

```
cat ~/mdl-single-conn.patch
diff --git a/pkg/dbconn/metadatalock.go b/pkg/dbconn/metadatalock.go
index 1520d41..d693725 100644
--- a/pkg/dbconn/metadatalock.go
+++ b/pkg/dbconn/metadatalock.go
@@ -27,16 +27,16 @@ type MetadataLock struct {
 	closeCh         chan error
 	refreshInterval time.Duration
 	db              *sql.DB
-	lockName        string
+	lockNames       []string // Multiple lock names for multiple tables
 }

-func NewMetadataLock(ctx context.Context, dsn string, table *table.TableInfo, config *DBConfig, logger loggers.Advanced, optionFns ...func(*MetadataLock)) (*MetadataLock, error) {
-	if table == nil {
-		return nil, errors.New("metadata lock table info is nil")
+func NewMetadataLock(ctx context.Context, dsn string, tables []*table.TableInfo, config *DBConfig, logger loggers.Advanced, optionFns ...func(*MetadataLock)) (*MetadataLock, error) {
+	if len(tables) == 0 {
+		return nil, errors.New("no tables provided for metadata lock")
 	}
-
 	mdl := &MetadataLock{
 		refreshInterval: refreshInterval,
+		lockNames:       make([]string, 0, len(tables)),
 	}

 	// Apply option functions
@@ -44,6 +44,11 @@ func NewMetadataLock(ctx context.Context, dsn string, table *table.TableInfo, co
 		optionFn(mdl)
 	}

+	// Compute lock names for all tables
+	for _, tbl := range tables {
+		mdl.lockNames = append(mdl.lockNames, computeLockName(tbl))
+	}
+
 	// Setup the dedicated connection for this lock
 	// Use the provided config but ensure MaxOpenConnections is 1 for metadata locks
 	dbConfig := *config // Copy the config
@@ -54,43 +59,37 @@ func NewMetadataLock(ctx context.Context, dsn string, table *table.TableInfo, co
 		return nil, err
 	}

-	// Function to acquire the lock
-	getLock := func() error {
+	// Function to acquire all locks
+	getLocks := func() error {
+		// Acquire locks for all tables
 		// https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html#function_get-lock
-		var answer int
[...]

PR #525, opened 2025-11-01
Have metadatalocks support multiple locks in one connection

morgo:

Fixes https://github.com/block/spirit/issues/524


pkg/dbconn/metadatalock.go
```diff
@@ -168,7 +169,15 @@ func (m *MetadataLock) CloseDBConnection(logger loggers.Advanced) error {
...
 func (m *MetadataLock) GetLockName() string {
-	return m.lockName
+	// For backwards compatibility, return the first lock name
+	if len(m.lockNames) > 0 {
+		return m.lockNames[0]
+	}
+	return ""
+}
+
+func (m *MetadataLock) GetLockNames() []string {
```
Kiran01bm: appears unused ?

pkg/dbconn/metadatalock.go
```diff
@@ -54,43 +59,37 @@ func NewMetadataLock(ctx context.Context, dsn string, table *table.TableInfo, co
...
+			if err := mdl.db.QueryRowContext(ctx, stmt).Scan(&answer); err != nil {
+				return fmt.Errorf("could not acquire metadata lock for %s: %s", lockName, err)
+			}
+			if answer == 0 {
+				// 0 means the lock is held by another connection
+				logger.Warnf("could not acquire metadata lock for %s, lock is held by another connection", lockName)
+				return fmt.Errorf("could not acquire metadata lock for %s, lock is held by another connection", lockName)
+			} else if answer != 1 {
+				// probably we never get here, but just in case
+				return fmt.Errorf("could not acquire metadata lock %s, GET_LOCK returned: %d", lockName, answer)
```
Kiran01bm, line 78: if we do hit this block - should we defer and release the previously acquired locks ex: 4th of the 5th fails and we release locks for 1,2 and 3 ?

morgo: It should be handled, the code does this:
```
	if err = getLocks(); err != nil {
		mdl.db.Close() // close if we are not returning an MDL.
		return nil, err
	}
```

If the connection is closed, the locks are released. So any partial lock acquisition prior to a failure will be released.

Issue #526, opened 2025-11-03
error reading binlog stream: Last sync error or closed


tyd:
We had a table migration going and for some reason there was an issue reading a binlog... Spirit retried a couple billion times (see screenshot below) and was never successful. We finally noticed (will be adding alerting!) and killed it. This looks like a spot that needs to work

**Log extract:**
> time="2025-11-02T14:46:39Z" level=error msg="error reading binlog stream: Last sync error or closed, try sync and get event again, current position: (mysql-bin-changelog.008138, 0)"

<img width="1437" height="646" alt="Image" src="https://github.com/user-attachments/assets/c9b353bc-17a3-454d-afa8-6d6a9406ffa8" />

morgo:
I believe this might be fixed in https://github.com/block/spirit/pull/514.

The replication client now supports reconnect. In theory it looked like the client auto-reconnects, but in practice I noticed in the tests it wasn't always, so I added it.

I haven't seen it in the wild though, so it's good (but unfortunate) to see you hit it.

tyd:
> We've been standardizing on (1) since this is our preferred behavior

Agree that option 1 is much preferred over unbounded retries (2 billion!). I'll put a PR together with a reasonable limit and we can adjust as your team sees fit. Will need to put some thought into how to test it as well.

PR #527, opened 2025-11-03
feat: Add maxRecreateAttempts logic

tyd:
Per discussion in #526, here's a PR to introduce `maxRecreateAttempts` and to panic when the limit is reached. I set default to 10 which seemed reasonable.

**Notes:**
- `maxRecreateAttempts` is a var so we can override it in test to a smaller value. Outside of the test context it's value comes from the `const maxRecreateAttemptsDefault`
- Testing panics isn't easy, so I had to create a couple tests to accomplish it via a sub-process. Please check those closely to make sure that approach is OK.
- Added some code to fix a potential bug introduced in #514. After a failed streamer recreation, set `c.streamer = nil` so the next loop iteration detects the failure immediately. This forces recreation on each iteration until we hit the max attempts limit, rather than waiting for 5 more `GetEvent()` failures or hitting a nil pointer crash.

**New Tests:**
- `TestMaxRecreateAttempts`
- `TestMaxRecreateAttemptsPanic`
- `testMaxRecreateAttemptsPanicSubprocess`

### DIFF
--- a/pkg/dbconn/metadatalock.go
+++ b/pkg/dbconn/metadatalock.go
@@ -27,21 +27,26 @@ type MetadataLock struct {
 	closeCh         chan error
 	refreshInterval time.Duration
 	db              *sql.DB
-	lockName        string
+	lockNames       []string // Multiple lock names for multiple tables
 }
 
-func NewMetadataLock(ctx context.Context, dsn string, table *table.TableInfo, config *DBConfig, logger loggers.Advanced, optionFns ...func(*MetadataLock)) (*MetadataLock, error) {
-	if table == nil {
-		return nil, errors.New("metadata lock table info is nil")
+func NewMetadataLock(ctx context.Context, dsn string, tables []*table.TableInfo, config *DBConfig, logger loggers.Advanced, optionFns ...func(*MetadataLock)) (*MetadataLock, error) {
+	if len(tables) == 0 {
+		return nil, errors.New("no tables provided for metadata lock")
 	}
-
 	mdl := &MetadataLock{
 		refreshInterval: refreshInterval,
+		lockNames:       make([]string, 0, len(tables)),
 	}
 
 	// Apply option functions
 	for _, optionFn := range optionFns {
 		optionFn(mdl)
+	}
+
+	// Compute lock names for all tables
+	for _, tbl := range tables {
+		mdl.lockNames = append(mdl.lockNames, computeLockName(tbl))
 	}
 
 	// Setup the dedicated connection for this lock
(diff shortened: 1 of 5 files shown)


morgo:
Sorry, you are getting a CI failure due to https://github.com/block/spirit/issues/522 - it is unrelated.

If this is ready for review, I can take a look tomorrow.

pkg/repl/client_test.go
```diff
@@ -758,3 +760,110 @@ func TestAllChangesFlushed(t *testing.T) {
...
+	// Test that we can modify it for testing
+	originalValue := maxRecreateAttempts
+	defer func() { maxRecreateAttempts = originalValue }()
+
+	maxRecreateAttempts = 5
+	assert.Equal(t, 5, maxRecreateAttempts)
+
+	maxRecreateAttempts = originalValue
+	assert.Equal(t, maxRecreateAttemptsDefault, maxRecreateAttempts)
+}
```
morgo: I think this will trigger the race detector because its read in other tests concurrently?

What we've done in other tests like migration is:
- `maxRecreateAttemptsDefault` is a var, but we comment to say it's practically a const.
- In the TestMain() (i.e. pkg/migration/migration_test.go:22-29) we overwrite these vars for test. I'm OK with overwriting it to 3 for the tests: that's sufficient to test it retries, with a retry of a retry.
- This will also eliminate your need for `maxRecreateAttempts = maxRecreateAttemptsDefault`.

pkg/repl/client.go
```diff
@@ -401,7 +424,15 @@ func (c *Client) readStream(ctx context.Context) {
...
-				c.logger.Warnf("Too many consecutive errors (%d), attempting to recreate streamer", consecutiveErrors)
+				recreateAttempts++
+				c.logger.Warnf("Too many consecutive errors (%d), attempting to recreate streamer (attempt %d/%d)",
+					consecutiveErrors, recreateAttempts, maxRecreateAttempts)
+
+				// Check if we've exceeded the maximum number of recreation attempts
+				if recreateAttempts >= maxRecreateAttempts {
+					panic(fmt.Sprintf("failed to recreate binlog streamer after %d attempts, current position: %v, giving up",
+						recreateAttempts, c.getBufferedPos()))
+				}
```
morgo, line 434: I realize this is in a go routine, so returning errors is challenging, but the repl client runs in a context, and the `cancelFunc` is embedded as `c.cancelFunc()`, so you may be able to simplify and remove the panic?

pkg/repl/client_test.go
```diff
@@ -758,3 +760,110 @@ func TestAllChangesFlushed(t *testing.T) {
...
+	}
+	if !strings.Contains(outputStr, "Failed to recreate streamer") {
+		t.Errorf("Expected to see recreation attempt. Output:\n%s", outputStr)
+	}
+
+	// If we DID get the max attempts panic message, verify it's correct
+	if strings.Contains(outputStr, "failed to recreate binlog streamer after") {
+		if !strings.Contains(outputStr, "giving up") {
+			t.Errorf("Panic message should contain 'giving up'. Output:\n%s", outputStr)
+		}
```
morgo, line 798: The technique here to run a sub-process is very clever :-) But I think it's better to either add a recover for panic, or see if you can get this to work correctly with context canceling. If you pursue context canceling, we need to make sure it works end-to-end because the migration caller should also detect that the replClient has failed and bail.

pkg/repl/client.go
```diff
@@ -213,9 +220,14 @@ func (c *Client) setBufferedPos(pos mysql.Position) {
...
 
 // getBufferedPos returns the buffered position under a mutex.
+// If bufferedPos has no name (not yet set), it falls back to flushedPos.
 func (c *Client) getBufferedPos() mysql.Position {
 	c.Lock()
 	defer c.Unlock()
+	if c.bufferedPos.Name == "" {
+		// If no buffered position, use flushed position
+		return c.flushedPos
+	}
```
morgo, line 229: This looks fine. Good catch.

pkg/repl/client_test.go
tyd: Thanks, feedback applied.

Issue #537, opened 2025-11-20
MoveTables: support copying to a set of resharded tables


morgo:
Currently MoveTables logically copies data from one database schema to another.

However, we have the use case that we want to copy from one source to N destinations based on a shard_key (in vitess language: primary vindex).

PR #540, opened 2025-11-26
move: Add sharded applier

morgo:

Fixes https://github.com/block/spirit/issues/537


pkg/applier/keyrange.go
```diff
@@ -0,0 +1,71 @@
...
+
+	var start, end uint64
+	var err error
+
+	// Parse start
+	if parts[0] == "" {
+		start = 0
+	} else {
+		// Pad to 16 hex chars (64 bits) and parse
+		padded := parts[0] + strings.Repeat("0", 16-len(parts[0]))
```
eeSeeGee, line 42: Should there maybe be a safety check that this is actually [0-9a-f]+?

pkg/table/tableinfo.go
```diff
@@ -43,8 +43,18 @@ type TableInfo struct {
 	statisticsLastUpdated       time.Time
 	statisticsLock              sync.Mutex
 	DisableAutoUpdateStatistics atomic.Bool
+
+	// Sharding configuration (for ShardedApplier)
+	// These are set per-table when using multi-table migrations with different sharding keys
+	VindexColumn string     // Column name to extract and hash (e.g., "user_id")
```
eeSeeGee: Should this be PrimaryVindexColumn? (Or ShardingColumn for the benefit of folks who are less familiar with vitess)?

pkg/applier/sharded.go
```diff
@@ -0,0 +1,808 @@
...
+	shardRows := make([][]rowData, len(a.shards))
+	for _, row := range rows {
+		// Extract the vindex column value
+		if vindexOrdinal >= len(row) {
+			return fmt.Errorf("vindex column ordinal %d exceeds row length %d", vindexOrdinal, len(row))
+		}
+		vindexValue := row[vindexOrdinal]
+
+		// Apply the hash function to get the hash value
+		hashValue, err := vindexFunc(vindexValue)
```
eeSeeGee: Would it be consistent to call this the keyspace ID?

pkg/applier/sharded.go
```diff
@@ -0,0 +1,808 @@
...
+		if err != nil {
+			return fmt.Errorf("vindex function error: %w", err)
+		}
+
+		// Find which shard's key range contains this hash value
+		shardID := -1
+		for i, shard := range a.shards {
```
eeSeeGee, line 211: There's a part of my brain that feels like this is a binary tree situation, but it is perhaps a part that overthinks things

pkg/table/tableinfo.go
morgo: I agree with changing to`ShardingColumn`. I was trying to make it non-vitess specific in other parts of the code. I did not achieve this here, and think its worth modifying it and being agnostic.

pkg/applier/sharded.go
morgo: Yes, but per the first comment - I actually think it should be `vindexFunc` that is renamed (to `HashFunc`), and we continue to call this `hashValue`.

pkg/applier/sharded.go
```diff
@@ -0,0 +1,808 @@
...
+			if lock != nil {
+				if err = lock.ExecUnderLock(ctx, deleteStmt); err != nil {
+					err = fmt.Errorf("failed to execute delete under lock on shard %d: %w", shard.shardID, err)
+				} else {
+					// We can't know the actual affected rows when using lock, so estimate
+					affected = 0
+				}
+			} else {
+				// Execute as a retryable transaction
+				affected, err = dbconn.RetryableTransaction(ctx, shard.writeDB, false, shard.dbConfig, deleteStmt)
```
eeSeeGee, line 580: Is this using the same delete statement on all shards for expediency, or would we actually expect the same keys to be on multiple shards somehow?

morgo: We don't expect the same keys to be on multiple shards.

We have to broadcast only because in the tracking of deletes, we don't track the shard_key (a limitation). We could fix it later, but its a bit of work.

PR #549, opened 2026-01-05
Make write threads for applier configurable

morgo:

Previously we used 40 write threads regardless. This is going to be a problem when running concurrent tests in CI (the appliers are still experimental, so there aren't that many tests for them yet).

I've changed it to be configurable. This changes the signature of `NewSingleTargetApplier` such that it can return errors. This is the same as `NewShardedApplier`.

### DIFF
--- a/pkg/applier/sharded.go
+++ b/pkg/applier/sharded.go
@@ -94,7 +94,7 @@ func NewShardedApplier(targets []Target, cfg *ApplierConfig) (*ShardedApplier, e
 			keyRange:            kr,
 			chunkletBuffer:      make(chan shardedChunklet, defaultBufferSize),
 			chunkletCompletions: make(chan shardedChunkletCompletion, defaultBufferSize),
-			writeWorkersCount:   int32(max(1, cfg.Threads/len(targets))), // Divide workers among shards, minimum 1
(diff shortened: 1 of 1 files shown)


pkg/move/move.go
```diff
@@ -13,6 +13,7 @@ type Move struct {
 	TargetDSN             string        `name:"target-dsn" help:"Where to copy the tables to." default:"spirit:spirit@tcp(127.0.0.1:3306)/dest"`
 	TargetChunkTime       time.Duration `name:"target-chunk-time" help:"How long each chunk should take to copy" default:"5s"`
 	Threads               int           `name:"threads" help:"How many chunks to copy in parallel" default:"2"`
+	WriteThreads          int           `name:"write-threads" help:"How many concurrent write threads to use per target" default:"2"`
```
nonbb, line 16: would it be apparent when errors pop up that we'd need to tweak the default threads up here?

nonbb: i.e. tests timing out/slow?

morgo: It's was not obvious why the tests were failing, but I reduced the concurrency and it worked.

This is similar to scenarios like OOM and CPU load. Programs start performing unreliably.

I tried to consider that it might be something like a race, and thus there's a bug that is fixable. But I don't think that's the case.

PR #550, opened 2026-01-05
refactor/improve appliers

morgo:

Three main improvements in one:

1. There some related bugs in the buffered copier/applier when invalid data is copied (tested by `TestBufferedCopierDataTypeConversionError`):

- `NewDatum` needs to return an error, and not panic if a value is invalid.
   * This unfortunately results in a lot of other changes.
- When an invalid INSERT fails, the copier never finishes. This is fixed by removing incomplete work from pending tasks in the sharded and single_target appliers (`delete(a.pendingWork, completion.workID)`).
- The buffered copier also needs to return an error if there were failed operations.

2. The functionality used to SQL escape strings has been moved to the `Datum` system, and the `EscapeMySQLType` function is removed. This simplifies code so we only need one implementation of escaping.

3. I've also refactored the implementation of `pkg/applier/sharded.go` to use a channel and remove the weird handling of having a different `targetTable` to `sourceTable` structure. We don't support this in the sharded applier, and keep the semantics clean. There is also a 60 second timeout for any individual task, which feels appropriate to have when there are many fan out RPC queries.


pkg/applier/applier.go
```diff
@@ -6,17 +6,19 @@ import (
...
 const (
-	chunkletMaxRows     = 1000        // Maximum number of rows per chunklet
-	chunkletMaxSize     = 1024 * 1024 // Maximum size in bytes per chunklet (1 MiB)
-	defaultBufferSize   = 128         // Size of the shared buffer channel for chunklets
-	defaultWriteWorkers = 40          // Number of write workers
+	chunkletMaxRows     = 1000             // Maximum number of rows per chunklet
+	chunkletMaxSize     = 1024 * 1024      // Maximum size in bytes per chunklet (1 MiB)
+	defaultBufferSize   = 128              // Size of the shared buffer channel for chunklets
+	defaultWriteWorkers = 2                // Number of write workers, default low for tests, but in practice we can use 40+
+	chunkTaskTimeout    = time.Second * 60 // Timeout for any task (copy chunk, delete keys, upsert rows)
```
nonbb, line 21: do you ever foresee this being an issue for larger chunks?

morgo: No. We use a maximum of 1000 rows currently, or 1MB. To exceed this it would have to be very slow (less than 16 rows/s inserted).

So this is just a failsafe that is about 2 orders of magnitude from the norm.

Issue #556, opened 2026-01-09
invalid table id, no corresponding table map


morgo:
Forked from https://github.com/block/spirit/issues/548 / https://github.com/block/spirit/issues/552

These errors are rare, but possible in the testsuite:
```
2026-01-09T18:33:23.1794156Z test-1   | {"Time":"2026-01-09T18:33:23.12873675Z","Action":"output","Package":"github.com/block/spirit/pkg/migration","Test":"TestCheckpointResumeDuringChecksum","Output":"ia\\\\xf7n\\\\v\\\\\\\\\\\\xf0\\\\xbb\\\\xdcv\\\", Err: table id 3713: invalid table id, no corresponding table map event\\ngithub.com/pingcap/errors.Trace\\n\\t/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20250523034308-74f78ae071ee/juju_adaptor.go:15\\ngithub.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).parseEvent\\n\\t/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/replication/binlogsyncer.go:865\\ngithub.com/go-mysql-org/go-mysql/replication.(*BinlogSyncer).onStream\\n\\t/go/pkg/mod/github.com/go-mysql-org/go-mysql@v1.13.0/replication/binlogsyncer.go:818\\nruntime.goexit\\n\\t/usr/local/go/src/runtime/asm_amd64.s:1693\" current_position=\"(binlog.001534, 210693)\"\n"}

```

It has to be a bug in the go-mysql-org not tracking table map events correctly. We need to decide how to handle it (return an error or skip it and rely on the checksum etc).

The current behavior is that it causes the replication syncer to recreate itself on a backoff until it times out, because the event is a poison pill that can't be moved past.

PR #558, opened 2026-01-18
fix replication syncer recreate logic + small test bugs

morgo:

Four minor fixes in one, which fix all the test instability issues known

## Fix 1

Fixes https://github.com/block/spirit/issues/556

We restart the binary log syncer from the start of the file to ensure the table map is fetched.

There is some logic to make sure we don't attempt this if its the initial file we started syncing on, since while subscribing to replication changes is idempotent (we can repeat some changes), we can not go back before our start position because the table could be of a different structure at that point (it's an unlikely scenario, but once we start a migration we guarantee and panic if there are any structure changes).

## Fix 2

There looks to be a race discovered in the tests:

```
2026-01-17T18:06:27.2872826Z test-1           | 2026/01/17 17:59:26 INFO syncer is closed
2026-01-17T18:06:27.2873049Z test-1           | --- FAIL: TestSecondaryEngineAttribute (1.61s)
2026-01-17T18:06:27.2873144Z test-1           |     migration_test.go:674: 
2026-01-17T18:06:27.2873308Z test-1           |         	Error Trace:	/app/pkg/migration/migration_test.go:674
2026-01-17T18:06:27.2873436Z test-1           |         	Error:      	Received unexpected error:
2026-01-17T18:06:27.2873543Z test-1           |         	            	invalid connection
2026-01-17T18:06:27.2873675Z test-1           |         	Test:       	TestSecondaryEngineAttribute
```
And:

```
2026-01-17T18:06:21.3131661Z test-1           | 2026/01/17 18:04:02 INFO syncer is closed
2026-01-17T18:06:21.3131816Z test-1           | --- FAIL: TestSecondaryEngineAttribute (5.07s)
2026-01-17T18:06:21.3131919Z test-1           |     migration_test.go:674: 
2026-01-17T18:06:21.3132082Z test-1           |         	Error Trace:	/app/pkg/migration/migration_test.go:674
2026-01-17T18:06:21.3132205Z test-1           |         	Error:      	Received unexpected error:
2026-01-17T18:06:21.3132402Z test-1           |         	            	context canceled
2026-01-17T18:06:21.3132543Z test-1           |         	Test:       	TestSecondaryEngineAttribute
```

Sometimes the timer is not stopped immediately, so the change is we force stop it as soon as the Exec query is finished. We also add some safety logic to a minimum timer of 0.9 seconds. This should address these racey tests.

## Fix 3

The checksum does not retry, leading to this error:

```
2026-01-18T19:49:33.7313544Z test-1           | --- FAIL: TestDeferCutOverE2E (85.55s)
2026-01-18T19:49:33.7313832Z test-1           |     runner_test.go:1695: 
2026-01-18T19:49:33.7314104Z test-1           |         	Error Trace:	/app/pkg/migration/runner_test.go:1695
2026-01-18T19:49:33.7314360Z test-1           |         	            				/usr/local/go/src/runtime/asm_amd64.s:1693
2026-01-18T19:49:33.7314574Z test-1           |         	Error:      	Received unexpected error:
2026-01-18T19:49:33.7315461Z test-1           |         	            	timed out waiting to catch up to source position: (mysql-bin.001133, 1710), current position is: (mysql-bin.000871, 12491354)
2026-01-18T19:49:33.7315669Z test-1           |         	Test:       	TestDeferCutOverE2E
2
```

This looks to be a regression from https://github.com/block/spirit/pull/507

## Fix 4

The concurrent tests can be unreliable because golang defaults to the same number of concurrent tests as cores (16). But since each test is already multi-threaded and the runner also hosts MySQL, this is too much.

I've reduced the concurrency to 4, and fixed https://github.com/block/spirit/issues/559 so that we are not just overloading MySQL leading to thrashing/false positive cases.

The test timeout also increases to 10 minutes because we are getting close to that.

### DIFF
--- a/pkg/repl/client.go
+++ b/pkg/repl/client.go
@@ -22,13 +22,7 @@ import (
 )
 
 const (
-	// binlogTrivialThreshold is the number of pending changes below which
-	// flushing is considered up to date. We can never quite get to zero, so we have
-	// to allow a certain number:
-	//   - Too low: we never get to cutover
-	//   - Too high: we take too long to cutover and might timeout, since we will only
-	//     ever hold the tablelock for 30s max (DefaultTimeout)
-	binlogTrivialThreshold = 1000
+	binlogTrivialThreshold = 10000
 	// DefaultBatchSize is the number of rows in each batched REPLACE/DELETE statement.
 	// Larger is better, but we need to keep the run-time of the statement well below
 	// dbconn.maximumLockTime so that it doesn't prevent copy-row tasks from failing.


pkg/dbconn/dbconn.go
```diff
@@ -206,16 +206,18 @@ func ForceExec(ctx context.Context, db *sql.DB, tables []*table.TableInfo, dbCon
...
+	duration := time.Duration(threshold) * time.Second
+	timer := time.AfterFunc(duration, func() {
 		err := KillLockingTransactions(ctx, db, tables, dbConfig, logger, []int{connId})
 		if err != nil {
 			return // just return, we can't do much more here
 		}
 	})
-	defer timer.Stop()
 	_, err = trx.ExecContext(ctx, sqlescape.MustEscapeSQL(stmt, args...))
+	timer.Stop() // Stop the timer immediately after the DDL completes
```
eeSeeGee, line 220: Without a defer is there any risk the timer spins off into forever?

eeSeeGee (approved):

pkg/dbconn/dbconn.go
morgo: There shouldn't be, since there's only one code path.

Go calls defers deterministically (LIFO), so in theory it should be functionality neutral to make this change. I don't understand why this should fix the race TBH. I am wondering if there is a thread scheduling difference for defers.

Issue #534, opened 2025-11-14
Add checks system of at least configuration check for MoveTables


morgo:
MoveTables uses buffered copies with require `binlog_row_image = FULL`. We check for this in migrations, but not in moves.

There needs to be some way of preventing running if it is not supported.

PR #560, opened 2026-01-20
Add move checks 

morgo:

Fixes https://github.com/block/spirit/issues/534

The checks system in migrations moves some of the non-mutating inspections of configuration, state, options into a separate package. Because it depends on `Resources` which are specific to migrations, it is hard to repurpose this to moves. So I've moved it to  pkg/migration/check/ and created a new pkg/move/check with similar semantics.

Full list of changes:

* Moves isMinimalRBRTestRunner to testutils.IsMinimalRBRTestRunner and uses it for all packages as a way to detect if buffered tests should run (replacing move.settingsCheck).
* Moves pkg/check to pkg/migration/check/
* Creates pkg/move/check, and adds a new check for configuration
* Functionality from pkg/move/runner.go to determine if a resume is possible or a new migration is moved into the checks system (target_state and resume_state).

### DIFF
--- a/pkg/move/check/target_state.go
+++ b/pkg/move/check/target_state.go
@@ -2,6 +2,7 @@ package check
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"log/slog"
 	"slices"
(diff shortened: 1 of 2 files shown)


pkg/move/check/resume_state_test.go
```diff
@@ -0,0 +1,252 @@
...
+			SourceTables: []*table.TableInfo{sourceTable},
+		}
+		err := resumeStateCheck(t.Context(), r, slog.Default())
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "checkpoint table")
+		assert.Contains(t, err.Error(), "does not exist")
+	})
+
+	// Create checkpoint table for remaining tests
+	_, err = sourceDB.ExecContext(t.Context(), `CREATE TABLE resume_src._spirit_checkpoint (
```
eeSeeGee, line 72: Is it potentially annoying for future you to not create the checkpoint with the actual checkpoint code?

pkg/move/check/resume_state.go
```diff
@@ -0,0 +1,83 @@
...
+		return errors.New("no source tables available for resume validation")
+	}
+
+	// Check 1: Verify checkpoint table exists on source
+	checkpointTableName := "_spirit_checkpoint"
+	var checkpointExists int
+	err := r.SourceDB.QueryRowContext(ctx,
+		"SELECT 1 FROM information_schema.TABLES WHERE table_schema = ? AND table_name = ?",
+		r.SourceConfig.DBName, checkpointTableName).Scan(&checkpointExists)
+	if err == sql.ErrNoRows {
```
eeSeeGee, line 38: Does anything in the checkpoint table itself need to be checked at this stage?

pkg/move/check/resume_state_test.go
morgo: There will be a cyclical dependency because that would require `runner.go:createCheckpointTable()` to be called, which needs to import the `move` package.

If I extracted `createCheckpointTable()` to its own package for checkpoints that would make this possible - but there is not too much benefit to that right now.

pkg/move/check/resume_state.go
morgo: I _could_, but restoring the checkpoint is also able to return an error in `pkg/move/runner.go:resumeFromCheckpoint()`

My preference is not to because if I change the checkpoint format (it happens) the code is non-dry. This is because the `check` should only check and is not expected to load or mutate structures.

If I move checkpoints to a separate package, that would be a different argument though; they could call the same helper.

pkg/move/check/target_state.go
```diff
@@ -0,0 +1,112 @@
...
+
+// validateExistingTargetTable checks that an existing table in the target database
+// is empty (zero rows) and has a schema that matches the source table.
+// This allows move-tables to work with pre-created tables, which is necessary
+// for declarative schema management workflows where tables must be tracked
+// in the migrations directory before data is moved.
+func validateExistingTargetTable(ctx context.Context, target applier.Target, tableName string, targetIndex int, sourceTable *table.TableInfo, logger *slog.Logger) error {
+	// Check 1: Table must have zero rows
+	var rowCount int64
+	err := target.DB.QueryRowContext(ctx, fmt.Sprintf("SELECT COUNT(*) FROM `%s`.`%s`", target.Config.DBName, tableName)).Scan(&rowCount)
```
eeSeeGee: I'm wondering if there's scenarios where this check is running erroneously on an already filled table and select count(*) unexpectedly takes forever. i.e. you're actually just interested in "Is there ONE row"

morgo: I think this is a reasonable concern, I'll fix it.

PR #579, opened 2026-01-27
Add documentation for replication client

morgo:

Provides documentation for the replication client package.


pkg/repl/README.md
```diff
@@ -0,0 +1,175 @@
...
+**Limitations:**
+- **Single-threaded flushing**: Must process changes sequentially, limiting throughput
+- **Poor deduplication**: Stores all intermediate states, only merging consecutive operations
+- **Higher memory usage**: Can accumulate many operations for the same key
+- **No support for optimizations**: Key above and key below watermark optimizations are not possible
+
+**Example scenario:**
+```
+Binlog events:  INSERT(id=1), UPDATE(id=1), DELETE(id=2), UPDATE(id=1)
+Delta queue:    [INSERT(1), UPDATE(1), DELETE(2), UPDATE(1)]
```
aparajon: Should we be showing a non-integer type primary key here?

pkg/repl/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+// This check should be redundant, but we verify everything is applied
+if !client.AllChangesFlushed() {
+    return errors.New("changes still pending")
+}
+
+// Safe to cutover now
+```
+
+The `client.Flush()` will retry in a loop until the number of pending changes is considered trivial (currently <10K). It is important to handle errors correctly here, because `FlushUnderTableLock` may fail if it can't flush the pending changes fast enough. This is your cue to abandon the cutover operation for now, and try again when the server is under less load.
```
aparajon, line 164: Is there force-cutover behavior that can help here?

aparajon (approved):

pkg/repl/README.md
morgo: Force cutover applies to the lock acquisition. This threshold is about once we have the lock, what we can reasonably do before we decide that we're taking too long.

The `repl.FlushUnderTableLock()` call requires that you have to have pre-created the lock, which is done here:
https://github.com/block/spirit/blob/35bd921c81641d1b961a48585822fb432cae4172/pkg/migration/cutover.go#L122-L129

(This is where force / non-force matters).

pkg/repl/README.md
morgo: Yes, good call out.

PR #580, opened 2026-01-27
Bug fixes + Improve test coverage of sharded move

morgo:

This adds tests for sharded move operations and fixes some bugs:

- The sharded applier is not idempotent like the single_target applier is w.r.t. Start().
- The checksum was not using Non-Generated Columns
- The applier mishandled extracting the shard key from a table with generated columns
- The checksum was using chunk.NewTable.QuotedName, which includes the schema name (leading to failures).

### DIFF
--- a/pkg/migration/runner.go
+++ b/pkg/migration/runner.go
@@ -248,6 +248,10 @@ func (r *Runner) Run(ctx context.Context) error {
 	// The watermark optimizations can prevent some keys from being flushed,
 	// which would cause flushedPos to not advance, leading to a mismatch
 	// with bufferedPos and causing AllChangesFlushed() to return false.
+	// Note: this line is not strictly required for correctness, since
+	// FlushUnderLock() will now ignore watermark optimizations explicitly,
+	// but having the non-underlock flushes also disable these optimizations
+	// is cleaner.
 	r.replClient.SetWatermarkOptimization(false)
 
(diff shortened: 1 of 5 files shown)


pkg/move/move_sharded_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for i := 1; i <= 100; i++ {
+		_, err := db.ExecContext(t.Context(), "INSERT INTO source_sharded.users (id, user_id, name, `values`) VALUES (?, ?, ?, ?)",
+			i, i, fmt.Sprintf("User %d", i), fmt.Sprintf("user%d@example.com", i))
+		assert.NoError(t, err)
+	}
+
+	// Create target databases
+	testutils.RunSQL(t, `DROP DATABASE IF EXISTS dest_sharded_0`)
+	testutils.RunSQL(t, `CREATE DATABASE dest_sharded_0`)
```
aparajon, line 65: Can we make a test util that does both of these together?

morgo: Maybe as a followup?

I've thought about adding a helper for this which also registers a t.Cleanup() task so if the test exits successfully it also drops the database or table it created. We are using this pattern everywhere through the tests currently.

aparajon: Sounds good to me. This was mostly a readability suggestion since it would reduce some lines of test code

PR #581, opened 2026-01-27
only flush binlog if buffered position has not advanced

### LINKED ISSUE #565: BlockWait is too aggressive in flushing binary logs
From:

https://github.com/block/spirit/blob/8ae94169c4662e64b138b2de990019196c024941/pkg/repl/client.go#L938-L967

This loop calls `flush binary logs` to advance the binary log (because there are some events which won't count in `readStream()`, and we don't want an off-by-1 where the synced position is behind the source's position but isn't actually).

We don't need to advance it that frequently though. The assumption was a 100ms sleep would be OK.

We have evidence that that doesn't appear to be the case, and had a system where the `flush binary logs` calls were 500ms+ in average latency. So assuming they block new changes being written, it limits the ability to flush changes.

I propose we change it to:

1. either 1 or 2s (better, and 1 line change)
2. we only call `flush binary logs` **if** c.getBufferedPos() hasn't advanced since our last loop (best, still easy to implement).

(2) is obviously better. I don't see any downsides.

eduardo-landin-mestre:

Fixes https://github.com/block/spirit/issues/565

As suggested, these changes decrease the frequency with which spirit flushes binary logs by:
- only calling `FLUSH BINARY LOGS` if the buffered position hasn't advanced


pkg/repl/client.go
```diff
@@ -950,19 +952,32 @@ func (c *Client) BlockWait(ctx context.Context) error {
...
+				c.flushedBinlogs++
 			}
+			prevPos = currPos
+			first = false
+
 			if c.getBufferedPos().Compare(targetPos) >= 0 {
 				return nil // we are up to date!
 			}
 			// We are not caught up yet, so we need to wait.
-			time.Sleep(100 * time.Millisecond)
```
morgo, line 965: I think we need to revert this to 100ms, just based on the tests taking a lot longer and you often hitting https://github.com/block/spirit/issues/587 versus most other PRs rarely hitting it (I will try to fix it, but I think for the happy path where the binlog is advancing 100ms is not an issue).

pkg/repl/client_test.go
```diff
@@ -440,21 +440,40 @@ func TestBlockWait(t *testing.T) {
...
 	testutils.RunSQL(t, "ANALYZE TABLE blockwaitt1")
 	testutils.RunSQL(t, "ANALYZE TABLE blockwaitt1")
 
 	// We wait up to 10s again.
 	// although it should be quick.
+	client.flushedBinlogs = 0
+	start = time.Now()
 	assert.NoError(t, client.BlockWait(t.Context()))
+	target = int(time.Since(start).Seconds()) + 1
+	assert.LessOrEqual(t, client.flushedBinlogs, target)
```
morgo: I like the way you've done this test, but I think you can do one better:

- Verify that time.Since(start) is greater than that 100ms loop (maybe move it to a constant).
- Start a go routine that is inserting into a completely unrelated table before line 443 every 80ms or so. This should ensure that flushedBinlogs always equals zero.
- Assert that flushedBinlogs is exactly zeroo, not this target.

This ensures `flush binary logs`-less advancing, which is the core functionality this offers.

eduardo-landin-mestre: I see 🤔 

> Verify that time.Since(start) is greater than that 100ms loop (maybe move it to a constant)

I think we can achieve the same result by removing the `flushedBinlogs` increment from the `getCurrentBinlogPosition` function. If we wait 100ms we risk `BlockWait` exiting on the very first iteration. I tried this in [8c4f62b](https://github.com/block/spirit/pull/581/commits/8c4f62bd3d8896b6c48400bafac07054bb2ad504)

I also updated the test so that it
1/ kicks off the go routine that inserts into the unrelated table every 100 nanoseconds
2/ sleeps for 3s
3/ starts `BlockWait`
4/ checks that `flushedBinlogs` is still 0 by the time `BlockWait` exits

The combination of 1/ and 2/ seems to be enough to get `BlockWait` to perform 2 iterations of the main for loop (although not guaranteed).

Test output:
```
=== RUN   TestBlockWait
2026/02/01 16:48:38 INFO create BinlogSyncer config="{ServerID:48914607 Flavor:mysql Host:127.0.0.1 Port:8033 User:tsandbox Password: Localhost: Charset: SemiSyncEnabled:false RawModeEnabled:false TLSConfig:0xc00044e000 ParseTime:false TimestampStringLocation:UTC UseDecimal:false UseFloatWithTrailingZero:false RecvBufferSize:0 HeartbeatPeriod:0s ReadTimeout:0s MaxReconnectAttempts:0 DisableRetrySync:false VerifyChecksum:false DumpCommandFlag:0 Option:<nil> Logger:0xc000044280 Dialer:0x10051d680 RowsEventDecodeFunc:<nil> TableMapOptionalMetaDecodeFunc:<nil> DiscardGTIDSet:false EventCacheCount:10240 FillZeroLogPos:false SynchronousEventHandler:<nil>}"
2026/02/01 16:48:38 INFO begin to sync binlog from position position="(binlog.000944, 157)"
2026/02/01 16:48:38 INFO Connected to server flavor=mysql version=8.0.33
2026/02/01 16:48:38 INFO rotate to next binlog file=binlog.000944 position=157
2026/02/01 16:48:41 INFO rotate to next binlog file=binlog.000945 position=4
2026/02/01 16:48:41 INFO rotate to next binlog file=binlog.000945 position=4
2026/02/01 16:48:41 INFO waiting to catch up to source position target_position="(binlog.000945, 741)" current_position="(binlog.000945, 449)"
[BlockWait] still need to catch up, next iter...
[BlockWait] exiting....
```

pkg/repl/client.go
eduardo-landin-mestre: Fixed in [8c4f62b](https://github.com/block/spirit/pull/581/commits/8c4f62bd3d8896b6c48400bafac07054bb2ad504) 🫡

morgo (approved):

PR #590, opened 2026-02-05
fix for issue-479: implement key{Above|Below}Watermark optimisations for the composite chunker

### LINKED ISSUE #479: Implement keyAboveWatermark for composite chunker with restrictions
It is documented that key{Above|Below}Watermark only applies to the optimistic chunker. We should add it to the composite chunker with restrictions:

- Only applies to key[0].
- Only applies when key[0] is numeric (initially, we could relax this in future).

The relevant code is:

https://github.com/block/spirit/blob/30d3e41c050adb7eeb0bb215d690c81086b5ff31/pkg/table/chunker_composite.go#L480-L490

The advantage of implementing this feature is as follows:
1. For KeyAbove it eliminates some changes needing to be applied. This may be less important compared to auto_inc tables, because users are less likely to be inserting/modifying data at the end of the key range.
2. For KeyBelow it delays applying changes until after the copier is guaranteed to be past that point. This can be really useful in preventing contention cases (deadlocks, lock waits), which have been observed a few times between applier and copier.

Kiran01bm:
## Summary
Implements watermark optimizations for composite chunker. Watermark optimizations are like performance hints, not correctness guarantees. The changes in this PR are conservative by design i.e when in doubt, keep events or buffer (inefficient but safe and better than not having the optimisation)

### Key Changes:
1. Prevents deadlocks and lock wait timeouts between copier and applier threads during migration by buffering binlog events that conflict with active copy regions.
2. Performance Benefit: Reduces unnecessary event processing by discarding events for key ranges not yet copied.
3. Also has a improvement/fix for Optimistic chunker : Fixed `KeyBelowLowWatermark` error handling which prevents potential copier/applier contention during error conditions

### Scope:
Optimizations work for **all comparable types** (numeric, string, temporal) in `key[0]`, with conservative fallback for binary types.
- **Multi-layer safety**: Optimizations provide performance hints, checksum validates correctness using MySQL comparison, and retry mechanism fixes any discrepancies
- **VARCHAR**: Go string comparison differs from MySQL collation, but checksum catches and fixes
- **DATETIME**: Go temporal comparison may differ from MySQL semantics, but checksum validates
- **Numeric**: Exact match between Go and MySQL comparison
- **Binary types**: Use conservative behavior (buffer everything) to avoid comparison issues

## Related Issues
- Fixes #479 - KeyAboveWatermark for composite chunker
- Related to #478 - Flush correctness issue (mutex during flush) - also helps mitigate performance overhead by reducing the number of events that need to be flushed.

### DIFF
--- a/pkg/migration/runner_test.go
+++ b/pkg/migration/runner_test.go
@@ -1038,7 +1038,7 @@ func TestE2EBinlogSubscribingCompositeKeyVarchar(t *testing.T) {
 	// Copy remaining chunks
 	for {
 		chunk, err := m.copyChunker.Next()
-		if err == table.ErrTableIsRead {
+		if errors.Is(err, table.ErrTableIsRead) {
 			break
 		}
 		assert.NoError(t, err)
@@ -1184,7 +1184,7 @@ func TestE2EBinlogSubscribingCompositeKeyCollation(t *testing.T) {
 	// Copy remaining chunks
 	for {
 		chunk, err := m.copyChunker.Next()
-		if err == table.ErrTableIsRead {
+		if errors.Is(err, table.ErrTableIsRead) {
 			break
 		}
 		assert.NoError(t, err)
@@ -1350,7 +1350,7 @@ func TestE2EBinlogSubscribingCompositeKeyBinary(t *testing.T) {
 	// Copy remaining chunks (this will copy the 0x50xxxx and 0x61xxxx ranges)
 	for {
 		chunk, err := m.copyChunker.Next()
-		if err == table.ErrTableIsRead {
+		if errors.Is(err, table.ErrTableIsRead) {
 			break
 		}
 		assert.NoError(t, err)
@@ -1364,14 +1364,14 @@ func TestE2EBinlogSubscribingCompositeKeyBinary(t *testing.T) {
 	// VERIFY: Row should NOT be missing (binary comparison is exact)
 	t.Log("→ Verifying row status BEFORE checksum...")
 	var targetCountBefore int
-	err = m.db.QueryRow(`SELECT COUNT(*) FROM _e2et_binary_new WHERE name = 0x500050 AND id = 9999`).Scan(&targetCountBefore)
+	err = m.db.QueryRowContext(t.Context(), `SELECT COUNT(*) FROM _e2et_binary_new WHERE name = 0x500050 AND id = 9999`).Scan(&targetCountBefore)
 	assert.NoError(t, err)
 	assert.Equal(t, 1, targetCountBefore, "Row should be present - it was copied in later chunks")
 	t.Log("✓ CONFIRMED: Row is PRESENT in target table (copied in later chunks)")
 
 	// Verify row exists in source
 	var sourceCount int
-	err = m.db.QueryRow(`SELECT COUNT(*) FROM e2et_binary WHERE name = 0x500050 AND id = 9999`).Scan(&sourceCount)
+	err = m.db.QueryRowContext(t.Context(), `SELECT COUNT(*) FROM e2et_binary WHERE name = 0x500050 AND id = 9999`).Scan(&sourceCount)
 	assert.NoError(t, err)
 	assert.Equal(t, 1, sourceCount, "Source table should have the row")
 
@@ -1383,13 +1383,13 @@ func TestE2EBinlogSubscribingCompositeKeyBinary(t *testing.T) {
 	// VERIFY: Row still exists (checksum should not need to fix anything)
 	t.Log("→ Verifying row status AFTER checksum...")
 	var targetCountAfter int
-	err = m.db.QueryRow(`SELECT COUNT(*) FROM _e2et_binary_new WHERE name = 0x500050 AND id = 9999`).Scan(&targetCountAfter)
+	err = m.db.QueryRowContext(t.Context(), `SELECT COUNT(*) FROM _e2et_binary_new WHERE name = 0x500050 AND id = 9999`).Scan(&targetCountAfter)
 	assert.NoError(t, err)
 	assert.Equal(t, 1, targetCountAfter, "Row should still be present after checksum")
 
 	// Verify the data is correct
 	var data string
-	err = m.db.QueryRow(`SELECT data FROM _e2et_binary_new WHERE name = 0x500050 AND id = 9999`).Scan(&data)
+	err = m.db.QueryRowContext(t.Context(), `SELECT data FROM _e2et_binary_new WHERE name = 0x500050 AND id = 9999`).Scan(&data)
 	assert.NoError(t, err)
 	assert.Equal(t, "INSERTED_DURING_MIGRATION", data, "Data should match")
 
@@ -1504,7 +1504,7 @@ func TestE2EBinlogSubscribingCompositeKeyDateTime(t *testing.T) {
 	// Copy remaining chunks
 	for {
 		chunk, err := m.copyChunker.Next()
-		if err == table.ErrTableIsRead {
+		if errors.Is(err, table.ErrTableIsRead) {
 			break
 		}
 		assert.NoError(t, err)
--- a/pkg/table/chunker_composite_test.go
+++ b/pkg/table/chunker_composite_test.go
@@ -817,12 +817,12 @@ func TestCompositeChunkerWatermarkOptimizations(t *testing.T) {
 
 	// Verify data was inserted correctly
 	var count int
-	err = db.QueryRow("SELECT COUNT(*) FROM compositewatermarkopt_t1").Scan(&count)
+	err = db.QueryRowContext(t.Context(), "SELECT COUNT(*) FROM compositewatermarkopt_t1").Scan(&count)
 	require.NoError(t, err, "Failed to count rows")
 	require.Equal(t, 3000, count, "Expected 3000 total rows")
 
 	var countA1 int
-	err = db.QueryRow("SELECT COUNT(*) FROM compositewatermarkopt_t1 WHERE a = 1").Scan(&countA1)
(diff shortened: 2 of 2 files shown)


pkg/table/chunker_composite.go
```diff
@@ -481,16 +481,85 @@ func (t *chunkerComposite) Progress() (uint64, uint64, uint64) {
...
+	// Use CompareGreaterThanOrEqual which supports all types (numeric, string, temporal)
+	return keyDatum.CompareGreaterThanOrEqual(t.chunkPtrs[0])
 }
 
-// KeyBelowLowWatermark is not yet supported for composite chunker.
+// KeyBelowLowWatermark checks if a key is below the low watermark.
+// This optimization works with comparable types in key[0] (first column): numeric, string, temporal.
+// Binary types (VARBINARY, BLOB, BINARY) use conservative behavior (always return true, buffer everything).
+// Note: Watermark optimizations are disabled before checksum phase (see runner.go),
+// so there is no risk of checksum corruption even if collation comparison differs slightly.
```
morgo: This comment is not correct (nor the implementation):

- Binary string types are possible to be the same as MySQL.
- Varchars however are non-comparable, and very difficult

Here is an easy test:
```
mysql> select 'aa' > 'AA' ;
+-------------+
| 'aa' > 'AA' |
+-------------+
|           0 |
+-------------+
1 row in set (0.00 sec)

mysql> select _binary 'aa' > _binary 'AA' ;
+-----------------------------+
| _binary 'aa' > _binary 'AA' |
+-----------------------------+
|                           1 |
+-----------------------------+
1 row in set (0.00 sec)
```

This is because 'aa' = 'AA' using the default collation (it is not greater or less than), but if you take the binary values:

0x41 = A
0x61 = a

So when we are copying the key using chunked ranges like key between X and Y, we can replicate the binary order from spirit and will always be safe.

The collated order is basically impossible to implement in spirit. Collated orders get even more complicated than this, consider this in `utf8mb4_czech_ci ` etc:

"h" < "ch" < "i" (because "ch" is a distinct letter in the Czech alphabet)

pkg/table/chunker_optimistic.go
```diff
@@ -568,15 +568,15 @@ func (t *chunkerOptimistic) KeyBelowLowWatermark(key0 any) bool {
...
-		return false
+		return true
 	}
 	watermarkDatum, err := NewDatum(t.watermark.UpperBound.Value[0].Val, t.chunkPtr.Tp)
 	if err != nil {
-		// If we can't convert the watermark, return false to be safe
+		// If we can't convert the watermark, return true to be safe (buffer everything)
 		t.logger.Error("failed to create watermarkDatum in KeyBelowLowWatermark", "error", err)
-		return false
+		return true
```
morgo: I think this implementation should be changed in this PR?

pkg/table/chunker_composite.go
```diff
@@ -481,16 +481,85 @@ func (t *chunkerComposite) Progress() (uint64, uint64, uint64) {
...
+	keyDatum, err := NewDatum(key0, t.chunkPtrs[0].Tp)
+	if err != nil {
+		// If we can't convert the key, return false to be safe (don't discard the row)
+		t.logger.Error("failed to create datum in KeyAboveHighWatermark", "key", key0, "error", err)
+		return false
+	}
+
+	// Check if key is greater than or equal to the current chunkPtr[0]
+	// Use CompareGreaterThanOrEqual which supports all types (numeric, string, temporal)
+	return keyDatum.CompareGreaterThanOrEqual(t.chunkPtrs[0])
```
morgo: I think we can probably extend GreaterThanOrEqual rather than add a new function.

pkg/table/chunker_composite.go
```diff
@@ -481,16 +481,85 @@ func (t *chunkerComposite) Progress() (uint64, uint64, uint64) {
...
+	if err != nil {
+		// If we can't convert the key, return false (assume not below watermark)
+		t.logger.Error("failed to create keyDatum in KeyBelowLowWatermark", "key", key0, "error", err)
+		return false
+	}
+
+	// Key is below watermark if watermark.UpperBound[0] > key
+	// Use CompareGreaterThan which supports all types (numeric, string, temporal)
+	// t.watermark.UpperBound represents the maximum value that has been safely copied in that chunk
+	return t.watermark.UpperBound.Value[0].CompareGreaterThan(keyDatum)
```
morgo: Same comment as CompareGreaterThanOrEqual; it should be possible to extend the existing function

pkg/table/chunker_composite_test.go
```diff
@@ -771,3 +771,430 @@ func TestCompositeChunkerReset(t *testing.T) {
...
+			SELECT n + 1 FROM seq WHERE n < 1000
+		)
+		SELECT 2, n, 1 FROM seq`)
+	testutils.RunSQL(t, `INSERT INTO compositewatermarkopt_t1 (a, b, c)
+		WITH RECURSIVE seq AS (
+			SELECT 1 AS n
+			UNION ALL
+			SELECT n + 1 FROM seq WHERE n < 1000
+		)
+		SELECT 3, n, 1 FROM seq`)
```
morgo, line 808: This is functionally correct, but doesn't perform well. If you need data fast use a cartesian product:
```
INSERT INTO compositewatermarkopt_t1 SELECT .. FROM compositewatermarkopt_t1 a JOIN compositewatermarkopt_t1 b JOIN c;
```

pkg/table/chunker_composite_test.go
```diff
@@ -771,3 +771,430 @@ func TestCompositeChunkerReset(t *testing.T) {
...
+	assert.True(t, comp.KeyBelowLowWatermark(1))
+	assert.True(t, comp.KeyBelowLowWatermark(100))
+
+	assert.NoError(t, comp.Close())
+}
+
+// TestCompositeChunkerWatermarkNonNumeric tests that watermark optimizations
+// work correctly with non-numeric (VARCHAR) first columns.
+// Since watermark optimizations are disabled before checksum (see runner.go),
+// there's no risk of corruption even if collation comparison differs slightly.
```
morgo, line 935: There's no risk of corruption, but we should be able to create a testcase that:

1. Misses an initial change because of collation order.
2. This is fixed by the checksum

We should be able to demo the first part of this in this package because you can step through it chunk by chunk.

We can then create an E2E version of this in migration/runner.go showing the checksum fix.

pkg/table/chunker_optimistic.go
Kiran01bm: just trying to understand your comment - i'll ask a 2 part question - a> do you agree with this change ? and b> if yes, should i move this to a standalone pr ?

pkg/table/chunker_composite_test.go
Kiran01bm: good point, i've now added an e2e test to explicitly demonstrate this in migration/runner.go (and a unit test as well) - i've also attempted to make the test deterministic - so this is how the flow looks like:

```
→ Copy first chunk (watermark: 'KEY0167') 
→ Insert 'key0050' 
→ KeyAboveHighWatermark: `go` says 'key' > 'KEY' → Event discarded 
→ Row missing from target 
→ Checksum detects 
→ Checksum fixes 
→ Test validates
```

pkg/table/chunker_composite.go
Kiran01bm: right, thanks for the explanation - i've expanded the watermark optimization to binary types and added tests as well and made them deterministic as well to prevent race-conditions.. ptal..

pkg/table/chunker_optimistic.go
morgo: Sorry, should not be changed. Copilot is also picking this up in its comment.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/go-ini/ini v1.67.0
 	github.com/go-mysql-org/go-mysql v1.13.0
 	github.com/go-sql-driver/mysql v1.9.3
+	github.com/google/uuid v1.6.0
```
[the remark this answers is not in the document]
Kiran01bm: this is correct as go's `uuid` is being used in packages like runner_test.go to generate uuid tokens..

pkg/table/chunker_optimistic.go
```diff
@@ -568,15 +568,15 @@ func (t *chunkerOptimistic) KeyBelowLowWatermark(key0 any) bool {
 	// upperBound to the key to decide what to return.
 	keyDatum, err := NewDatum(key0, t.chunkPtr.Tp)
 	if err != nil {
-		// If we can't convert the key, return false to be safe (assume it's not below watermark)
+		// If we can't convert the key, return true to be safe (buffer everything)
 		t.logger.Error("failed to create keyDatum in KeyBelowLowWatermark", "key", key0, "error", err)
-		return false
+		return true
```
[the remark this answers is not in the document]
Kiran01bm: sounds good - thank you - this is changed now..

pkg/migration/runner_test.go
```diff
@@ -926,6 +940,597 @@ func TestE2EBinlogSubscribingCompositeKey(t *testing.T) {
...
+		t.Log("✓ SUCCESS: Row was MISSING before checksum, PRESENT after checksum")
+		t.Log("✓ This proves KeyAboveHighWatermark discarded the event due to collation difference")
+		t.Log("✓ Checksum detected and fixed the missing row")
+	}
+
+	assert.NoError(t, m.Close())
+
+	t.Log("✓ Row was initially DISCARDED by KeyAboveHighWatermark due to Go vs MySQL collation difference")
+	t.Log("✓ Checksum phase detected the missing row and fixed the discrepancy")
+	t.Log("✓ Migration completed successfully with all data intact")
```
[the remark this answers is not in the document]
Kiran01bm: thanks for highlighting - i'd like to remove it altogether.

Issue #592, opened 2026-02-09
DSN construction fails with passwords containing special characters


deanbaker:
# Problem

When a MySQL password contains special characters such as @, ?, &, or / (e.g. AWS IAM authentication tokens), Spirit fails to connect because the DSN string is malformed.

The DSN was being constructed via fmt.Sprintf and manual string concatenation in both `pkg/dbconn/conn.go` (newDSN) and `pkg/migration/runner.go` (dsn()). Characters like @ in a password break the `user:password@tcp(host)/db` format since `@` is used as a delimiter.

## Fix
Use the go-sql-driver/mysql library's structured Config type and its ParseDSN() / FormatDSN() methods instead of manual string construction.
This ensures passwords are properly escaped regardless of content.

## Changes:
  - `pkg/dbconn/conn.go`: Refactored newDSN() to set TLS, session variables, and driver options directly on the parsed `mysql.Config` struct, then call `cfg.FormatDSN()`
  - `pkg/migration/runner.go`: Refactored `dsn()` to use `mysql.NewConfig()` + `cfg.FormatDSN()`

PR #593, opened 2026-02-09
Fix DSN construction to support passwords with special characters

deanbaker:
## Summary

Fixes https://github.com/block/spirit/issues/592

- DSN construction via `fmt.Sprintf` and manual string concatenation breaks when passwords contain special characters (`@`, `?`, `&`, `/`) — common in AWS IAM authentication tokens
- Refactored `newDSN()` in `pkg/dbconn/conn.go` and `dsn()` in `pkg/migration/runner.go` to use the `go-sql-driver/mysql` library's structured `Config` type with `ParseDSN()` / `FormatDSN()` instead of manual string building
- Added test cases covering passwords with `@`, `?`, `&`, and full AWS IAM-style tokens to verify correct round-tripping

## Test plan

- [x] Unit tests added for passwords with special characters (`@`, multiple `@`, mixed special chars, empty password, AWS IAM-style token)
- [ ] Existing tests pass (`go test ./pkg/dbconn/... ./pkg/migration/...`)
- [ ] Manual verification with an IAM auth token DSN

### DIFF
--- a/pkg/dbconn/conn.go
+++ b/pkg/dbconn/conn.go
@@ -281,8 +281,6 @@ func newDSN(dsn string, config *DBConfig) (string, error) {
 	// See also: https://github.com/go-sql-driver/mysql?tab=readme-ov-file#rejectreadonly
 	cfg.RejectReadOnly = true
 	cfg.InterpolateParams = config.InterpolateParams
-	// Allow mysql_native_password authentication
-	cfg.AllowNativePasswords = true
 	// Allow cleartext password authentication only when TLS is configured
 	// (required for AWS RDS IAM auth, safe because the connection uses TLS).
 	cfg.AllowCleartextPasswords = cfg.TLSConfig != ""
--- a/pkg/dbconn/replica_tls_preservation_test.go
+++ b/pkg/dbconn/replica_tls_preservation_test.go
@@ -322,11 +322,11 @@ func TestTLSParameterEdgeCases(t *testing.T) {
 		description string
 	}{
 		{
-			name:        "Mixed case TLS parameter not recognized by ParseDSN",
+			name:        "Mixed case TLS parameter is ignored by ParseDSN",
 			dsn:         "user:pass@tcp(host:3306)/db?TLS=false",
 			config:      &DBConfig{TLSMode: "REQUIRED"},
 			expected:    "user:pass@tcp(host:3306)/db?tls=required&TLS=false",
-			description: "Mixed case TLS parameter is not recognized by ParseDSN, so TLS config is added",
+			description: "Mixed case 'TLS' (uppercase) is not recognized by go-sql-driver's ParseDSN — it is ignored and passed through as an unknown query parameter, so our TLS config is added alongside it",
 		},
 		{
 			name:        "TLS in middle of query params",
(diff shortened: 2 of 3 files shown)


pkg/dbconn/conn.go
```diff
@@ -255,50 +245,51 @@ func newDSN(dsn string, config *DBConfig) (string, error) {
...
 	// This behaviour has been observed during blue/green upgrades and failover on AWS Aurora.
 	// See also: https://github.com/go-sql-driver/mysql?tab=readme-ov-file#rejectreadonly
-	ops = append(ops, fmt.Sprintf("%s=%s", "rejectReadOnly", "true"))
-	// Set interpolateParams
-	ops = append(ops, fmt.Sprintf("%s=%t", "interpolateParams", config.InterpolateParams))
+	cfg.RejectReadOnly = true
+	cfg.InterpolateParams = config.InterpolateParams
 	// Allow mysql_native_password authentication
-	ops = append(ops, fmt.Sprintf("%s=%s", "allowNativePasswords", "true"))
+	cfg.AllowNativePasswords = true
```
Kiran01bm: since `AllowNativePasswords = true` is the default in the MySQL driver, the explicit setting on line 287 is kind of redundant and can be removed.. - https://github.com/go-sql-driver/mysql/blob/v1.9.3/dsn.go#L96

pkg/dbconn/dsn_tls_test.go
```diff
@@ -216,8 +218,8 @@ func TestEnhanceDSNWithTLS_EdgeCases(t *testing.T) {
 				cfg.TLSMode = "REQUIRED"
 				return cfg
 			}(),
-			expectedResult: "user:pass@tcp(localhost:3306)/db?TLS=skip-verify",
-			description:    "Mixed case TLS parameter should be preserved",
+			expectedResult: "user:pass@tcp(localhost:3306)/db?tls=required&TLS=skip-verify",
+			description:    "Mixed case TLS parameter is not recognized by ParseDSN, so TLS config is added",
```
Kiran01bm: may be reword as to highlight that `TLS=skip-verify` is not honoured.

pkg/dbconn/fallback_test.go
Kiran01bm, line 1: im on board with dropping support for case-insensitive TLS config but the fallback behavior is still critical and functional - so it should have test coverage..

whats changed here is just the implementation detail to use `newDSN()` with TLSMode="DISABLED" instead of the old dedicated `createFallbackDSN()` which had case-insensitive support.

pkg/dbconn/replica_tls_preservation_test.go
```diff
@@ -322,11 +322,11 @@ func TestTLSParameterEdgeCases(t *testing.T) {
...
 	}{
 		{
-			name:        "Mixed case TLS parameter preserved",
+			name:        "Mixed case TLS parameter not recognized by ParseDSN",
 			dsn:         "user:pass@tcp(host:3306)/db?TLS=false",
 			config:      &DBConfig{TLSMode: "REQUIRED"},
-			expected:    "user:pass@tcp(host:3306)/db?TLS=false",
-			description: "Mixed case TLS parameter should be preserved",
+			expected:    "user:pass@tcp(host:3306)/db?tls=required&TLS=false",
+			description: "Mixed case TLS parameter is not recognized by ParseDSN, so TLS config is added",
```
Kiran01bm: we can improve the wording here to specifically call-out that the mixed-case args are ignored..

PR #596, opened 2026-02-22
enable specific linters

morgo:

This is a followup to https://github.com/block/spirit/pull/578

It enables some specific non-default linters which are high value, but still keeps the configuration simple.


Kiran01bm (approved):
love this!

PR #617, opened 2026-02-25
Add linter for rename column

morgo:

Part of https://github.com/block/spirit/issues/446

Example output:
```
mtocker@BLKD5234QXY40 bin % ./spirit diff --source-dsn="spirit:spirit@tcp(127.0.0.1:3306)/source"  --target-alter="alter table t1_old rename column val to val2"
-- [WARNING] rename_column: Column rename detected in table "t1_old": "val" to "val2". Renaming a column cannot be done atomically across application pods, and ORMs that generate column names at compile time (e.g. jOOQ) will break until code is recompiled (Table: t1_old, Column: val) Suggestion: Use ADD COLUMN + DROP COLUMN instead of RENAME COLUMN. This is the only safe approach

alter table t1_old rename column val to val2;
```


pkg/lint/lint_rename_column.go
```diff
@@ -0,0 +1,92 @@
...
+					newName = spec.NewColumnName.Name.O
+				}
+				violations = append(violations, Violation{
+					Linter: l,
+					Location: &Location{
+						Table:  change.Table,
+						Column: strPtr(oldName),
+					},
+					Message:    fmt.Sprintf("Column rename detected in table %q: %q to %q. Renaming a column cannot be done atomically across application pods, and ORMs that generate column names at compile time (e.g. jOOQ) will break until code is recompiled", change.Table, oldName, newName),
+					Severity:   SeverityWarning,
```
aparajon, line 66: Should this be a higher severity?

morgo: For now most things are warnings. We will have to revisit all the severity levels though. I'd like to have a 1 pager which explains them and why.

Issue #618, opened 2026-02-25
Handle SET/ENUM reorder better


morgo:
Some interesting problems:

1. Reordering a SET column will make the checksum fail. This is because values appear in the string representation in order. So "val1,val2" != "val2,val1" in a CRC32. Since the checksum is now compulsory, its better to fast-fail this in a pre-check. The workarounds to normalize the order are complex/anoying.

2. Re-ordering ENUM values is unsafe in buffered migrations (does not apply to moves, because move never transforms). This appears to be because of the replication protocol having the ordinal value of the ENUM, which changes. We should detect enum changes + buffered and refuse as a preflight check.

PR #619, opened 2026-02-25
Add set/enum handling

morgo:

Fixes https://github.com/block/spirit/issues/618

### DIFF
--- a/scripts/setup-git-hooks.sh
+++ b/scripts/setup-git-hooks.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# Setup script to configure Git hooks for the Spirit project
+# This enables automatic linting checks before pushes
+
+set -e
+
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+HOOKS_DIR="$REPO_ROOT/.githooks"
+GIT_HOOKS_DIR="$REPO_ROOT/.git/hooks"
+
+echo "Setting up Git hooks for Spirit..."
+
+# Check if Docker is available
+if ! command -v docker &> /dev/null; then
+    echo "⚠️  Warning: Docker is not installed or not in PATH"
+    echo "   Git hooks will be configured but won't work until Docker is available"
+    echo ""
+fi
+
+# Configure Git to use our custom hooks directory
+echo "Configuring Git to use .githooks directory..."
+git config core.hooksPath "$HOOKS_DIR"
+
+# Make all hook scripts executable
+echo "Making hook scripts executable..."
+chmod +x "$HOOKS_DIR"/*
+
+echo ""
+echo "✅ Git hooks setup complete!"
+echo ""
+echo "The following hooks are now active:"
+echo "  - pre-push: Runs golangci-lint in Docker before each push"
+echo ""
+echo "To bypass hooks for a specific push (not recommended):"
+echo "  git push --no-verify"
+echo ""
+echo "To disable hooks entirely:"
+echo "  git config --unset core.hooksPath"
+echo ""


pkg/migration/runner_test.go
```diff
@@ -2963,3 +2963,259 @@ func TestPasswordMasking(t *testing.T) {
...
+		for i := 0; i < 50; i++ {
+			if ctx.Err() != nil {
+				return
+			}
+			_, _ = dmlDB.ExecContext(ctx, `INSERT INTO setreorder (perms) VALUES ('read')`)
+			_, _ = dmlDB.ExecContext(ctx, `INSERT INTO setreorder (perms) VALUES ('write,execute')`)
+			_, _ = dmlDB.ExecContext(ctx, `INSERT INTO setreorder (perms) VALUES ('read,write,execute')`)
+			_, _ = dmlDB.ExecContext(ctx, fmt.Sprintf(`UPDATE setreorder SET perms = 'read,write' WHERE id = %d`, i*3+1))
+			_, _ = dmlDB.ExecContext(ctx, fmt.Sprintf(`UPDATE setreorder SET perms = 'execute' WHERE id = %d`, i*3+2))
+			_, _ = dmlDB.ExecContext(ctx, fmt.Sprintf(`UPDATE setreorder SET perms = 'read,write,execute' WHERE id = %d`, i*3+3))
```
aparajon, line 3205: Perhaps we can explain the indexes here for future reference?

pkg/migration/runner_test.go
```diff
@@ -2963,3 +2963,259 @@ func TestPasswordMasking(t *testing.T) {
...
+			if ctx.Err() != nil {
+				return
+			}
+			_, _ = dmlDB.ExecContext(ctx, `INSERT INTO enumreorder (status) VALUES ('active')`)
+			_, _ = dmlDB.ExecContext(ctx, `INSERT INTO enumreorder (status) VALUES ('inactive')`)
+			_, _ = dmlDB.ExecContext(ctx, `INSERT INTO enumreorder (status) VALUES ('pending')`)
+			// Also update some existing rows to exercise the update path
+			_, _ = dmlDB.ExecContext(ctx, fmt.Sprintf(`UPDATE enumreorder SET status = 'active' WHERE id = %d`, i*3+1))
+			_, _ = dmlDB.ExecContext(ctx, fmt.Sprintf(`UPDATE enumreorder SET status = 'inactive' WHERE id = %d`, i*3+2))
+			_, _ = dmlDB.ExecContext(ctx, fmt.Sprintf(`UPDATE enumreorder SET status = 'pending' WHERE id = %d`, i*3+3))
```
aparajon, line 3079: Similar question regarding explaining the meaning of these indexes

pkg/migration/runner_test.go
```diff
@@ -2963,3 +2963,259 @@ func TestPasswordMasking(t *testing.T) {
...
+			time.Sleep(time.Millisecond)
+			if ctx.Err() != nil {
+				return
+			}
+		}
+		// Insert rows with each ENUM value during the copy phase.
+		// These will be picked up by the binlog and replayed.
+		// We ignore errors because Spirit may kill our connection during
+		// the checksum/lock phase (force-kill of blocking transactions).
+		for i := 0; i < 50; i++ {
```
aparajon, line 3069: Any significance here to the value of `50`?

aparajon (approved):

pkg/migration/runner_test.go
morgo: No significance.

The part that breaks is the replication stream, because it only retains the ordinal enum position. Because of the keyAboveWatermark optimization we need to insert/update the rows at the start of the table. Picking 50 is a pretty safe value to avoid a race:

Because we can't easily know where the copier low watermark is at, we just spam single row updates enough times. We pick different rows to prevent triggering another optimization, which is that updates to the same key are merged.

pkg/migration/runner_test.go
morgo: Not much significance other than the row IDs are safely below the low watermark, and they are distinct.

PR #620, opened 2026-02-25
feat(lint): add IndexColumnExistsLinter to validate index columns exist

aparajon:
## Summary

Adds a new linter that validates index columns actually exist in the table. This catches errors like `INDEX idx_foo (nonexistent_column)` that would fail at execution time with:

```
Key column 'nonexistent_column' doesn't exist in table
```

## What it checks

- **CREATE TABLE statements**: validates all index columns exist in the table definition
- **ALTER TABLE ADD INDEX**: validates columns exist in the existing table
- **Case-insensitive matching**: handles MySQL's default case-insensitive column names
- **ADD COLUMN + ADD INDEX**: accounts for columns being added in the same ALTER statement

## Severity

Returns `SeverityError` since these statements will always fail at execution time.

## Example

```sql
-- Existing table
CREATE TABLE users (
    id BIGINT PRIMARY KEY,
    full_name VARCHAR(200)
);

-- This ALTER has a typo (full_name1 instead of full_name)
ALTER TABLE users ADD INDEX full_name1 (full_name1);
```

Linter output:
```
[ERROR] Index 'full_name1' references column 'full_name1' which does not exist in table 'users'
Suggestion: Column 'full_name1' does not exist in table 'users'. Available columns: id, full_name
```

### DIFF
--- a/pkg/lint/lint_index_column_exists.go
+++ b/pkg/lint/lint_index_column_exists.go
@@ -1,0 +1,136 @@
+package lint
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/block/spirit/pkg/statement"
+	"github.com/pingcap/tidb/pkg/parser/ast"
+)
+
+func init() {
+	Register(&IndexColumnExistsLinter{})
+}
+
+// IndexColumnExistsLinter validates that index columns actually exist in the table.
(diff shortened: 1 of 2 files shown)


aparajon:
Closing as per https://github.com/block/spirit/pull/621

Issue #587, opened 2026-01-30
Flaky test:TestResumeFromCheckpointE2E


morgo:
```
2026-01-30T22:22:11.1003769Z test-1           | --- FAIL: TestResumeFromCheckpointE2E (320.16s)
2026-01-30T22:22:11.1003861Z test-1           |     move_test.go:179: 
2026-01-30T22:22:11.1004002Z test-1           |         	Error Trace:	/app/pkg/move/move_test.go:179
2026-01-30T22:22:11.1004128Z test-1           |         	            				/app/pkg/move/move_test.go:68
2026-01-30T22:22:11.1004255Z test-1           |         	Error:      	Received unexpected error:
2026-01-30T22:22:11.1004856Z test-1           |         	            	checksum failed after 3 attempts. This likely indicates either a bug in Spirit, or a manual modification to the _new table outside of Spirit. Please report @ github.com/block/spirit

```

I have also seen this test timeout and not fail. It's always just this test.

PR #633, opened 2026-02-26
fix race

morgo:

Fixes https://github.com/block/spirit/issues/587

I believe this finally fixes this issue. The problem appears to be the racy background routine starting.

I modified the test to step through copying, and it's OK if it finishes. I then dump a checkpoint, which should pass at this point.

I did finally manage to reproduce the issue locally by running the tests in a loop. After this fix, I couldn't reproduce locally anymore.


Kiran01bm (approved):
lgtm

PR #635, opened 2026-02-26
mark linters stable

morgo:

Fixes https://github.com/block/spirit/issues/446

This renames the lint options so they are no longer experimental. Importantly, it also removes linter configuration (for now) which I think could be controversial. Let me explain that decision:

- We have a philosophy of [decisions not options](https://github.com/block/spirit/blob/main/.github/CONTRIBUTING.md).
- We know that the levels for linters is not perfect right now. If we allow configuration, we don't get the feedback loop of what to change. I'd prefer to force the conversation.
- For advanced usage, users can always use the API instead
- I am fine to revisit this later

### DIFF
--- a/pkg/migration/resume-from-checkpoint.md
+++ b/pkg/migration/resume-from-checkpoint.md
@@ -1,0 +1,73 @@
+# Resume from Checkpoint
+
+Spirit automatically checkpoints progress during a migration, allowing it to resume from where it left off if the process is killed or restarted. This is useful for long-running migrations on large tables, where restarting from scratch would be expensive.
+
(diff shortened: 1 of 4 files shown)


docs/migrate.md
```diff
@@ -114,6 +116,20 @@ Note that the checksum, if enabled, will be computed after the sentinel table is
...
 The host (and optional port) to use when connecting to MySQL. If no port is provided, 3306 is used.
 
+### lint
+
+- Type: Boolean
+- Default value: `false`
+
+Spirit can optionally run lint checks before executing a migration. This uses the same linting engine as [`spirit lint`](lint.md) and [`spirit diff`](diff.md), but runs inline as part of the migration process.
+
+### lint-only
```
aparajon, line 126: Is this option any different than using `spirit lint` by itself?

aparajon (approved):

docs/migrate.md
morgo: Not materially. The use case is I can see someone wanting to use the same command but first pre-checking it with lint-only for errors.

Issue #638, opened 2026-02-26
add mysql 8.0.42 to testing matrix


morgo:
This is the second Aurora LTS version.

We keep 8.0.28 in CI even though its ancient because its the first LTS version.

PR #639, opened 2026-02-26
bump CI versions to include latest Aurora LTS + mysql community

morgo:

Fixes https://github.com/block/spirit/issues/638

Also bumps the non-Aurora versions to the last 8.0 version.

### DIFF
--- a/docs/diff.md
+++ b/docs/diff.md
@@ -87,4 +87,4 @@ spirit diff --source-dsn "..." --target-dir ./schema-v2/ | mysql -u root mydb
 ## See Also
 
 - [`spirit lint`](lint.md) — lint an entire schema without diffing
-- [Inline linting](migrate.md#inline-linting) — run lint checks as part of `spirit migrate`
+- [`spirit migrate --lint`](migrate.md#lint) — run lint checks inline as part of `spirit migrate`
--- a/docs/lint.md
+++ b/docs/lint.md
@@ -108,4 +108,4 @@ Each violation has one of three severity levels:
 ## See Also
 
 - [`spirit diff`](diff.md) — compare two schemas and lint the changes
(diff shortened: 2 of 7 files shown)


.github/workflows/mysql8.0.42-docker.yml
```diff
@@ -0,0 +1,16 @@
+name: MySQL 8.0.42 (Aurora LTS version) /w docker-compose
```
aparajon, line 1: Should this be `8.0.45`?

compose/replication-tls/8.0.42.yml
```diff
@@ -0,0 +1,7 @@
+services:
+  mysql:
+    image: mysql:8.0.42
+    ports: [ '8042:3306' ]
+  mysql_replica:
+    image: mysql:8.0.42
```
aparajon, line 6: Should these be `8.0.45`?

.github/workflows/mysql8.0.42-docker.yml
morgo: No, this is from: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraMySQLReleaseNotes/AuroraMySQL.Updates.3100.html

Aurora designates a specific version as LTS. There are two MySQL versions this applies to:
8.0.22
8.0.42

So we test these specifically (we have systems on both of these LTSes).

For MySQL community releases we test the latest: 8.0.45.

compose/replication-tls/8.0.42.yml
morgo: No I think this is correct. It is tested from `.github/workflows/mysql8.0.42-docker.yml‎`, which is the Aurora version test.

PR #642, opened 2026-02-26
fix: validate binlog position before creating replClient on resume

aparajon:
## Summary

When `resumeFromCheckpoint` runs after a stop/start cycle (e.g., volume change), the checkpoint's binlog position may reference a purged file. Previously, this was detected late — after creating the replClient and adding subscriptions — causing `newMigration` to fail with "subscription already exists for table X".

Now we validate the binlog file exists (via `SHOW BINARY LOGS`) early in `resumeFromCheckpoint`, before creating any resources. If the file is gone, we return immediately and let `setup()` fall back to `newMigration` with no partial state to clean up.

### DIFF
--- a/pkg/migration/resume-from-checkpoint.md
+++ b/pkg/migration/resume-from-checkpoint.md
@@ -34,22 +34,40 @@ When a new Runner starts (`Runner.Run()` → `setup()`), it always attempts `res
 
 If any step fails (and strict mode is not enabled), Spirit logs the reason and falls back to `newMigration()`, which starts the migration from scratch. This means resume is best-effort — Spirit will always make forward progress even if the checkpoint is unusable.
 
+## Background: how MySQL binary logs work
+
+MySQL binary logs (binlogs) are a sequence of files that record every data-changing operation on the server. MySQL creates a series of numbered files — `mysql-bin.000001`, `mysql-bin.000002`, etc. — rotating to a new file when the current one reaches `max_binlog_size` (default 1GB) or the server restarts. Every INSERT, UPDATE, DELETE, and DDL statement is recorded.
+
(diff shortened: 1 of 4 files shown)


pkg/migration/runner.go
```diff
@@ -698,6 +698,15 @@ func (r *Runner) setup(ctx context.Context) error {
...
 
+		// Clean up any partially initialized state from resumeFromCheckpoint.
+		// This is important because resumeFromCheckpoint may have created a replClient
+		// and added subscriptions before failing (e.g., at replClient.Run()).
+		// Without cleanup, newMigration would fail with "subscription already exists".
+		if r.replClient != nil {
+			r.replClient.Close()
+			r.replClient = nil
+		}
+
```
morgo: Currently we optimistically try to resumeFromCheckpoint and then catch any error and then run newMigration.

This is what's causing the problem; it got a lot further into resuming from checkpoint than we expected.

We could cleanup the replClient like this, but an alternative is we do an errors.Is() on the returned error and be a bit more selective about what is resumable.

I'm not sure, do you have opinions?

(The downside of the current fix is it might be a yak shaving if we have to start cleaning up other resources here later.)

pkg/migration/resume-from-checkpoint.md
```diff
@@ -0,0 +1,97 @@
+# Resume from Checkpoint
+
+Spirit supports resuming migrations from a checkpoint. This document explains the checkpoint mechanism, potential failure modes, and how they're handled.
+
```
morgo, line 4: For the technical docs, I've been putting them in the packages. So `pkg/migration/` rather than `docs/`

This separates the audience - it's a lot harder to understand this

morgo (approved):

PR #663, opened 2026-03-17
Enhance load schema with filtering

morgo:

This adds optional filtering to the LoadSchemaFromDB API. We are always filtering out the same things in our downstream API usage, so it makes sense to push it into here.

### DIFF
--- a/pkg/table/table_schema_test.go
+++ b/pkg/table/table_schema_test.go
@@ -78,18 +78,6 @@ func TestLoadSchemaFromDB_PreservesAutoIncrement(t *testing.T) {
 	assert.Contains(t, tables[0].Schema, "AUTO_INCREMENT=")
 }
 
-func TestLoadSchemaFromDB_MultipleOptsReturnsError(t *testing.T) {
-	dbName := testutils.CreateUniqueTestDatabase(t)
-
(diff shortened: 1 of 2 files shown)


pkg/table/table_schema_test.go
```diff
@@ -77,3 +77,103 @@ func TestLoadSchemaFromDB_PreservesAutoIncrement(t *testing.T) {
...
 }
+
+func TestLoadSchemaFromDB_MultipleOptsReturnsError(t *testing.T) {
+	dbName := testutils.CreateUniqueTestDatabase(t)
+
+	db, err := sql.Open("mysql", testutils.DSNForDatabase(dbName))
+	require.NoError(t, err)
+	defer func() { _ = db.Close() }()
+
+	_, err = LoadSchemaFromDB(t.Context(), db, FilterOptions{}, FilterOptions{})
```
aparajon: Should `FilterOptions` be renamed to `FilterOption` since we are using the varargs form?

morgo: I think this is a good point, it's not quite go idiomatic. It takes a struct with multiple options, whereas normally in golang you could chain multiple filters. I think maybe I'll rewrite it.

morgo: OK PTAL again

PR #667, opened 2026-03-18
Add PlanChanges and DeclarativeToImperative to API

morgo:

This adds `lint.PlanChanges()` and `statement.DeclarativeToImperative()`. They are related:
- PlanChanges() calls DeclarativeToImperative() first and then adds in lint suggestions
- You can call either, but I think most of our usages will call PlanChanges() to get the lint suggestions.

### DIFF
--- a/pkg/statement/declarative.go
+++ b/pkg/statement/declarative.go
@@ -18,7 +18,10 @@ import (
 // subcommand, strata, and GAP.
 //
 // The returned statements are ordered by table name, with CREATE/ALTER
-// statements before DROP statements.
+// statements before DROP statements. This ordering is a correctness property:
+// it ensures the output is safe to execute sequentially (e.g. a table
+// referenced by a foreign key will not be dropped before the referencing
+// ALTER runs).
 //
 // If opts is nil, NewDiffOptions() defaults are used for table diffs.
 func DeclarativeToImperative(current, desired []table.TableSchema, opts *DiffOptions) ([]*AbstractStatement, error) {
@@ -56,17 +59,9 @@ func DeclarativeToImperative(current, desired []table.TableSchema, opts *DiffOpt
 		}
 
 		// Both exist — compute ALTER TABLE diff.
-		a, err := ParseCreateTable(existingTable.Schema)
+		diffs, err := diffTable(name, existingTable.Schema, desiredTable.Schema, opts)
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse current schema for table %q: %w", name, err)
-		}
-		b, err := ParseCreateTable(desiredTable.Schema)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse desired schema for table %q: %w", name, err)
-		}
-		diffs, err := a.Diff(b, opts)
-		if err != nil {
-			return nil, fmt.Errorf("failed to diff table %q: %w", name, err)
+			return nil, err
 		}
 		changes = append(changes, diffs...)
 	}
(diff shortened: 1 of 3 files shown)


pkg/statement/declarative.go
```diff
@@ -0,0 +1,109 @@
...
+		diffs, err := a.Diff(b, opts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to diff table %q: %w", name, err)
+		}
+		changes = append(changes, diffs...)
+	}
+
+	// Tables in current but not in desired — emit DROP TABLE.
+	dropNames := make([]string, 0)
+	for name := range currentMap {
```
aparajon, line 71: Spirit's `CreateTable.Diff()` can panic on certain edge cases (e.g., identical schemas with formatting differences between MySQL's `SHOW CREATE TABLE` output and embedded schema files). Downstream consumers who currently wrap `Diff()` in `recover()` would lose that safety when switching to this API.

Worth considering adding panic recovery inside `DeclarativeToImperative` (or documenting that callers must handle panics themselves). Otherwise the clean single-call API becomes less safe than doing the diff manually.

*— Claude Code*

pkg/lint/declarative.go
```diff
@@ -0,0 +1,198 @@
...
+//
+// LintOnlyChanges is always set to true regardless of the provided lintConfig,
+// since PlanChanges only produces lint results for tables that have changes.
+//
+// Parameters:
+//   - current: the current table schemas (CREATE TABLE DDL)
+//   - desired: the desired table schemas (CREATE TABLE DDL)
+//   - diffOpts: options for the diff (nil uses defaults)
+//   - lintConfig: configuration for linting (nil uses defaults; LintOnlyChanges is always overridden to true)
+func PlanChanges(current, desired []table.TableSchema, diffOpts *statement.DiffOptions, lintConfig *Config) (*Plan, error) {
```
aparajon, line 99: Should `PlanChanges` accept `[]table.TableSchema` for current schemas, or would it also make sense to accept `[]*statement.CreateTable` directly? Downstream consumers that already have parsed `CreateTable` objects (e.g., from `table.LoadSchemaFromDB`) would avoid a round-trip through `ToTableSchema()` → `ParseCreateTable()` here.

Not blocking — the `TableSchema` input is cleaner as a public API contract. Just wondering if a convenience overload would help.

*— Claude Code*

pkg/lint/declarative.go
```diff
@@ -0,0 +1,198 @@
...
+	"strings"
+
+	"github.com/block/spirit/pkg/statement"
+	"github.com/block/spirit/pkg/table"
+)
+
+// PlannedChange represents a single DDL statement produced by declarative
+// schema diffing, along with any lint violations associated with it.
+type PlannedChange struct {
+	// Statement is the semicolon-terminated DDL (ALTER, CREATE, or DROP TABLE).
```
aparajon, line 14: The `PlannedChange` struct is exactly the shape downstream consumers need for rendering diff results (e.g., in CLI output or PR comments) — statement + table name + categorized violations in one place. Much better than having to correlate `RunLinters` violations back to statements manually. 👍

*— Claude Code*

pkg/statement/declarative.go
```diff
@@ -0,0 +1,109 @@
...
+	var sb strings.Builder
+	rCtx := format.NewRestoreCtx(format.DefaultRestoreFlags, &sb)
+	if err := ct.Raw.Restore(rCtx); err != nil {
+		return table.TableSchema{}, fmt.Errorf("failed to restore CREATE TABLE for %q: %w", ct.TableName, err)
+	}
+	return table.TableSchema{
+		Name:   ct.TableName,
+		Schema: sb.String(),
+	}, nil
+}
```
aparajon, line 133: The CREATE/ALTER-before-DROP ordering guarantee is important for consumers — it means the output is safe to execute sequentially (you won't drop a table that another table's FK references before the referencing ALTER runs). Worth calling this out more prominently in the doc comment, since it's a correctness property, not just a cosmetic choice.

*— Claude Code*

pkg/lint/declarative.go
morgo: No change right now, reasonable suggestion, but the other callers prefer `[]table.TableSchema`.

pkg/statement/declarative.go
morgo: It handles panics now.

Issue #690, opened 2026-04-11
Support N:M move (reshard) operations


morgo:
Currently reshards are limited to 1:n. We should support N:M.

morgo:
Followup from #691:

> also one more observation - i did not notice this earlier - there are no tests for len(Sources) > 1 i.e validating that preflight checks correctly iterate over multiple sources when N>1 - perhaps this can be improved..

morgo:
Re: the followup, I've added tests for this in #693. It was already handled by the configuration check, but it is now more consistently handled.

PR #691, opened 2026-04-11
check: refactor Resources to support multiple sources

morgo:

Part of https://github.com/block/spirit/issues/690

Replace individual SourceDB/SourceConfig/SourceDSN/TargetDSN fields with Sources []SourceResource slice in the Resources struct. This prepares the check package for N:M move operations where multiple source databases need to be validated.

All checks updated to iterate over Sources:
- configuration: validates MySQL config on all sources
- privileges: checks privileges on all sources via checkSourcePrivileges helper
- resume_state: uses Sources[0] for checkpoint validation
- target_state/table_compatibility: unchanged (use SourceTables only)

The runner wraps its single source into []SourceResource{...} for backward compatibility. No behavioral change for 1:1 moves.

### DIFF
--- a/pkg/move/check/configuration.go
+++ b/pkg/move/check/configuration.go
@@ -3,6 +3,7 @@ package check
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/move/check/configuration.go
```diff
@@ -3,67 +3,57 @@ package check
...
-		&binlogFormat,
-		&binlogRowImage,
-		&logBin,
-		&logSlaveUpdates,
-		&binlogRowValueOptions,
-		&performanceSchema,
-	)
-	if err != nil {
-		return err
+	for i, src := range r.Sources {
```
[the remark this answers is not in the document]
morgo: Fixed: moved the `len(r.Sources) == 0` check to the top of the function (it was dead code at the bottom since the loop wouldn't execute), and added an explicit `src.DB == nil` guard before dereferencing. See d12b68c.

pkg/move/check/configuration.go
```diff
@@ -3,67 +3,57 @@ package check
...
-	}
-
-	if logSlaveUpdates != "1" {
-		// This is a hard requirement unless we enhance this to confirm
-		// it's not receiving any updates via the replication stream.
-		return errors.New("log_slave_updates must be enabled")
+	// Preserve the original single-source error format for backward compatibility.
+	if len(r.Sources) == 0 {
+		return errors.New("no sources configured")
 	}
```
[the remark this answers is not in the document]
morgo: Fixed: removed the misleading backward-compatibility comment and moved the empty sources check to the top of the function where it actually guards against the empty case. The `source %d:` prefix is intentional — we don't need backward-compatible error strings.

pkg/move/check/resume_state.go
```diff
@@ -29,21 +29,26 @@ func resumeStateCheck(ctx context.Context, r Resources, logger *slog.Logger) err
...
+
+	// Check 1: Verify checkpoint table exists on sources[0] (by convention).
+	src0 := r.Sources[0]
 	checkpointTableName := "_spirit_checkpoint"
 	var checkpointExists int
-	err := r.SourceDB.QueryRowContext(ctx,
+	err := src0.DB.QueryRowContext(ctx,
 		"SELECT 1 FROM information_schema.TABLES WHERE table_schema = ? AND table_name = ?",
-		r.SourceConfig.DBName, checkpointTableName).Scan(&checkpointExists)
+		src0.Config.DBName, checkpointTableName).Scan(&checkpointExists)
```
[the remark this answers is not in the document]
morgo: Fixed: added explicit `src0.DB == nil || src0.Config == nil` guard after the `len(r.Sources) == 0` check, before any dereferencing.

pkg/move/runner.go
```diff
@@ -719,13 +719,14 @@ func (r *Runner) SetLogger(logger *slog.Logger) {
...
+		Sources: []check.SourceResource{{
+			DB:     r.source,
+			Config: r.sourceConfig,
+			DSN:    r.move.SourceDSN,
+		}},
 		Targets:        r.targets,
 		SourceTables:   r.sourceTables,
 		CreateSentinel: r.move.CreateSentinel,
-		SourceDSN:      r.move.SourceDSN,
-		TargetDSN:      r.move.TargetDSN,
```
Kiran01bm, line 728: removal of `TargetDSN` is this intentional ?

pkg/move/check/configuration.go
```diff
@@ -3,67 +3,59 @@ package check
...
-		&binlogRowImage,
-		&logBin,
-		&logSlaveUpdates,
-		&binlogRowValueOptions,
-		&performanceSchema,
-	)
-	if err != nil {
-		return err
+	if len(r.Sources) == 0 {
+		return errors.New("no sources configured")
```
Kiran01bm, line 17: seems like an inconsistency -  in privileges we state `// skip if no source DB connection yet (pre-run phase)` - but they both are configured with PreFlight scopes.. ?

https://github.com/block/spirit/pull/691/changes#diff-7a36da34caa55aabde6d9c8e661929c7cd8b9aa69bbc6c67305a2925f4f6bfdeR48

pkg/move/runner.go
morgo: yes, the removal is intentional (for now).

This is to do with check.Resources. We don't currently have any checks that check the target. If we add some, I can add it back, but it's better not to have unused code for now.

pkg/move/check/configuration.go
morgo: Good catch, will fix.

pkg/move/check/resume_state.go
```diff
@@ -29,21 +29,29 @@ func resumeStateCheck(ctx context.Context, r Resources, logger *slog.Logger) err
 		return errors.New("no source tables available for resume validation")
 	}
 
-	// Check 1: Verify checkpoint table exists on source
+	if len(r.Sources) == 0 {
+		return errors.New("no sources configured for resume validation")
+	}
+
+	// Check 1: Verify checkpoint table exists on sources[0] (by convention).
```
Kiran01bm, line 36: for the comment above - is this fine for N:M moves as well ?

morgo: Yes.  The checkpoint table only lives on one source, and it is always sources[0] (even if there are N sources).

I went with this design because otherwise we'd have to figure out how to reconcile that we can't easily atomically write to all sources. The only requirement is that the sources list must be in the same sorted order, otherwise it won't recover.

Kiran01bm (approved):

Kiran01bm:
also one more observation - i did not notice this earlier - there are no tests for `len(Sources) > 1` i.e validating that preflight checks correctly iterate over multiple sources when `N>1` - perhaps this can be improved..

PR #696, opened 2026-04-20
Fix flaky migration integration tests

aparajon:
## What's this?

Fixes two intermittently failing integration tests in `pkg/migration` that have been blocking CI on unrelated PRs.

## Changes

**`TestCutoverAtomicityWithConcurrentWrites`**
- Wrapped the existing checksum comparison in `assert.Eventually` (5s timeout, 250ms poll) to tolerate brief replication lag after the partial cutover
- The original single-shot comparison could catch a transient mismatch when residual replication events hadn't fully settled in CI Docker containers

**`TestResumeFromCheckpointE2EWithManualSentinel`**
- Reduced test data from ~400k rows (4x 100k) to ~100k rows (2x 50k)
- The resumed migration was timing out at 30s with only ~70% of rows copied on slow CI runners
- 100k rows is still plenty for the first migration to be mid-copy when checkpointed

## Verification

Both tests pass 5/5 locally against Docker MySQL with replication.

---
_Most of this was written by Claude Code — I just provided direction._

### DIFF
--- a/pkg/migration/cutover_test.go
+++ b/pkg/migration/cutover_test.go
@@ -320,25 +320,17 @@ func TestCutoverAtomicityWithConcurrentWrites(t *testing.T) {
 	assert.True(t, oldTableExists, "The _old table should exist after partial cutover")
 	assert.True(t, newTableExists, "The _new table should exist after partial cutover")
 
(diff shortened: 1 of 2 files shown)


morgo (approved):
Let's try it. Always hard to know with flaky tests.

PR #713, opened 2026-04-29
Improve table name length handling for CREATE TABLE and SkipDropAfterCutover

### LINKED ISSUE #712: Table name validation too restrictive with --skip-drop-after-cutover; CREATE TABLE bypasses validation
## Problem

Two issues with table name length enforcement in Spirit:

### 1. `--skip-drop-after-cutover` rejects table names > 43 characters at validation time

When `--skip-drop-after-cutover` is set, the `tableNameCheck` preflight check enforces a stricter limit of 43 characters (64 - 21 timestamp suffix chars). This **prevents migrations from ever starting** on tables with names between 44–56 characters long — even though the normal limit without this flag is 56 characters.

The error message is:
```
table name must be less than 43 characters when --skip-drop-after-cutover is set
```

This is overly restrictive. The correct approach is to truncate the table name portion in the renamed old table (similar to how [gap/tern handles this](https://github.com/squareup/gap/blob/28fa60727f0dc9414947d429d9fb997550dda769/apps/tern/engine/direct.go#L148-L190)), not to reject the migration entirely.

### 2. CREATE TABLE statements bypass all table name length validation

The CREATE TABLE path in `Runner.Run()` returns early before preflight checks run. This means a table can be created via Spirit with a name that Spirit cannot later migrate (because it needs room for metadata suffixes like `_<table>_chkpnt`, `_<table>_new`, etc.).

## Expected Behavior

1. Tables with names up to 56 characters should be migratable with `--skip-drop-after-cutover` — the old table name should be truncated to fit within MySQL's 64-character limit.
2. CREATE TABLE statements should enforce the same 56-character limit as ALTER TABLE, ensuring the created table can later be migrated by Spirit.

morgo:
Fixes #712

## Problem

1. **`--skip-drop-after-cutover` rejected tables with names 44–56 characters at validation time.** The preflight check enforced a 43-character limit when this flag was set, preventing migrations from ever starting — even though the actual fix is to truncate the table name in the rename operation.

2. **CREATE TABLE statements bypassed all table name length validation.** The CREATE TABLE path returns early before preflight checks run, so a table could be created with a name that Spirit cannot later migrate.

## Solution

- **Truncate instead of reject**: When `SkipDropAfterCutover` is true, `oldTableName()` now truncates the table name portion (using `utils.TruncateTableName`) so the timestamped old table name always fits within 64 characters. The preflight check no longer applies a stricter limit for this flag.

- **New `createTableNameCheck`**: A registered check at `ScopeStatement` validates CREATE TABLE names against Spirit's 56-character manageable limit (64 minus the longest metadata suffix). This runs via `runChecks(ScopeStatement)` before the non-ALTER early return in `Runner.Run()`.

- **New `ScopeStatement` check scope**: For checks that only require the parsed statement (no DB connection or table info needed).

- **`MaxTableNameLength` and `TruncateTableName` in `pkg/utils`**: Shared utilities to avoid circular imports between packages.

- **Simplified constants**: Replaced `init()`-calculated vars with plain constants (`MaxMigratableTableNameLength = 56`, `NameFormatTimestampExtraChars = 21`) for clarity.

### DIFF
--- a/pkg/migration/check/tablename.go
+++ b/pkg/migration/check/tablename.go
@@ -5,47 +5,31 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"strings"
 
 	"github.com/block/spirit/pkg/utils"
 )
 
 const (
-	// Formats for table names
+	// Formats for table names used by Spirit during migrations.
 	NameFormatCheckpoint   = "_%s_chkpnt"
 	NameFormatNew          = "_%s_new"
 	NameFormatOld          = "_%s_old"
 	NameFormatOldTimeStamp = "_%s_old_%s"
 	NameFormatTimestamp    = "20060102_150405"
-)
-
-var (
-	// The number of extra characters needed for table names with all possible
-	// formats. These vars are calculated in the `init` function below.
-	NameFormatNormalExtraChars    = 0
-	NameFormatTimestampExtraChars = 0
 
 	// MaxMigratableTableNameLength is the maximum table name length that Spirit
-	// can manage, accounting for the longest metadata suffix (e.g. _<table>_chkpnt).
-	// Calculated in init().
-	MaxMigratableTableNameLength = 0
+	// can manage. This is MySQL's 64-character limit minus the longest metadata
+	// suffix (_<table>_chkpnt = 8 extra characters), giving 56 usable characters.
+	MaxMigratableTableNameLength = utils.MaxTableNameLength - 8 // len("_") + len("_chkpnt") = 8
+
+	// NameFormatTimestampExtraChars is the number of extra characters added by
+	// the timestamped old table name format (_%s_old_%s with a 15-char timestamp).
+	// This equals len("_") + len("_old_") + len("20060102_150405") = 21.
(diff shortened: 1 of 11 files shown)


pkg/migration/check/tablename.go
```diff
@@ -25,6 +24,11 @@ var (
 	// formats. These vars are calculated in the `init` function below.
 	NameFormatNormalExtraChars    = 0
 	NameFormatTimestampExtraChars = 0
+
+	// MaxMigratableTableNameLength is the maximum table name length that Spirit
+	// can manage, accounting for the longest metadata suffix (e.g. _<table>_chkpnt).
+	// Calculated in init().
+	MaxMigratableTableNameLength = 0
```
aparajon: Should this be higher?

morgo: It is calculated in init(). That was the design that Claude came up with, I don't really like it though.. and I think we can try and simplify it to prevent misunderstanding.

aparajon (approved):

PR #737, opened 2026-04-30
status: write checkpoint immediately when WatchTask starts

### LINKED ISSUE #298: Immediately write checkpoint when copy begins
If a migration is interrupted before Spirit gets a chance to write to the `_chkpnt` table, progress is lost as Spirit is unable to pick up from where it left off and the following message is logged

```
INFO[0000] could not resume from checkpoint: reason=could not read from table '_sbtest1_chkpnt'
```

Ideally, Spirit would write to `_chkpnt` ASAP to avoid losing progress. Alternatively, an option to configure the interval at which checkpoints are taken may be sufficient.

morgo:
## Summary

Writes the first checkpoint immediately when the periodic checkpoint routine starts, rather than waiting for the first 50-second ticker interval.

## Motivation

Previously, if a migration crashed in the first 50 seconds of copying, no checkpoint existed and all progress was lost. This is particularly painful for large migrations where even a few seconds of progress represents meaningful work.

## Changes

- **`pkg/status/task.go`**: Added an immediate `DumpCheckpoint()` call at the start of `continuallyDumpCheckpoint()`, before entering the ticker loop. The immediate attempt gracefully handles `ErrWatermarkNotReady` (falls through to the periodic loop) since the watermark might not be ready at the very start.

- **`pkg/status/task_test.go`** (new file): Added tests verifying:
  - `DumpCheckpoint` is called immediately without waiting for the ticker
  - The routine continues normally when the immediate attempt returns `ErrWatermarkNotReady`

Fixes #298

### DIFF
--- a/pkg/status/task.go
+++ b/pkg/status/task.go
@@ -45,6 +45,27 @@ func continuallyDumpStatus(ctx context.Context, task Task, logger *slog.Logger) 
 }
 
 func continuallyDumpCheckpoint(ctx context.Context, task Task, logger *slog.Logger) {
+	// Attempt an immediate checkpoint write so that even short-lived
+	// migrations have a recovery point. If the watermark isn't ready
+	// yet (e.g. copier hasn't started), we just fall through to the
+	// periodic loop which will retry.
+	if err := task.DumpCheckpoint(ctx); err != nil {
+		switch {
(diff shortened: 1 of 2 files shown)


morgo:
I thought about this. I'm going to close it out; the patch is a bit larger than expected and we don't require this behavior internally.

PR #740, opened 2026-04-30
docs: recommend against --strict in migrate documentation

### LINKED ISSUE #736: Update docs to recommend against --strict
We added `--strict` for an internal user so spirit would retrofit to some of their previous usage.

On retrospect, it's not a great design, and we should instead recommend relying on idempotency, which is the default behavior. I am not suggesting we remove this option (yet), but we should discourage it because its usually the wrong choice.

morgo:
The default idempotent restart behavior is safer and more convenient for most users. `--strict` was added for a specific internal use case and can cause unexpected failures when checkpoints become stale.

This updates the `docs/migrate.md` to:
- Add a prominent warning recommending against `--strict`
- Explain the specific failure scenarios where strict mode causes errors
- Clarify that the default behavior (clean up and restart) is usually correct

Fixes #736

### DIFF
--- a/docs/migrate.md
+++ b/docs/migrate.md
@@ -274,10 +274,10 @@ There are some restrictions to `--statement`:
 
 By default, Spirit will automatically clean up old checkpoints before starting the schema change. This allows schema changes to always proceed forward, at the cost of potentially lost progress from a previous incomplete run.
 
-When set to `true`, if Spirit encounters a checkpoint belonging to a previous migration, it will validate that the alter statement matches the `--alter` parameter. If the validation fails (e.g., the ALTER was changed between runs, or the binlog position is no longer available), Spirit will exit with an error rather than silently restarting from scratch.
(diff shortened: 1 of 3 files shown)


Kiran01bm (approved):

Kiran01bm:
Other places that needs updating ?

1. **`pkg/migration/resume-from-checkpoint.md`** -- recommends `--strict` with caveats:
   - Line 59: *"Use `--strict` mode if losing progress silently is unacceptable."*
   - Lines 62–71 ("Strict mode" section): describes `--strict` neutrally as a feature to enable, with no caveat that it is not recommended for general use.

2. **`pkg/migration/migration.go` (line 44)** — Kong CLI help string for the flag:
   ```go
   Strict bool `name:"strict" help:"Exit on --alter mismatch when incomplete migration is detected" optional:"" default:"false"`
The help text does not mention that the flag is discouraged. Since spirit migrate --help is most users' first exposure to the flag, it should reflect the same "not recommended" framing as docs/migrate.md.

PR #777, opened 2026-05-01
fix(#771): validate connection with Ping in testutils.RunSQL

### LINKED ISSUE #771: flaky test: TestAllChangesFlushed (invalid connection)
```
2026-05-01T17:14:01.7840794Z test-1           | --- FAIL: TestAllChangesFlushed (10.36s)
2026-05-01T17:14:01.7840903Z test-1           |     subscription_test.go:19: 
2026-05-01T17:14:01.7841121Z test-1           |         	Error Trace:	/app/pkg/testutils/testing.go:121
2026-05-01T17:14:01.7841262Z test-1           |         	            				/app/pkg/repl/subscription_test.go:19
2026-05-01T17:14:01.7841391Z test-1           |         	            				/app/pkg/repl/client_test.go:661
2026-05-01T17:14:01.7841524Z test-1           |         	Error:      	Received unexpected error:
2026-05-01T17:14:01.7841634Z test-1           |         	            	invalid connection
2026-05-01T17:14:01.7841804Z test-1           |         	Test:       	TestAllChangesFlushed

```

morgo:
## Summary
- Fixes https://github.com/block/spirit/issues/771 (the RunSQL surface)
- `testutils.RunSQL` and `RunSQLInDatabase` open a fresh DB pool, run a single statement, and close the pool. Under heavy parallel test load (many `t.Parallel` tests each spinning up their own pool) we have observed transient \"invalid connection\" failures from the mysql driver on the very first query against a fresh pool — likely the server briefly dropping the connection during the handshake, or a TCP RST.
- Front the Exec with a `PingContext` that we retry on `driver.ErrBadConn` / `mysql.ErrInvalidConn` (up to 4 attempts, exponential backoff with jitter). Ping has no side effects, so retrying it is always safe. Once Ping succeeds the pool has a verified live connection, and the follow-up Exec is highly unlikely to hit the same transient handshake glitch.
- We deliberately do NOT retry the Exec itself: the mysql driver returns `ErrInvalidConn` for failures both before and after the query was sent, so a blind Exec retry could double-apply non-idempotent statements (INSERT, CREATE TABLE, etc.).

## Caveats
- This only addresses the RunSQL surface of #771 (e.g. `setupTestTables → DROP TABLE IF EXISTS` failing on first connection). The other pattern in the issue (an `m.Run()` returning \"invalid connection\" mid-migration, e.g. `TestPartitionedTable`) is a separate, deeper failure mode.

## Test plan
- [x] `go test -run TestAllChangesFlushed -count=3 ./pkg/repl/`
- [x] `go build ./... && go vet ./...`

### DIFF
--- a/pkg/testutils/testing.go
+++ b/pkg/testutils/testing.go
@@ -5,7 +5,10 @@ package testutils
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
+	"errors"
 	"fmt"
+	"math/rand"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


morgo:
Switching approach — Ping/Exec adds TOCTOU complexity for a problem that's better addressed by reducing test concurrency to ease pressure on the MySQL server. New PR coming.

Issue #701, opened 2026-04-24
Support for RENAME operations in limited contexts


morgo:
This is more of a community feature request so spirit can be used in more cases, since we tend not to do rename operations due to lack of safety.

But so spirit can be used in more contexts (and not have to use multiple OSC tools) we should consider supporting rename operations.

PR #829, opened 2026-05-08
fix(#828): backtick-quote PK column names in chunker and checksum SQL

### LINKED ISSUE #828: SQL syntax error when primary key contains reserved-word columns (e.g. `key`, `value`)
## Summary

Migrating a table whose primary key includes a column whose name is a MySQL
reserved word fails with a SQL syntax error during the first chunker
prefetch query. This affects any composite-PK table where one of the key
columns is `key`, `value`, `order`, etc.

## Repro

```sql
CREATE TABLE osm_points_of_interest (
  osm_id BIGINT NOT NULL,
  `key`   VARCHAR(64) NOT NULL,
  `value` TEXT,
  PRIMARY KEY (osm_id, `key`)
) ENGINE=InnoDB;
```

```
$ go run ./cmd/spirit migrate --alter="engine=innodb" --table=osm_points_of_interest
...
spirit: error: Error 1064 (42000): You have an error in your SQL syntax;
  check the manual that corresponds to your MySQL server version for the
  right syntax to use near 'key,value FROM `osm_points_of_interest`
  FORCE INDEX (PRIMARY)  ORDER BY osm_id,k' at line 1
```

## Root cause

`pkg/table/chunker_composite.go::Next()` builds the prefetch query with
`strings.Join(t.chunkKeys, ",")` for the SELECT list and ORDER BY clause
without backtick-quoting the column names:

https://github.com/block/spirit/blob/main/pkg/table/chunker_composite.go#L89-L107

```go
query := fmt.Sprintf("SELECT %s FROM %s FORCE INDEX (%s) %s ORDER BY %s LIMIT 1 OFFSET %d",
    strings.Join(t.chunkKeys, ","),  // <-- not quoted
    t.Ti.QuotedTableName,
    t.keyName,
    t.additionalConditionsSQL(false),
    strings.Join(t.chunkKeys, ","),  // <-- not quoted
    t.chunkSize,
)
```

For a PK of `(osm_id, key, value)` this produces

```
SELECT osm_id,key,value FROM `osm_points_of_interest` FORCE INDEX (PRIMARY)
  ORDER BY osm_id,key,value LIMIT 1 OFFSET N
```

…which fails because `key` is a reserved word.

The same bare-column issue exists in:

- `pkg/table/chunker_optimistic.go::nextChunkByPrefetching` (single-column,
  but same pattern — only triggers if the auto-inc PK column has a reserved
  name, much rarer in practice).
- `pkg/checksum/single.go::inspectDifferences` — the `CONCAT_WS` PK
  expression and the implicit ORDER BY through the `queryTemplate`'s
  unquoted column list.

The `WHERE` clause is built via `expandRowConstructorComparison`, which
*does* quote correctly, so chunkers happen to work for non-reserved column
names because the SELECT/ORDER-BY parts of the query are syntactically
valid even unquoted.

## Not a regression from #701

The unquoted `strings.Join` was introduced in commit `36877bb` (2023-08-02)
when composite-key chunking was first added. PR #702 / issue #701 only
touched the rename-check and column-mapping paths and does not affect this
codepath.

## Fix

Use the existing `table.QuoteColumns` helper (or an inline
`` strings.Join(quoted, ",") ``) at every call site. A single regression
test that builds a table with a reserved-word key column and walks the
chunker is enough — there are no existing tests for reserved-word PK
columns, which is why this slipped through.

morgo:
Fixes [#828](https://github.com/block/spirit/issues/828).

## Summary

- Migrating or moving a table whose primary key contains MySQL reserved words (e.g. `key`, `value`) failed immediately with Error 1064 because several SQL builders interpolated bare PK column names.
- Affected sites:
  - `pkg/table/chunker_composite.go::Next()` — composite prefetch query (the user-reported failure path).
  - `pkg/table/chunker_optimistic.go::nextChunkByPrefetching` — single-key prefetch query.
  - `pkg/table/tableinfo.go::setMinMax` — min/max query that runs during `SetInfo`.
  - `pkg/checksum/single.go::inspectDifferences` — PK list passed into `queryTemplate`.
- All four sites now backtick-quote columns. The composite chunker uses the existing `table.QuoteColumns` helper.

## Note on origin

This is **not** a regression from #701/#702. The unquoted `strings.Join` in the composite chunker has been there since `36877bb` (2023-08-02), when composite-key chunking was first introduced. PR #702 doesn't touch the prefetch query. The bug surfaced now only because no existing test covered a reserved-word PK column.

## Test plan

- [x] `pkg/table` regression test: composite chunker walks a table with PK `(osm_id, \`key\`)` to completion.
- [x] `pkg/table` regression test: optimistic chunker prefetch path with a single-column auto-inc PK named `\`key\``.
- [x] `pkg/migration` regression test: end-to-end ALTER through the copy path (the user's reported scenario).
- [x] `pkg/move` regression test: end-to-end move with a reserved-word composite PK.
- [x] Existing `pkg/table`, `pkg/checksum`, `pkg/copier` test suites still pass locally.

### DIFF
--- a/pkg/checksum/single.go
+++ b/pkg/checksum/single.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -122,7 +121,7 @@ func (c *SingleChecker) inspectDifferences(ctx context.Context, trx *sql.Tx, chu
 	}
 	sourceRows, err := trx.QueryContext(ctx, fmt.Sprintf(queryTemplate,
 		sourceChecksumCols,
-		strings.Join(chunk.Table.KeyColumns, ", "),
+		table.QuoteColumns(chunk.Table.KeyColumns),
 		chunk.Table.QuotedTableName,
 		chunk.String(),
 	))
@@ -146,7 +145,7 @@ func (c *SingleChecker) inspectDifferences(ctx context.Context, trx *sql.Tx, chu
 
 	targetRows, err := trx.QueryContext(ctx, fmt.Sprintf(queryTemplate,
 		targetChecksumCols,
-		strings.Join(chunk.NewTable.KeyColumns, ", "),
+		table.QuoteColumns(chunk.NewTable.KeyColumns),
 		chunk.NewTable.QuotedTableName,
 		chunk.String(),
 	))
--- a/pkg/migration/e2e_test.go
+++ b/pkg/migration/e2e_test.go
@@ -455,3 +455,26 @@ func TestMigrationCancelledFromTableModification(t *testing.T) {
 	require.Error(t, gErr)
 	require.NoError(t, m.Close())
 }
+
+// TestReservedWordPKMigration is a regression test for issue #828.
+// Migrating a table whose primary key includes columns named with MySQL
+// reserved words (like `key`/`value`) used to fail with a SQL syntax error
+// when the chunker_composite prefetch query joined chunkKeys without
+// backticks.
+func TestReservedWordPKMigration(t *testing.T) {
+	t.Parallel()
+	tt := testutils.NewTestTable(t, "reserved_word_pk_migrate", "CREATE TABLE reserved_word_pk_migrate ("+
+		"osm_id BIGINT NOT NULL, "+
+		"`key` VARCHAR(64) NOT NULL, "+
+		"`value` TEXT, "+
+		"PRIMARY KEY (osm_id, `key`)"+
+		")")
+	tt.SeedRows(t, "INSERT INTO reserved_word_pk_migrate (osm_id, `key`, `value`) "+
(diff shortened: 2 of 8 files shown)


Kiran01bm (approved):

aparajon (approved):

Kiran01bm:
these needs fixing as well ?

>https://github.com/block/spirit/blob/8db6dd43c0b75e737649d16af7b85a623286a597/pkg/table/chunker_composite.go#L92 and https://github.com/block/spirit/blob/8db6dd43c0b75e737649d16af7b85a623286a597/pkg/table/chunker_composite.go#L102 ?

morgo:
Good catch — yes, the index name passed into `FORCE INDEX(...)` had the same issue. A secondary index whose name is a reserved word (e.g. `key`) would break the prefetch query. Fixed in 44eb1cd by quoting `t.keyName` via `QuoteColumns` and extending the regression test to cover the path via `SetKey("key", "")`. Verified locally that `FORCE INDEX (\`key\`)` is accepted by MySQL.

PR #838, opened 2026-05-11
feat(#475): enable buffered-map dedup by default, remove force-enable-buffered-map flag

### LINKED ISSUE #475: Enable force-enable-buffered-map by default; remove option

morgo:
## Summary

Closes #475.

- Removes the `--force-enable-buffered-map` migration flag and the equivalent `Move.ForceEnableBufferedMap` field.
- `bufferedMap` now always uses LWW map dedup during the copy phase for non-memory-comparable PKs, with the FIFO queue kicking in post-copy via `SetWatermarkOptimization` (the previous "optimization" path is now the only path).
- Post-cutover checksum (`FixDifferences=true`) still repairs any in-window divergence, which is what made it safe to add the flag in the first place — there is no longer a technical reason to keep the opt-in around.
- Drops the "queue full-time" code path and the doc/test scaffolding that supported it (`TestBufferedMapQueueFullTimeDefault`, the dual-subtest split in `TestMoveWithVarcharPK`, the cutover-test note explaining why we *didn't* add a force-enable variant).

## Test plan

- [x] `go build ./...` and `go vet ./...` clean
- [x] `TestBufferedMapQueueModeRouting` (unit) passes against the new single-path logic
- [x] `TestBufferedMapQueueModeBackpressure` (unit) passes after switching to `watermarkOptimization=false` as the queue-mode trigger
- [x] Full CI run, including `TestMoveWithVarcharPK` and `TestCutoverAtomicityWithConcurrentWrites` against the unified bufferedMap

### DIFF
--- a/docs/migrate.md
+++ b/docs/migrate.md
@@ -20,7 +20,6 @@ spirit migrate --host mydb:3306 --username root --password secret \
 - [conf](#conf)
 - [database](#database)
 - [defer-cutover](#defer-cutover)
-- [force-enable-buffered-map](#force-enable-buffered-map)
 - [host](#host)
 - [lint](#lint)
 - [lint-only](#lint-only)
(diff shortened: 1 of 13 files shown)


aparajon (approved):

Kiran01bm:
qq: do we have any e2e test that drives meaningful traffic through the queue path (`post‑watermark‑off`) ex: with a VARCHAR PK and a forced cutover delay - intention: to potentially catch regressions in queue ordering/coalescing/backpressure.

morgo:
> qq: do we have any e2e test that drives meaningful traffic through the queue path (`post‑watermark‑off`) ex: with a VARCHAR PK and a forced cutover delay - intention: to potentially catch regressions in queue ordering/coalescing/backpressure.

We do. See: https://github.com/block/spirit/blob/27470e908058e4c121729be06f5c7d33f2946104/pkg/migration/cutover_test.go#L280-L307

The test `TestCutoverAtomicityWithConcurrentWrites` stresses this, and has been quite stable.

Issue #270, opened 2024-02-05
Add support for resume from GTID based source failover


morgo:
Algorithmically we can support resuming from a source (old term: master) node failure. But we don't test it and thus don't advertise it.

We should be able to do this. It will need switching the binlog based coordinates to be GTIDs (not positions) - which is fine, we have no dependency on either.. and lots of tests.

PR #841, opened 2026-05-13
docs: warn that resume across Spirit binary versions is not supported

Kiran01bm:
## What and Why ?
Spirit's resume-from-checkpoint logic only catches a checkpoint-table schema mismatch (via SELECT *); semantic drift across binary versions with an unchanged schema is silently accepted. 

This change documents this caveat in the README and under checkpoint-max-age, and update the strict section to list the schema-mismatch case.

## Next Step
Also, as a next step - I am thinking of adding a `spirit_version` column to the checkpoint table; refuse resume if it doesn't match `buildinfo.Version()`.

Behaviour wise - default: warn + restart fresh; under `--strict:` hard‑fail with a new sentinel error like `ErrSpiritVersionMismatch` added to the strict whitelist at [runner.go:799](https://github.com/block/spirit/blob/cc8c7143f2fb642ec0034021c7c2697c4ec2d5a6/pkg/migration/runner.go#L799)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ Spirit periodically saves the progress of a schema change to an internal checkpo
 
 When you consider that many migrations are best measured in _days_, this feature can save you a lot of lost work and improves the predictability of large-table schema migrations.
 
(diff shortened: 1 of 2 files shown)


morgo:
> Also, as a next step - I am thinking of adding a spirit_version column to the checkpoint table; refuse resume if it doesn't match buildinfo.Version().

I don't think we need to go that far, since the data within the table (provided columns line up) is safe to restore between versions. It is reasonable to expect that we will change the checkpoint format again though, for example if we implement https://github.com/block/spirit/issues/270 it will be a format change.

morgo (approved):

Issue #847, opened 2026-05-15
Buffered-map flush can produce spurious Duplicate-entry on unique keys when binlog batch reorders an in-flight swap


morgo:
## Summary

When the `bufferedMap` flush hands the applier a batch that contains a "swap pair" — two binlog `UPDATE` events from a single source-side transaction that move a unique value between two rows — the multi-row `INSERT ... ON DUPLICATE KEY UPDATE` statement can fail with `Error 1062` because the random iteration order of the map places the activating row before the deactivating row inside the batch.

The error reported in production:

```
Error 1062 (23000): Duplicate entry '1-1-paybal-...'
  for key '_balance_changes_new.uq_posted_balance_change_id'
```

The original report mentions a unique key that includes a `STORED` generated column (`is_posted_balance_change`), but generated columns are only a *trigger* for the bug, not its cause — see "Generated columns are not required" below.

## This is a regression of the #746 fix

The bug was introduced by the changes that fixed [#746](https://github.com/block/spirit/issues/746) — specifically [#821](https://github.com/block/spirit/pull/821), which removed the legacy `deltaMap` subscription type and made `bufferedMap` the only path. Worth understanding *how* #746's fix opened this failure mode, because the lesson generalises.

**Pre-#821 behaviour (the `deltaMap`):**
The `deltaMap` stored only PK hashes. At flush time it issued:

```sql
REPLACE INTO _new (cols) SELECT cols FROM source FORCE INDEX (PRIMARY) WHERE (pk) IN (...);
```

`REPLACE INTO` has a property that turns out to be load-bearing here: per the manual,

> REPLACE works exactly like INSERT, except that if an old row in the table has the same value as a new row for a PRIMARY KEY *or a UNIQUE index*, the old row is deleted before the new row is inserted.

That "or a UNIQUE index" part means each row processed in the multi-row REPLACE deletes *any* row that would collide with it on *any* unique index — not just the PK. For a swap pair `(A: slot='S'→NULL, B: slot=NULL→'S')`, REPLACE in any order works: B's insert sees A still holding 'S', deletes A (because of the unique-key conflict), then inserts B; A's deactivation event arrives in the same batch and re-inserts A with `slot=NULL`. End state is correct regardless of batch order. We had order-independence for free.

**Why #746 forced a change:**
[...]

PR #851, opened 2026-05-15
fix(#847): skip already-flushed events on rewind to avoid stale replay

morgo:
## Summary

Follow-up to #850, which is now merged. A production deployment of the swap-pair recovery surfaced a remaining failure mode: after `requestRewind` restarts the binlog reader at position 4 of `flushedPos.Name`, the syncer re-reads every event between position 4 and the old `flushedPos.Pos` — events that were already applied to the destination by a prior flush. Replaying those (now stale) row images on top of the newer destination state can recreate the very 1062 the rewind was meant to resolve.

### The scenario

User reports a `balance_changes` migration in `financialaccount_staging` that hit the swap-pair 1062, recovered as designed (cleared map, flipped `forceQueueMode`, rewound to position 4 of file `mysql-bin-changelog.002556` from `flushedPos.Pos=32895440`), and then abandoned post-rewind. Excerpt:

```
WARN buffered-map flush hit a duplicate-key collision; clearing pending changes, switching to queue mode (FIFO), and rewinding binlog to flushed position
   table=financialaccount_staging.balance_changes
   error=failed to upsert rows: ... Error 1062 (23000): Duplicate entry '1-1-paybal-wazf2VR-T3SwVdcV0SrscA' for key '_balance_changes_new.uq_posted_balance_change_id'
   changes_cleared=20 queue_cleared=0
INFO Recreating streamer from file start file=mysql-bin-changelog.002556 new_start_position=(mysql-bin-changelog.002556, 4) rewinding_to_flushed_pos=true
INFO begin to sync binlog from position position=(mysql-bin-changelog.002556, 4)
...
INFO releasing metadata lock lock_name=financialaccount_sta.balance_changes-598f7eb9
```

### Why it failed

Concrete reproduction of the post-rewind 1062:

1. Pre-T1 source state: row 17 has `posted_amount=NULL`, row 18 has `posted_amount=100` (so `is_posted_balance_change=1` on row 18; unique-key bucket `(1, 'P', 'T', 'A')` is held by row 18).
2. T1 swaps: row 17 → `posted_amount=100`, row 18 → `posted_amount=NULL`. Two UPDATE events at positions P1a (row 17) and P1b (row 18). Source-valid: between them no two rows have `is_posted_balance_change=1` on the same bucket.
3. Original flush applies both events in binlog order. Destination now matches source: row 17 active, row 18 inactive. flushedPos advances past P1b.
4. Many more transactions follow; eventually the swap-pair 1062 fires on an unrelated transaction. `handleFlushError` clears the map, flips `forceQueueMode`, requests a rewind. recreateStreamer restarts at position 4.
5. The syncer re-reads T1's events. They flow into the FIFO queue and out through `flushQueueLocked`:
   - Replay event P1a (`row 17 → posted_amount=100`): destination row 17 is *already* active (from the original apply). But row 18 was activated by a *later* transaction that was also in the rewound range. If the later transaction is ordered such that its row-18-activate hasn't replayed yet, the destination has row 18 still active *and* we're now upserting row 17 to active — `Error 1062` on `(1, 'P', 'T', 'A')`.

The general shape is "replaying an old `activate` event for row A while the destination's current state still has row B activated for the same unique-key bucket." FIFO ordering inside a single batch doesn't help here because the colliding events were applied across separate batches originally — the rewind doesn't reconstruct the inter-batch state.

### Fix

Filter binlog events at the readStream dispatch layer. A new helper `Client.eventAlreadyFlushed` compares `(currentLogName, ev.Header.LogPos)` against `c.flushedPos`; when the event ends at or before `flushedPos`, it is skipped. `bufferedPos` still advances normally through the existing position update after the switch, so the filter only suppresses subscription buffering, not the syncer's forward motion.

In normal forward-flowing operation the comparison is always false — events flow past `flushedPos` by construction since `flushedPos` only advances after its events have been applied — so the filter is a no-op outside the rewind path. After a rewind, the filter skips events ≤ `flushedPos` (already applied), buffers events > `flushedPos` (the unapplied tail), and lets the FIFO queue mode handle them in binlog order without any stale-replay collision risk.

## What changed

- [pkg/repl/client.go](https://github.com/block/spirit/blob/fix-rewind-replay/pkg/repl/client.go): add `Client.eventAlreadyFlushed(name, pos)`; call it at the top of the `RowsEvent` case in `readStream` and `break` out of the switch if true. Position 0 (FormatDescription, synthetic rotates) is never skipped.
- [pkg/repl/subscription_buffered_swap_test.go](https://github.com/block/spirit/blob/fix-rewind-replay/pkg/repl/subscription_buffered_swap_test.go): unit test covering all the edges — pos=0, earlier file, same file before / at / after `flushedPos`, later file.
- [pkg/repl/README.md](https://github.com/block/spirit/blob/fix-rewind-replay/pkg/repl/README.md): document the filter as step 4 of the recovery flow (the existing "applying re-read events in FIFO order" became step 5).

## Tests

- `TestEventAlreadyFlushedSkipsBeforeFlushedPos` — unit test for the helper.
- All existing repl swap-pair tests (`TestBufferedMapSwapPairRecoversViaQueueMode`, `TestSwapPairFullRecoveryViaQueueMode`, `TestFlushDoesNotRegressFlushedPosAfterRewind`) continue to pass — they didn't exercise this corner because their pre-flush state was always trivial.

## Test plan

- [x] `go test ./pkg/repl/... -count=1` passes
- [x] `go test ./pkg/repl/ -run 'TestEventAlreadyFlushedSkipsBeforeFlushedPos|TestFlushDoesNotRegressFlushedPos|TestBufferedMapSwapPair|TestSwapPair' -count=10` passes (catches flakiness)
- [x] `golangci-lint run ./pkg/repl/...` clean

### DIFF
--- a/pkg/repl/README.md
+++ b/pkg/repl/README.md
@@ -125,7 +125,16 @@ The subscription recovers automatically:
    `TableMapEvent`s — MySQL does not re-send TableMaps mid-file, so
    restarting at any later offset would leave the parser unable to
    decode subsequent `RowsEvent`s.
-4. Re-read events flow into the now queue-mode subscription in
+4. `readStream` filters out events ≤ `flushedPos` on the way in
+   (`Client.eventAlreadyFlushed`). Those events were already applied
+   to the destination by a prior flush; replaying them against the
+   now-newer destination state would overwrite correct rows with
+   stale row images and can recreate the very 1062 collision the
+   rewind was meant to resolve (e.g. an old "activate" event for
+   row A while the destination's current state has row B activated
+   for the same unique-key bucket). Only events strictly past
+   `flushedPos` enter the subscription's queue.
+5. The remaining events flow into the now queue-mode subscription in
    binlog (FIFO) order. `flushQueueLocked` carries that order into
    the multi-row upsert, so the swap pair lands deactivate-first
    and no longer collides. The subscription stays in queue mode for
--- a/pkg/repl/client.go
+++ b/pkg/repl/client.go
@@ -301,6 +301,34 @@ func (c *Client) setBufferedPos(pos mysql.Position) {
 	c.bufferedPos = pos
 }
 
+// eventAlreadyFlushed reports whether a binlog event at the given
+// position is at or before c.flushedPos and therefore corresponds to
+// changes already committed to the destination table by a prior flush.
+//
+// In normal forward-flowing operation this is always false — flushedPos
+// only advances *after* its corresponding events have been applied, so
+// every new event has a position strictly past it. It only matters
+// during the rewind path (block/spirit#847): recreateStreamer can
+// restart the binlog reader at position 4 of c.flushedPos.Name to pick
+// up the file's TableMap events, and the events between position 4 and
+// the old flushedPos.Pos are events we already applied. Replaying them
+// would overwrite *newer* destination state with stale row images and
+// can recreate the very 1062 collision the rewind was meant to resolve
+// (an old "activate" event for row A while the destination's current
+// state has row B activated for the same unique-key bucket). Skipping
+// them keeps the destination at its already-correct state for those
+// rows; only events strictly past flushedPos are buffered for replay.
+func (c *Client) eventAlreadyFlushed(currentLogName string, logPos uint32) bool {
+	if logPos == 0 {
+		// FormatDescription / synthetic rotate / etc.
+		return false
+	}
+	c.Lock()
+	defer c.Unlock()
+	eventPos := mysql.Position{Name: currentLogName, Pos: logPos}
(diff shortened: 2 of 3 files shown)


morgo:
Superseded by #853, which goes further: rather than detect-and-switch + rewind, it makes the applier itself idempotent by switching to `REPLACE INTO ... VALUES`. That removes the rewind machinery entirely (no `requestRewind`, no rewind branch in `recreateStreamer`, no `eventAlreadyFlushed` filter) and the corresponding failure modes — REPLACE's "delete any unique-key conflict before each insert" semantics make the multi-row VALUES list order-independent and re-replay-safe.

Closing this in favor of #853.

PR #875, opened 2026-05-20
docs(migrate): tuning parallel replication for Spirit workloads on binlog-based replicas

Kiran01bm:
## What and Why ?
Adds a "Tuning parallel replication for Spirit workloads" subsection under `docs/migrate.md` § `replica-max-lag`, expanding the existing "increase parallel replication threads" bullet with the two MySQL 8.0+ settings that actually move the needle for Spirit workloads:

- `binlog_transaction_dependency_tracking = WRITESET` on the **source** (3–10× apply boost)
- `replica_preserve_commit_order = OFF` on the **replica** (additional 1.3–2× on top) — gated on verifying no downstream consumer depends on source commit order

Briefly explains why Spirit's per-chunk `INSERT IGNORE ... SELECT` transactions head-of-line-block the replica under default `COMMIT_ORDER`, and notes the Aurora `pending-reboot` caveat.

Docs-only change.


docs/migrate.md
```diff
@@ -258,10 +258,19 @@ It is recommended that you use Spirit in combination with either parallel replic
...
+- Adjusting the configuration of your replicas to increase the parallel replication threads (see [Tuning parallel replication for Spirit workloads](#tuning-parallel-replication-for-spirit-workloads) below)
 - Temporarily disabling durability on the replica (i.e. `SET GLOBAL sync_binlog=0` and `SET GLOBAL innodb_flush_log_at_trx_commit=0`)
 - Increasing the `replica-max-lag` or disabling replica lag checking temporarily
 
+#### Tuning parallel replication for Spirit workloads
+
+Spirit's row-copier copies each chunk inside a single `INSERT IGNORE INTO _<table>_new (...) SELECT ... FROM <table> WHERE <pk range>` transaction (default up to 4 in flight, each sized to a `--target-chunk-time` of 500ms). Each one lands in the binlog as a single multi-row transaction — that is what creates the head-of-line-blocking opportunity under `COMMIT_ORDER`: the replica's parallel workers serialize behind that transaction, so a busy table with many indexes can drive the replica lag into the multi-hour range even when the source has spare CPU. Two MySQL 8.0+ replication settings make a large difference for this workload:
```
morgo: This is only true by default (unbuffered row copier). Rather than mention this maybe generalize it to say the copier threads are parallel and disjoint, and so they make good use of WRITESET extraction.

Kiran01bm: fair point - ive made the changes now.. hope this is ok now..

docs/migrate.md
```diff
@@ -258,10 +258,21 @@ It is recommended that you use Spirit in combination with either parallel replic
...
 - Temporarily disabling durability on the replica (i.e. `SET GLOBAL sync_binlog=0` and `SET GLOBAL innodb_flush_log_at_trx_commit=0`)
 - Increasing the `replica-max-lag` or disabling replica lag checking temporarily
 
+#### Tuning parallel replication for Spirit workloads
+
+Spirit's row-copier runs multiple chunks in parallel (default up to 4 in flight, each sized to a `--target-chunk-time` of 500ms), and each chunk covers a **disjoint primary-key range** of the source table. Every chunk lands in the binlog as its own multi-row transaction, and because the ranges are disjoint the chunks have no row-level write conflicts with each other. That is exactly the workload writeset-based dependency tracking is designed to parallelize on the replica — under the default `COMMIT_ORDER` scheduling, however, the replica's parallel workers must serialize on source commit order, so a busy table with many indexes can drive replica lag into the multi-hour range even when the source has spare CPU. Two MySQL 8.0+ replication settings make a large difference for this workload:
```
morgo:
```suggestion
Spirit's row-copier runs multiple chunks in parallel (default up to 4 in flight, each sized to a `--target-chunk-time` of 500ms), and each chunk covers a **disjoint primary-key range** of the source table. Every chunk lands in the binlog as its own multi-row transaction, and because the ranges are _disjoint_ the chunks have no row-level write conflicts with each other.

This workload is exactly the right shape to execute in parallel on replicas, but under MySQL 8.0 defaults (`COMMIT_ORDER` scheduling) there is minimum parallelism, and it requires the following configuration changes:

```

morgo (approved):

PR #879, opened 2026-05-21
enhance status API

### LINKED ISSUE #844: enhance status API
When spirit is invoked via the API, it is possible to call the status API:

https://github.com/block/spirit/blob/a1c422e2a42907a047f30c9025b68a1abeae8256/pkg/status/progress.go#L3-L21

We use this internally, but noticed two limitations we would like to address:

1. On recovery the state is effectively "reset" because it must walk through copyRows, checksum, etc - even if these states are brief. For GUI clients if they had a pod failure while waiting on the sentinel this can look confusing. What might be easier is advertising the CurrentState as "recovering". We can leave the Summary and TableProgress unchanged -- it's just a string change of the human readable state.

2. We should provide context on if the migration is throttled. Currently this is only available by looking at the logs.

For an implementation of (1) the thoughts are roughly:
if usedResumeFromCheckpoint == true *and* the copier progress && checksum progress is very high.

morgo:
Closes #844.

## Summary

Surfaces resume-from-checkpoint outcome and the active throttler on the runner/Progress API so external clients (GUIs, automation) can summarize migration state without scraping logs.

> **Note:** This branch originally also removed `--strict` mode, but that landed separately in #910. After rebasing onto `main`, this PR is scoped to the status-API enhancements only.

### Runner getters (migration + move)

- `UsedResumeFromCheckpoint() bool`
- `ResumeError() error` — preserves typed `status.Err*` values for `errors.Is` matching

Both fields are guarded by a `sync.RWMutex` so concurrent reads via the getters cannot race with the `setup()` goroutine's write.

### Throttler interface

`Throttler` gains a `String() string` method — empty when not throttled, descriptive when active. Implemented on:

- `Noop` → always `""`
- `Mock` → `"mock throttler (always throttled)"`
- `Replica` → `"replica lag %dms exceeds tolerance %s"`
- `CommitLatency` → `"Aurora commit latency %s exceeds threshold %s"`
- `ActiveThreads` → `"Aurora active threads %d exceed vCPUs %d"`
- `multiThrottler` → joins non-empty children with `; `

### Progress struct

`status.Progress` gains a `Throttler throttler.Throttler` field. Guaranteed non-nil — both runners initialize `r.throttler` to `&throttler.Noop{}` so callers can dispatch on it without nil checks. Surfacing the throttler itself (rather than cherry-picking fields) means future throttler-specific data is available without amending `Progress`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/throttler/...` — green
- [ ] CI to run the full `-race` suite against MySQL
- [ ] Reviewer to sanity-check that `Progress.Throttler` (interface value in a returned struct) is acceptable shape for downstream consumers

### DIFF
--- a/pkg/migration/binlog_test.go
+++ b/pkg/migration/binlog_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/block/spirit/pkg/status"
 	"github.com/block/spirit/pkg/table"
 	"github.com/block/spirit/pkg/testutils"
+	"github.com/block/spirit/pkg/throttler"
 	"github.com/block/spirit/pkg/utils"
 	"github.com/stretchr/testify/require"
 )
@@ -126,7 +127,7 @@ func TestE2EBinlogSubscribingCompositeKey(t *testing.T) {
 	m := NewTestRunner(t, "e2et1", "ENGINE=InnoDB", WithBuffered(false))
 	defer utils.CloseAndLog(m)
 	require.Equal(t, "initial", m.status.Get().String())
-	require.Equal(t, status.Progress{CurrentState: status.Initial, Summary: "", Tables: nil}, m.Progress())
+	require.Equal(t, status.Progress{CurrentState: status.Initial, Summary: "", Tables: nil, Throttler: &throttler.Noop{}}, m.Progress())
 
 	// Usually we would call m.Run() but we want to step through
 	// the migration process manually.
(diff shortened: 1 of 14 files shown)


morgo:
Forking the strict removal to https://github.com/block/spirit/pull/910 while we experiment more with this.

PR #886, opened 2026-05-28
[codex] Add sentinel readiness status

deanbaker:
🤖

## Context
We want to support an automated sentinel-drop pipeline that runs once a day during low-traffic windows. That pipeline should only drop `_spirit_sentinel` when Spirit is actually ready to proceed to cutover.

For multi-day migrations, the sentinel table can exist for a long time while copy/checksum work is still in progress. With the current table-existence-only signal, automation cannot distinguish "sentinel exists because cutover is deferred and ready" from "sentinel exists because the migration is still copying/checksumming." Dropping it too early can remove the operator safety gate before Spirit has reached the cutover wait point.

## Summary
- Add a `status` column to Spirit-created sentinel tables.
- Initialize sentinel rows as `status = 'copying'`.
- Mark sentinel rows as `status = 'ready'` only when Spirit reaches the sentinel wait point before cutover.
- Keep legacy/manual sentinel tables compatible: if the sentinel exists without a `status` column, Spirit still waits for it to be dropped.
- Treat a sentinel dropped during the ready-status update race as a normal release, not a migration failure.
- Document the readiness signal for both `migrate` and `move`.

## Operator Impact
Automation can now query the sentinel row and only drop the table when it sees `ready`, making daily low-traffic cutover windows safe for migrations that run across multiple days.

## Validation
- `go test ./pkg/migration ./pkg/move -run '^$'`


morgo:
We've changed the design of defer-cutover so that it blocks after the checksum has already run. This means that when its ready, it cuts over immediately.

W.r.t. detecting if its ready, I would like to address this through the status/Progress API: https://github.com/block/spirit/issues/844

We support using the API and invoking spirit through golang as a first-class feature. I'd rather support/lean into this than have external code read the artifacts it creates, since if we change the file formats it can break external scripts in ways that are difficult to predict. We want to retain the flexibility to be able to change these artifacts as well.

With that, I'm going to close this for now, but please provide feedback on the status/Progress API improvements. We'd be happy to include your use cases.

PR #920, opened 2026-06-10
escape separator in hash keys to prevent collisions

morgo:
## Problem

`utils.HashKey` built composite-PK map keys by joining components with the literal separator `-#-`, and `UnhashKeyToString` split on it — with no escaping. Since go-mysql decodes VARCHAR PKs as Go strings, PK values containing `-#-` are realistic. Two failure modes:

1. **Silent change loss**: `('a-#-b', 'c')` and `('a', 'b-#-c')` produce the identical key `a-#-b-#-c`, so the second binlog event overwrites the first row's buffered change in the subscription's dedup map — one row's change is never applied (checksum is the only backstop).
2. **Permanently failing flush**: a single-column PK value containing `-#-` is split into a multi-value tuple, producing `DELETE ... WHERE (pk) IN (('a','b'))` → MySQL error 1241 "Operand should contain 1 column(s)", blocking the flush forever.

## Fix

Escape each component when hashing (`\` → `\\`, then `#` → `\#`) before joining: every user-data `#` is backslash-prefixed while the separator's `#` never is, making the split unambiguous. `UnhashKeyToString` unescapes each part before SQL-escaping, so the final SQL contains the original value, correctly quoted.

Hash keys are purely in-memory (dedup maps and applier flush paths) and never persisted to checkpoints, so resume compatibility is unaffected. Keys without special characters hash byte-identically to before.

## Testing

- Round-trip and collision unit tests (separator, bare `#`, `\`, empty components, boundary dashes, non-string types).
- `TestBufferedMapSeparatorInPKValues`: end-to-end map-mode buffered subscription with composite VARCHAR PKs; the two previously-colliding rows occupy distinct slots and flush correctly through the real applier.
- `TestSingleTargetApplierDeleteKeysSeparatorInValues`: covers the previous error-1241 case and asserts only the intended row is deleted.
- All three new tests verified to FAIL against the old code.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -17,22 +17,55 @@ const (
 
 // HashKey is used to convert a composite key into a string
 // so that it can be placed in a map.
+// Each component is escaped before joining so that values containing
+// the separator cannot collide with key boundaries: without escaping,
+// ("a-#-b", "c") and ("a", "b-#-c") would hash to the same string,
+// causing one row's buffered change to silently overwrite another's.
 func HashKey(key []any) string {
-	var pk []string
+	pk := make([]string, 0, len(key))
 	for _, v := range key {
-		pk = append(pk, fmt.Sprintf("%v", v))
+		pk = append(pk, escapeHashComponent(fmt.Sprintf("%v", v)))
 	}
 	return strings.Join(pk, PrimaryKeySeparator)
 }
 
+// escapeHashComponent escapes a single key component so the separator can
+// never appear inside it: backslashes are doubled first, then every "#" is
+// prefixed with a backslash. After escaping, every "#" originating from
+// user data is preceded by a backslash, while the bare "#" in the
+// separator never is — so splitting on the separator is unambiguous.
+func escapeHashComponent(s string) string {
+	s = strings.ReplaceAll(s, `\`, `\\`)
+	return strings.ReplaceAll(s, `#`, `\#`)
+}
+
+// unescapeHashComponent reverses escapeHashComponent: each backslash
+// escapes the byte that follows it.
+func unescapeHashComponent(s string) string {
+	if !strings.Contains(s, `\`) {
+		return s
+	}
+	var b strings.Builder
+	b.Grow(len(s))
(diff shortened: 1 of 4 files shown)


pkg/utils/utils_test.go
```diff
@@ -29,6 +29,56 @@ func TestHashAndUnhashKey(t *testing.T) {
...
+		want string
+	}{
+		{"plain single", []any{"hello"}, "'hello'"},
+		{"plain composite", []any{"a", "b"}, "('a','b')"},
+		{"separator in single key", []any{"a-#-b"}, "'a-#-b'"},
+		{"separator in composite components", []any{"a-#-b", "c"}, "('a-#-b','c')"},
+		{"separator in second component", []any{"a", "b-#-c"}, "('a','b-#-c')"},
+		{"bare hash", []any{"a#b", "c#"}, "('a#b','c#')"},
+		{"backslash", []any{`a\b`, `c\`}, `('a\\b','c\\')`}, // sqlescape doubles backslashes for SQL
+		{"backslash before hash", []any{`a\#b`}, `'a\\#b'`},
```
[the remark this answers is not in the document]
morgo: False positive.

aparajon (approved):

pkg/utils/utils.go
```diff
@@ -17,22 +17,55 @@ const (
...
+// unescapeHashComponent reverses escapeHashComponent: each backslash
+// escapes the byte that follows it.
+func unescapeHashComponent(s string) string {
+	if !strings.Contains(s, `\`) {
+		return s
+	}
+	var b strings.Builder
+	b.Grow(len(s))
+	for i := 0; i < len(s); i++ {
+		if s[i] == '\\' && i+1 < len(s) {
```
eeSeeGee, line 53: How would an unaccompanied slash end up at the end of the string? 🤔

eeSeeGee (approved):

pkg/utils/utils.go
morgo: It can't - that branch is purely defensive. I'm going to leave it in, but I'll add a comment.

PR #938, opened 2026-06-10
narrow PREFERRED TLS fallback and normalize tls-mode case

morgo:
## Problem
Two TLS bugs in `pkg/dbconn`:
1. **PREFERRED silently downgrades the whole pool to plaintext on ANY ping failure** — a transient blip, exhausted connections, bad credentials, or a certificate-verification failure all quietly dropped TLS for the rest of the migration with no log line. PREFERRED is the default mode.
2. **TLS mode strings are documented case-insensitive but matched exact-case** in several places; most importantly `GetTLSConfigForBinlog` turned a lowercase `disabled` into a non-nil TLS config for the binlog/replication connection on RDS hosts while the main pool ran plaintext.

## Fix
Narrow the PREFERRED fallback to the one case it was designed for, detected via go-sql-driver's `mysql.ErrNoTLS` sentinel (the only error meaning "server doesn't support TLS"); everything else propagates as an error and the genuine downgrade is logged at warn level. Normalize `TLSMode` at the entry points and add an explicit `DISABLED` case to the binlog switch.

## Testing
Unit test for the error classifier; case-insensitivity across pool creation and binlog config (DISABLED → nil); the narrowed fallback returns an error on a non-TLS failure. The binlog-DISABLED-on-RDS test was verified to fail on the unfixed code. Full pkg/dbconn and pkg/change suites pass.

### DIFF
--- a/pkg/dbconn/conn.go
+++ b/pkg/dbconn/conn.go
@@ -7,6 +7,7 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


eeSeeGee (approved):
Spooky bug!

PR #985, opened 2026-06-23
copier: expose remaining copy time as a structured ETA

aparajon:
`Progress` carried the copy ETA only inside the human-readable `Summary` string, so programmatic consumers (e.g. a control plane rendering per-table progress) had to parse log output to recover it — and couldn't tell "still measuring" from a real estimate.

This adds a structured ETA to the status API:
- `status.ETA` — a small value bundling `State` and `Duration`:
  - `State`: `measuring` (no copy rate yet, the `TBD` window), `ready` (`Duration` is valid), `due` (copy essentially complete, the `DUE` case), or none (not in row copy). Lets a consumer render "calculating…" during the initial window instead of a misleading `0`.
  - `Duration` (`time.Duration`): the estimated remaining row-copy time, valid only when `State == ready`.
- `status.Progress.ETA` — the structured estimate.
- `Copier.GetETAState() status.ETA` — computes the state and duration in one read so they stay consistent.

A shared `etaEstimate` helper backs `GetETA` and `GetETAState`, removing the duplicated estimate math the buffered and unbuffered copiers each carried and keeping the string and structured forms in sync.

### DIFF
--- a/pkg/copier/copier.go
+++ b/pkg/copier/copier.go
@@ -55,12 +55,12 @@ func etaEstimate(copiedRows, totalRows uint64, pct float64, rowsPerSecond uint64
 type Copier interface {
 	Run(ctx context.Context) error
 	GetETA() string
-	// GetETAState returns whether a copy ETA is available and, when it is
-	// (status.ETAReady), the estimated remaining time (0 otherwise). It is the
-	// structured counterpart of GetETA, returning the state and duration together
-	// so callers see a consistent pair and can distinguish "still measuring" from
-	// a real estimate or a near-complete copy.
-	GetETAState() (status.ETAState, time.Duration)
+	// GetETAState returns the structured copy ETA: its availability (so callers
+	// can distinguish "still measuring" from a real estimate or a near-complete
+	// copy) and, when available, the estimated remaining time. It is the
+	// structured counterpart of GetETA, computed in a single read so the state
+	// and duration are always consistent.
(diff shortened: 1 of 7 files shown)


pkg/copier/copier.go
```diff
@@ -30,6 +53,12 @@ const (
 type Copier interface {
 	Run(ctx context.Context) error
 	GetETA() string
+	// GetETAState returns whether a copy ETA is available and, when it is
+	// (status.ETAReady), the estimated remaining seconds (0 otherwise). It is the
+	// structured counterpart of GetETA, returning the state and seconds together
+	// so callers see a consistent pair and can distinguish "still measuring" from
+	// a real estimate or a near-complete copy.
+	GetETAState() (status.ETAState, int64)
```
morgo: I would rather return `time.Duration` than seconds as an int64.

morgo (commented):
Do you want to merge in whats in https://github.com/block/spirit/pull/879 as well? These changes all serve the same purpose essentially.

pkg/copier/copier.go
aparajon: 🤖 Done in 787b505 — `GetETAState` now returns `(status.ETAState, time.Duration)` and `status.Progress.ETA` is a `time.Duration`, dropping the int64 seconds conversion so the value stays idiomatic end to end.

morgo (approved):

PR #986, opened 2026-06-23
status: expose structured checksum progress

aparajon:
`Progress` carried checksum progress only inside the `Summary` string (`"Checksum Progress=X/Y Z%"`), so programmatic consumers (e.g. a control plane rendering per-table progress) had to parse log output to recover it.

This adds:
- `Checker.GetProgressData() (rowsChecked, rowsTotal uint64)` — the structured verified/total rows, on both the single and distributed checkers.
- `status.Progress.Checksum` (`ChecksumProgress{RowsChecked, RowsTotal}`) — populated while the migration is in the checksum phase, zero otherwise.

A shared `formatChecksumProgress` helper backs both `GetProgress` (the display string) and `GetProgressData`, so the two can't drift.


pkg/checksum/checksum.go
```diff
@@ -94,6 +94,10 @@ type Checker interface {
 	// Run performs the checksum operation.
 	Run(ctx context.Context) error
 	GetProgress() string
+	// GetProgressRows returns the structured checksum progress — rows verified so
+	// far and the total to verify — so callers can render progress without
+	// parsing GetProgress's display string.
+	GetProgressRows() (rowsChecked, rowsTotal uint64)
```
morgo: I think you can change `GetProgress` to return what you have as `status.ChecksumProgress` (although there might be some cyclic import issues to work out).

When we want the string version we can then call `progress.String()`. This also makes the code simpler when there is more than one checksum implementation.

morgo (approved):

PR #997, opened 2026-06-29
checkpoint: extract shared checkpoint-table package (all three runners)

morgo:
## What

Introduces `pkg/checkpoint`, which owns the checkpoint table all three runners depend on (migration, move, datasync), and rewires each runner to use it. This is the third pillar of the runner-triplet de-duplication, after `pkg/status` and `pkg/sentinel` (#994).

Six commits, each self-contained:
1. **Extract `pkg/checkpoint` (migration + move)** — the package + the two runners that already shared the append-row model.
2. **Bring datasync on** — adds a `Persistent` mode + `Exists()` so the third runner joins too.
3. **Cap to a single row** — `REPLACE` on `id=1` instead of appending (bounds datasync's continuous sync; atomic).
4. **Key on the connection's schema; add move `--force`** — drop the schema arg (use `DATABASE()`, like sentinel); move gains `--force` wipe-and-restart recovery, mirroring datasync.
5. **Block concurrent multi-table migrations per schema** — a schema-scoped lock; lets the package drop its `Shared` mode entirely.
6. **Rename `Owned` → `Transient`** — name the mode for what it is (a checkpoint for one finite run), not why DROP+CREATE is safe.

## Why

`pkg/migration`, `pkg/move`, and `pkg/datasync` each hand-rolled their own checkpoint table: near-identical `CREATE TABLE` DDL, the create/drop lifecycle, the write, and the resume read. They had already drifted (move still carried a stale `binlog_positions` column; datasync a separate `source_position`/`updated_at` schema) — exactly the kind of silent fork this codebase can't afford in resume/checkpoint (data-loss-adjacent) territory.

## The package

`pkg/checkpoint` owns the **table mechanics**; the runners own the **policy**.

- Keys on the connection's selected schema (`DATABASE()` / unqualified names), like `pkg/sentinel` — Vitess-friendly; callers point the `*sql.DB` at the right schema rather than passing one.
- `Write` keeps a **single row** (`REPLACE` on `id=1`): atomic (a crash never leaves *no* checkpoint) and bounded (a continuous sync can't grow it). `ReadLatest` reads it; `Exists` reports table presence (datasync's resume signal); `IsIncompatible` tells an unreadable cross-version checkpoint apart from a transient read error.
- Two `Mode`s differ only in `Create`: **`Transient`** (DROP+CREATE — a checkpoint for one finite run: single-table & atomic multi-table migration, move) and **`Persistent`** (CREATE IF NOT EXISTS, never cleared — a checkpoint for a continuous run: datasync, whose existence is its resume signal).

Resume **policy** stays per-runner: migration (statement match, original-table-name collision, max-age), move (multi-source positions, checksum-watermark discard, pre-flight check-framework gating), datasync (three-state routing off `Exists`).

## Schema

All three now read/write one DDL. move adopts it (`binlog_positions` → `binlog_position`, gains `original_table_name`); datasync adopts it (`source_position` → `binlog_position`, `created_at` replaces `updated_at`). Cross-version resume was never supported, so no data migration: `ReadLatest`'s explicit column list fails the read on an incompatible table and aborts.

## Concurrency (behaviour change)

Atomic multi-table migrations all coordinate through one fixed-name `_spirit_checkpoint`/`_spirit_sentinel` per schema. The per-table metadata lock did **not** serialize them (disjoint tables share no table lock), so two could run concurrently and race on the shared checkpoint — the only reason the package needed a statement-scoped "Shared" mode.

This PR **blocks** that instead: a multi-table migration takes an additional **schema-scoped `GET_LOCK`** (via the existing `dbconn.MetadataLock`), held for the whole run, so a second one in the same schema fails fast with a clear error. Single-table migrations are unaffected — serialized per-table, free to run concurrently. With concurrency blocked, `_spirit_checkpoint` has a single owner, so the `Shared` mode is gone. Documented user-facing in README's "Atomic Multi-table changes".

## Recovery on an unreadable / incompatible checkpoint

`checkpoint.IsIncompatible` distinguishes a genuinely-unusable checkpoint from a transient read failure, so a blip never triggers a destructive recovery.
- **migration** — falls back to a fresh migration automatically.
- **move** — `--force` (new) wipes the target tables + checkpoint and copies fresh; without it, an unresumable non-empty target is a hard error that names `--force`.
- **datasync** — `--force` drops the target DB and starts fresh; the non-force path gives an actionable "drop it or re-run with --force" error.

## Testing

All under `-race`: migration (default + `-tags singleversion` resume: composite/varchar/binary PK, mismatch, collision, max-age, resume-during-checksum) including the new `TestMultiTableMigrationBlockedPerSchema`; move (single/multi-source resume, max-age, `deleteAboveWatermark`, checksum-watermark invariants, sharded determinism, `--force`); datasync (resume, no-watermark-row, force, incompatible-checkpoint recovery, copy-only, checkpoint-write-fatal); `dbconn` metadata lock; `pkg/checkpoint` (Transient/Persistent lifecycles, single-row + Exists assertions). Build, lint, and `go vet ./...` clean.


morgo:
🤖 Breaking this up into smaller, focused PRs (and dropping the description baggage these iterations accumulated):

- **#998** — `migration: block concurrent atomic multi-table migrations; use DATABASE() for own DDL` (independent, open now)
- **checkpoint extraction** — stacks on #998; opens once it lands
- **move `--force` recovery** — stacks on the checkpoint PR

The full combined work is preserved on the `quality/checkpoint-pkg` branch for reference. Closing this in favour of the split.

PR #1020, opened 2026-06-30
fix(statement): escape index and column identifiers in GetMissingSecondaryIndexes

### LINKED ISSUE #1019: GetMissingSecondaryIndexes does not escape index and column identifiers
The identifier-quoting unification in 6429c3c (#1014) routed the table name in `GetMissingSecondaryIndexes` through `sqlescape.EscapeIdentifier`, but the index name and the index column names in the same function still wrap the raw value in backticks:

```go
// pkg/statement/parse_create_table.go
fmt.Fprintf(&sb, " `%s`", constraint.Name)          // index name
...
fmt.Fprintf(&sb, "`%s`", key.Column.Name.String())  // index column
```

A name containing a backtick breaks out of the quoting, so the generated `ALTER TABLE` is invalid. For a source whose index ``i`x`` is on column ``c`1`` and a target missing that index, the output is:

```
ALTER TABLE `t` ADD INDEX `i`x` (`c`1`)
```

when it should double the backticks like the rest of the function already does (the table name and the index `COMMENT` are escaped, only the index/column identifiers were missed):

```
ALTER TABLE `t` ADD INDEX `i``x` (`c``1`)
```

This is the same break-out #1014 set out to remove. I have a small fix that routes both sites through `sqlescape.EscapeIdentifier` with a regression test, happy to send it.

mahirhir:
Follow-up to #1014 / 6429c3c, which unified identifier quoting through `sqlescape.EscapeIdentifier`. That sweep escaped the table name in `GetMissingSecondaryIndexes`, but the index name and the index column names in the same function were left wrapping the raw value in backticks:

```go
fmt.Fprintf(&sb, " `%s`", constraint.Name)
fmt.Fprintf(&sb, "`%s`", key.Column.Name.String())
```

When an index or column name contains a backtick, the output breaks out of the identifier quoting and is invalid SQL, which is the case #1014 set out to fix. For a source index ``i`x`` on column ``c`1`` the generated statement was:

```
ALTER TABLE `t` ADD INDEX `i`x` (`c`1`)
```

and is now:

```
ALTER TABLE `t` ADD INDEX `i``x` (`c``1`)
```

Both sites now go through `sqlescape.EscapeIdentifier`, matching the table name and the already-escaped index `COMMENT` in the same function. Output is unchanged for backtick-free identifiers, so the existing expectations are untouched. Added a backtick case to `TestGetMissingSecondaryIndexes`; the package unit tests and `go vet` pass (the live-MySQL cases need a running database).

Closes #1019

### DIFF
--- a/pkg/statement/parse_create_table.go
+++ b/pkg/statement/parse_create_table.go
@@ -1526,7 +1526,7 @@ func GetMissingSecondaryIndexes(sourceCreateTable, targetCreateTable, tableName 
 
 		// Add index name if present
 		if constraint.Name != "" {
-			fmt.Fprintf(&sb, " `%s`", constraint.Name)
+			fmt.Fprintf(&sb, " %s", sqlescape.EscapeIdentifier(constraint.Name))
 		}
 
 		// Add columns
@@ -1538,7 +1538,7 @@ func GetMissingSecondaryIndexes(sourceCreateTable, targetCreateTable, tableName 
 			switch {
 			case key.Column != nil:
 				// Regular column reference
-				fmt.Fprintf(&sb, "`%s`", key.Column.Name.String())
+				sb.WriteString(sqlescape.EscapeIdentifier(key.Column.Name.String()))
 				// Add length if specified
 				if key.Length > 0 {
 					fmt.Fprintf(&sb, "(%d)", key.Length)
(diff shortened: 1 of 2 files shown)


morgo (approved):

morgo:
Thanks for the fix!

PR #1022, opened 2026-06-30
checksum: honor yield timeout in distributed checker

### LINKED ISSUE #993: checksum: DistributedChecker ignores CheckerConfig.YieldTimeout (move never yields long checksum transactions)
## Summary
`CheckerConfig.YieldTimeout` is honored only by `SingleChecker`. `DistributedChecker` — returned by `checksum.NewChecker` whenever `CheckerConfig.Applier != nil` ([`pkg/checksum/checksum.go:158`](https://github.com/block/spirit/blob/main/pkg/checksum/checksum.go#L158)) — never reads it. The `yieldTimeout` field exists only on `SingleChecker` ([`pkg/checksum/single.go:42`](https://github.com/block/spirit/blob/main/pkg/checksum/single.go#L42)), which uses it to bound a checksum pass (`yieldCtx`, single.go:~507) and return `ErrYieldTimeout` so the caller can release and recreate the long-running `REPEATABLE READ` transaction — curbing InnoDB history-list growth on the source.

## Impact
`pkg/move` always constructs its checkers with `Applier: r.applier`, so it always uses `DistributedChecker`. As a result move's checksum passes — including the **continuous** checksum during the sentinel wait, which can run for hours (bounded only by `sentinelWaitLimit`, 48h) — never yield. On a busy source this holds a long-lived read view and can grow the history list unbounded.

By contrast, migration's primary checker ([`pkg/migration/runner.go:661`](https://github.com/block/spirit/blob/main/pkg/migration/runner.go#L661)) sets no `Applier`, so it is a `SingleChecker` and does honor the yield.

## Proposed fix
Plumb `yieldTimeout` into `DistributedChecker` and have it periodically end and recreate its `REPEATABLE READ` transaction pool(s) between passes/chunks, mirroring `SingleChecker`'s logic (`single.go` `yieldCtx` + `ErrYieldTimeout` handling).

## Context
Found while reviewing #988, which attempted to set `YieldTimeout` on move's checker configs. Copilot correctly flagged that as a no-op because move's checker is a `DistributedChecker`, so #988 was closed in favor of this issue.

🤖 Filed by Claude Code on behalf of @mtocker.

aouxwoux:
Closes #993.

## Summary
- pass `CheckerConfig.YieldTimeout` through to `DistributedChecker`
- add distributed checksum yield/resume handling that mirrors `SingleChecker`
- bound each distributed checksum pass with a yield context so source/target transaction pools are closed and recreated from the low watermark
- add a regression test for the config path that previously dropped `YieldTimeout`

## Tests
- `go test ./pkg/checksum -run '^TestDistributedCheckerHonorsYieldTimeoutConfig$' -count=1`
- `go test ./pkg/checksum -run '^$' -count=1`
- `go test ./... -run '^$' -count=1`

I also tried `go test ./pkg/checksum -run 'TestYieldTimeout|TestDistributedChecksum' -count=1`, but this local machine has no MySQL service at `127.0.0.1:3306`, so the existing integration tests fail immediately with connection refused before exercising the checksum logic.

### DIFF
--- a/pkg/checksum/checksum.go
+++ b/pkg/checksum/checksum.go
@@ -166,6 +166,7 @@ func NewChecker(sourceDBs []*sql.DB, chunker table.Chunker, feeds []change.Sourc
 			fixDifferences: config.FixDifferences,
 			maxRetries:     config.MaxRetries,
 			applier:        config.Applier,
+			yieldTimeout:   config.YieldTimeout,
 		}, nil
 	}
 	return &SingleChecker{
--- a/pkg/checksum/distributed.go
+++ b/pkg/checksum/distributed.go
@@ -49,6 +49,8 @@ type DistributedChecker struct {
 	differencesFound atomic.Uint64
 	recopyLock       sync.Mutex
 	maxRetries       int
+	yieldTimeout     time.Duration
+	yieldsPerformed  atomic.Uint64 // number of yield/resume cycles performed
 }
 
 var _ Checker = (*DistributedChecker)(nil)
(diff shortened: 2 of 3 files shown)


morgo (approved):

morgo:
Awesome, thank you!

PR #1037, opened 2026-07-02
fix(dbconn): stop interpreting user DDL as an escape format string

morgo:
## Problem

Several call sites build the `sqlescape` format string by concatenating raw user SQL onto a trusted prefix, e.g. `"ALTER TABLE %n ALGORITHM=INSTANT, " + c.stmt.Alter`. `sqlescape.EscapeSQL` interprets `%n` / `%?` / `%%` **anywhere** in the text — including inside string literals of the user's DDL:

- An `--alter` containing `%n` or `%?` in a literal (e.g. `COMMENT '100%new'`, `CHECK (name LIKE 'a%?')`) fails escaping with `missing arguments`. On the default force-kill path this goes through `sqlescape.MustEscapeSQL`, which **panics the process** — with the kill timer already armed.
- `%%` in a literal (e.g. `DEFAULT '50%% off'`) is silently collapsed to a single `%`, so spirit executes **different DDL than the user wrote** (`'50% off'` instead of `'50%% off'`).

Non-ALTER statements (`CREATE TABLE` / `DROP TABLE` / `RENAME TABLE`) were also executed by passing the raw user statement as the format string with zero args.

## Fix

User SQL is now always data, never a format string. `sqlescape` gains a `%r` verb that splices its argument into the SQL **verbatim** — no quoting, no format interpretation. The spliced text is never re-scanned, so `%n` / `%?` / `%%` sequences inside it are inert. (`sqlescape` is maintained as a hard fork and no longer synced from TiDB, so extending the verb set is safe; its README now documents this.)

- `%r` accepts only the new `sqlescape.RawSQL` named type — anything else, including a plain `string`, is an error. The conversion at the call site is an explicit assertion that the text is safe to splice verbatim, and `grep -rn 'RawSQL('` is a complete audit list of every raw splice in the tree. *(Adopted from the adversarial review's Finding 1.)*
- Call sites keep a compile-time-constant format string and pass the user clause as an argument, e.g. `dbconn.ForceExec(..., "ALTER TABLE %n ALGORITHM=INSTANT, %r", tableName, sqlescape.RawSQL(alter))`. Non-ALTER user statements execute via `dbconn.Exec(ctx, db, "%r", sqlescape.RawSQL(stmt))`.
- `ForceExec` now escapes with the error-returning `EscapeSQL` **before** the kill timer is armed — a bad format string returns an error instead of panicking while a timer that kills other connections is pending. Kill-timer/retry semantics are unchanged.

Call sites audited and fixed (all places where a non-constant format string reached `Exec`/`ForceExec`):

- `pkg/migration/change.go` `alterNewTable` — `ALGORITHM=COPY` attempt and the plain retry
- `pkg/migration/change.go` `attemptInstantDDL` — `ForceExec` and `Exec` variants
- `pkg/migration/change.go` `attemptInplaceDDL` — `ForceExec` and `Exec` variants
- `pkg/migration/runner.go` — non-ALTER single-statement execution (`stmt.Statement` was the format string)

All other `Exec` / `ForceExec` / `EscapeSQL` call sites were audited (re-verified against current main) and use compile-time constant format strings (incl. `checkpoint.go`'s `"CREATE TABLE %n " + tableDDL`, which concatenates two package constants); they are unchanged. `pkg/move` / `pkg/datasync` already execute fetched DDL via plain `ExecContext`.

## Testing

New regression tests (the instant-path test reproduces the pre-fix process panic `missing arguments, need 2-th arg, but only got 1 args`):

- `pkg/dbconn/sqlescape` `TestEscapeSQLRawVerb`: verbatim splice (with `%n`/`%?`/`%%` payloads), composition with `%n`/`%?`, missing-argument errors, and rejection of non-`RawSQL` arguments (plain `string`, `int`)
- `pkg/dbconn` `TestExecRawVerb` (literals reach the server verbatim; the same text placed in the format string still errors with `missing arguments` — assertion pinned per the adversarial review's Finding 2 — and a plain-string `%r` argument is rejected), `TestForceExecRawVerb` (no format interpretation and the MDL-blocker force-kill still works), `TestForceExecBadFormatString` (bad format returns an error before the kill timer is armed, instead of panicking)
- `pkg/migration` `TestPercentSignsInDDLLiterals`: unchanged from the previous revision of this PR — instant path with `%n`/`%?` in a comment, instant path with `%%` in DEFAULT/COMMENT (byte-identical to a plain-client sibling table), copy path via `alterNewTable`, and the non-ALTER `--statement` path

Runs against local compose MySQL 8.0:

- full `./pkg/dbconn/...` suite: pass
- `pkg/migration` `TestPercentSignsInDDLLiterals`: pass
- `go build`, `go vet`, `gofmt`, `golangci-lint`: clean


aparajon:
🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head `39a907bb`.

**Verdict: correct, well-scoped, and safe to land — the audit claim in the PR body holds under independent re-verification, and the regression test genuinely has teeth.** I attacked this from three directions: whether `%r` can be re-scanned or leak an argument, whether the `ForceExec` reorder changed transaction/kill-timer semantics, and above all whether the call-site audit missed a path where user SQL still reaches a format string. All three held. Nothing blocks; two optional notes below.

The most valuable thing I can report is the mutation check. I reverted only the production files (`change.go`, `runner.go`, `dbconn.go`) at head while keeping the new tests, and `TestPercentSignsInDDLLiterals` failed with exactly the panic the PR describes — `panic: missing arguments, need 2-th arg, but only got 1 args` — then passed again on restore. The test reproduces the real defect, not a proxy for it.

### Findings

**1. (optional, hardening) `%r` accepts a plain `string`, so a future misuse has no visible smell at the call site.** All four current uses pass SQL the operator authored (`--alter` / `--statement`), which spirit already executes on their behalf, so no trust boundary is crossed today and this is not a regression — the old `+` concatenation was equally verbatim. The asymmetry is in what the code *looks* like: `%r` sits beside `%n` and `%?` and reads like another escaped verb, while performing zero escaping, so someone later reaching for it with a config value or a value read from a table would introduce injection with none of the "string concatenation into SQL" smell that reviewers and linters key on. A named type — `type RawSQL string`, with the `'r'` case erroring on anything else — would cost four call-site conversions and make `grep -rn 'RawSQL('` a complete, permanent audit list of every raw splice in the tree.

**2. (optional, test hygiene) The negative case in `TestExecRawVerb` can pass for the wrong reason after a regression.** The last block creates `execraw_percent2` with the bad text in the format string and asserts only `require.Error`. On a clean database the escape fails and no table is created, so it's correct today. But if `%n`/`%?` interpretation ever stops erroring, that run fails once and leaves the table behind — and from then on the `CREATE` returns "table already exists", which satisfies `require.Error` and masks the regression permanently for that database. `require.ErrorContains(t, err, "missing arguments")` pins it to the escape failure and is immune to that.

### Action items

1. *(optional, Finding 1)* Consider a `sqlescape.RawSQL` named type for the `%r` argument so every raw splice is explicit and greppable.
2. *(optional, Finding 2)* Tighten the negative assertion in `TestExecRawVerb` to `require.ErrorContains(t, err, "missing arguments")`.

Re-ran the call-site audit independently rather than trusting the body: every `dbconn.Exec` / `dbconn.ForceExec` / `EscapeSQL` / `MustEscapeSQL` site in `pkg/` and `cmd/` uses a compile-time-constant format string, `checkpoint.go`'s `"CREATE TABLE %n "+tableDDL` really is two constants and `tableDDL` contains no `%` at all, `pkg/move/check/schema_compare.go` concatenates *after* escaping and its output is display-only, `RetryableTransaction` passes statements to `ExecContext` untouched so the copier/applier data paths never reach the escaper, and `pkg/move` / `pkg/datasync` execute fetched DDL via plain `ExecContext`; `%r` splices into `buf`, which `escapeSQL` never re-scans, and it advances `argPos` exactly like `%n`/`%?`, so composition in either order and the missing-argument and non-string errors all behave as the unit test asserts; `FormatSQL`/`MustFormatSQL` share `escapeSQL`, so the verb is consistent across every entry point; the `ForceExec` reorder leaves the `err`-keyed commit/rollback defer on the same variable and the retry on the same escaped text, and it strictly shrinks the blast radius by returning before any transaction is opened rather than panicking with a kill timer armed — after this, no remaining `MustEscapeSQL` site can be reached with caller-controlled text; `go build ./...`, `go vet`, the full `./pkg/dbconn/...` suite (39s) and `TestPercentSignsInDDLLiterals` all pass locally at head against MySQL 8.0.45; CI 11/11 green; Copilot reviewed 8/8 files with no comments.

aparajon (approved):
🤖 Approving on Armand's behalf after the adversarial correctness review above (no blocking findings). *This stamp was left by Claude Code (claude-opus-5).*

PR #1066, opened 2026-07-17
fix(checksum): compare JSON columns through a text round-trip

morgo:
## Problem

A production migration of a hot ~5.2M-row table of ~50KB JSON documents (`MODIFY COLUMN score_data JSON NULL, ADD COLUMN ...`) failed deterministically with:

```
checksum failed: checksum found differences on every attempt (3/3). This likely indicates either a bug in Spirit, or a manual modification to the _new table outside of Spirit.
```

Each attempt found *fresh* mismatches at the newest ids, with equal chunk row counts and content-only per-row differences. Repairs (recopy) succeeded, but the next attempt — a full re-pass, ~4.5h on this table — found new ones.

## Root cause

A JSON document can hold scalar types that **JSON text cannot express**: DECIMAL (creatable server-side via `JSON_OBJECT('a', CAST(169.09 AS DECIMAL(12,6)))`) and temporal/binary opaques. Spirit's buffered copier and binlog applier write JSON values as *text*, and JSON text numbers only ever parse back as INTEGER/UNSIGNED/DOUBLE — so on the target these types degrade (DECIMAL→DOUBLE, opaques→string).

The degraded value is still **equal under JSON comparison** (`src.j = new.j` returns 1), but its canonical rendering differs: a DECIMAL renders at its declared scale (`169.090000`) while the re-parsed DOUBLE renders shortest (`169.09`). The checksum hashes `CONCAT(CAST(col AS json))`, i.e. the rendered text, so every copier/applier-written row containing such a value mismatched. Under concurrent DML, divergence flows in faster than recopy repairs it → every attempt fails → migrations of such tables are **unfinishable**. (This is the documented residual of go-mysql's `RenderJSONAsMySQLText`, go-mysql PR #1147: type tags are unrepresentable in text by construction, not a renderer bug.)

## Fix

`castExpr` now checksums JSON columns as

```sql
CAST(CAST(col AS char CHARACTER SET utf8mb4) AS json)
```

on **both** sides — render to text, re-parse — collapsing each document to its text-expressible form. That is exactly the fidelity the text-based copy/replication pipeline delivers, and no less:

- genuine value differences still hash differently (`169.10` vs `169.09` is caught)
- int vs double stays distinct (`1` vs `1.0` — text *does* preserve this, so no over-relaxation)
- plain doubles are bit-stable through the round-trip (MySQL renders shortest-form, exact round-trip)

The cast order matters: the inner char cast is the *render* that collapses type tags for JSON sources, and an *identity* for text sources, where the outer parse then normalizes whitespace/key order/number notation — so the same expression also covers `TEXT→JSON` conversion migrations. (The reverse order, parse-then-render, would be a no-op for JSON sources and would not fix this.)

All checkers (single, distributed, continuous, and per-row `inspectDifferences`) build their expressions through `ChecksumExprs`, so one seam covers them consistently; nothing outside the checksum uses `castableTp`.

## Trade-off, stated plainly

The DECIMAL→DOUBLE degradation is real and is now invisible to the checksum. It is inherent to text-mediated writes — the strict checksum could only make such migrations fail, not preserve the type tags. A possible follow-up is a preflight notice when sampled JSON contains DECIMAL-typed values, making the normalization visible instead of silent.

## Tests

The underlying issue is a straightforward SQL fact about `castExpr`'s text round-trip, not something that needs concurrent DML or a live migration to demonstrate — so coverage is three focused tests in `pkg/checksum/json_test.go`, built directly against static source/target tables via `SingleChecker.runChecksum` (the same lightweight pattern `TestCorruptChecksum`/`TestChangeDataTypeDatetime` already use):

- `TestJSONChecksumDecimalTypeDegradation`: a DECIMAL-typed JSON scalar, round-tripped through text the way the applier would write it, passes the checksum.
- `TestJSONChecksumDecimalGenuineMismatchStillCaught`: a genuine value difference through the same round-trip still fails — confirms the fix isn't over-relaxed.
- `TestJSONChecksumTextToJSONConversion`: a LONGTEXT→JSON conversion where the source is `JSON_PRETTY`-formatted (whitespace differs) but semantically equivalent JSON still passes.
- `TestCastExpr` (new unit) pins the expression shapes.
- `TestBufferedMapJSONNumberRoundTrip` deliberately keeps the *strict* cast — it pins go-mysql's renderer text fidelity, a stronger property than the checksum needs; comments updated accordingly.


pkg/migration/checksum_json_dml_test.go
```diff
@@ -0,0 +1,444 @@
...
+	// first attempt's recopy and the retry would then pass vacuously.)
+	dmlDone := make(chan struct{})
+	go func() {
+		defer close(dmlDone)
+		if !waitForCopyRows(ctx, m) {
+			return
+		}
+		docs := [2]string{buildScoreDoc("1.0", 0), buildScoreDoc("2.0", 0)}
+		largeDocs := [2]string{buildScoreDoc("1.0", 900), buildScoreDoc("2.0", 900)}
+		for i := 0; ; i++ {
```
eeSeeGee: So gross, don't have any better ideas

eeSeeGee (approved):

PR #1067, opened 2026-07-20
Rename MetadataLock to AdvisoryLock

Kiran01bm:
## Summary
Spirit's `MetadataLock` is a `GET_LOCK` user-level advisory lock, but its name — and log lines like `attempting to acquire metadata lock` — led engineers reading migration logs to conclude the migration was blocking (or blocked by) application traffic. This lock never blocks queries or DDL; it only serializes spirit processes.

## What
- Renames the user-facing log and error strings from "metadata lock" to "advisory lock" (`pkg/dbconn`, `pkg/migration`, `pkg/move`).
- Renames the type and API: `MetadataLock` → `AdvisoryLock`, `NewMetadataLock` → `NewAdvisoryLock`; `pkg/dbconn/metadatalock{,_test}.go` → `advisorylock{,_test}.go`, plus call sites, test names, and local variables.
- Comments and help text that refer to genuine table locks / server-side MDL (e.g. `kill.go`, `--skip-force-kill` help, `pkg/migration/README.md`) are intentionally unchanged.

## Why
Advisory-lock contention means another spirit process; table-lock/MDL contention means hot application traffic. The shared naming kept misleading operators during incidents. Note for downstream users (e.g. schemabot): the exported type/constructor names change — a hard but trivial fix, per review discussion.


pkg/dbconn/metadatalock.go
```diff
@@ -23,6 +23,13 @@ var (
 	refreshInterval = 1 * time.Minute
 )
 
+// MetadataLock ensures that only one spirit migration operates on a table at
+// a time. Despite the name, this is *not* a MySQL table metadata lock (MDL):
+// it is a user-level advisory lock (GET_LOCK) held on a dedicated connection,
+// so it is invisible to application traffic and never blocks queries or DDL.
+// Log and error messages refer to it as an "advisory lock" to avoid confusion
+// with the server-side MDL that every ALTER TABLE briefly acquires.
+// TODO: rename the type (and this file) to AdvisoryLock in a follow-up.
```
morgo: I think you can rename here. This is used by schemabot, but it will be easy to fix from a hard breakage.

Nit: 
> Despite the name, this is *not* a MySQL table metadata lock (MDL)

It is not a table lock, but it is an MDL (they use the same system internally). It's cleaner to say "This may be confused for a table lock, which it is not".

morgo:
Basic idea LGTM, please go ahead.

pkg/dbconn/metadatalock.go
Kiran01bm: sounds good - thank you i have folded the rename into this PR (type, constructor, files, tests, call sites) and adopted the wording suggestion - ptal

morgo (approved):

PR #1072, opened 2026-07-23
Memory-based dynamic chunking for the buffered copier

morgo:
## Problem

The dynamic chunker sizes chunks by **wall-clock time per chunk** (`--target-chunk-time`, panic-shrink at 5×). That signal was designed for the unbuffered `INSERT IGNORE … SELECT` copier, where the fed-back duration measures a single server-side statement that scales with chunk size. With the buffered (DBLog) copier — the default since v0.12 — the fed-back duration instead measures **read + applier-queue-wait + write/commit**, and the dominant term (queue wait) is **size-independent**. Under write-side backpressure the loop diverges: shrinking the chunk doesn't shrink the wait, so each panic-shrink begets the next until the chunk size pins at the 10-row floor and throughput collapses.

Observed on an internal Aurora load-test: the same `ADD INDEX` table copy that completed in ~9h30m on v0.11.3 (unbuffered) projected 60–110h on v0.15.1 (buffered) and never finished, with chunk size oscillating near the 10-row floor while `copier-is-throttled=false`.

## Change

Give the two copiers two signals, matched to what their fed-back number actually measures:

- **Buffered copier (default): byte budget.** `dynamicChunkSizer.TargetChunkBytes` servos row-count against p90 bytes-per-chunk. Identical servo math (`newRows = chunkSize × target / p90`), same growth cap / floor / ceiling, same 5× panic-shrink (now on bytes). bytes/row is a stable property of the data, so the loop converges under exactly the backpressure that collapses the time loop. The buffered copier reports the in-memory size of the rows it already holds via `Chunk.ActualBytes`.
- **Unbuffered copier (`--unbuffered`) and checksum: time budget.** `--target-chunk-time` now applies to these only (help text updated). The unbuffered `INSERT IGNORE … SELECT` and the server-side checksum CRC never see row bytes, and their time signal is clean.

### New flag: `--target-chunk-size`

The byte budget is exposed as `--target-chunk-size` (bytes, default `table.DefaultTargetChunkBytes` = 16 MiB) on **migrate**, **move**, and **sync**. A zero value means "use the default" and is normalized in each program's existing default-setting path (`migration.normalizeOptions`, `move`/`datasync` `NewRunner`), so callers that construct the structs programmatically (bypassing Kong defaults) still get the byte budget rather than accidentally falling back to the time signal on the copy chunker. A reflect-based test in each package pins the literal Kong `default` tag to `table.DefaultTargetChunkBytes` so the two can't silently drift.

Why 16 MiB: a *byte* budget maps ~1:1 to pages read per chunk regardless of row width (~1024 16KB pages ≈ 16 InnoDB extents), keeping each read solidly inside InnoDB linear read-ahead / Aurora batched prefetch. Memory is not the binding constraint — the 100k-row per-read cap and the 128×1 MiB applier buffer dominate and are independent of this value.

The composite and optimistic chunkers now share the full sizing machinery (`feedbackTime`/`feedbackBytes`/`panicShrink`/`panicShrinkBytes`); the only optimistic-specific behavior (prefetch-mode switch) is injected via a pre-update hook.

## Applier statement-size splitting: kept

Considered retiring the applier's `MaxStatementSizeBytes` batching now that chunk memory is bounded at the inflow. Keeping it: it guards `max_allowed_packet` (rendered SQL, a different quantity from raw inflow bytes) and is shared by the binlog-apply path, which has no chunk to bound it. At 16 MiB it is load-bearing — each chunk fans out to ~16–48 packet-sized chunklets (also the write-pool parallelism units).

## Scope

Wires the byte-budget copy path into all three always-buffered copy runners: **migration** (the one in the RCA), plus **move** and **datasync**. The checksum and continuous-checksum chunkers keep the time signal in every program. `CopierConfig.TargetChunkTime` is now vestigial (no copier reads it; the chunker owns all sizing) and is documented as such.

## Tests

- `TestByteSignalIgnoresInflatedTime` — fed identical inflated, size-independent times, the time sizer collapses to the floor while the byte sizer stays healthy (the thesis).
- `TestFeedbackBytesConverges`, `TestFeedbackBytesSkipsEmptyChunks`, `TestPanicShrinkBytes{AboveFloorLogs,AtFloorSuppressesFlood}`, `TestDatumByteSize`, `TestRowsByteSize` (incl. the all-empty-values-still-non-zero case).
- `--target-chunk-size` zero-value defaulting per program, plus the Kong-default drift guards.
- All new helpers at 100% line coverage; existing chunker/prefetch tests unchanged and green.


morgo:
🤖 Addressed the review feedback in 9d4e8862 (datumByteSize floor + comment typo) and 28a9fa66:

- **`chunk.ActualBytes == 0` for non-empty chunks**: `datumByteSize` now floors empty var-length values at 1 byte, so a non-empty chunk can no longer report 0 bytes — `bytes == 0` now means exactly "zero-row (gap) chunk." And `feedbackBytes` no longer skips those: an empty gap chunk is fed to the history (p90 → 0 → target above `MaxDynamicRowSize`), which is what walks the chunk size up across a large auto-increment gap and lets the optimistic chunker switch to prefetch mode. This fixes the `TestChangeDatatypeLossless` timeout where the byte path previously crawled the gap row-by-row.
- **`innodb_read_ahead_threshold` typo**: fixed.
- **PR description drift**: the description is updated to reflect the new `--target-chunk-size` flag and the move/datasync wiring landing in this PR.

aparajon (approved):

PR #1078, opened 2026-07-24
Fix self-minted JSON checksum failures with asymmetric source/target casts

morgo:
## Problem

MySQL's JSON *text* parser misrounds numbers whose shortest decimal form needs
17 significant digits by ±1 ulp (MySQL bugs #116160 / #112904, unfixed through
8.0.45). Only JSON-text parsing is affected — `CAST(str AS DOUBLE)` parses
correctly. Writing p for that parser and r for the server's JSON render, some
doubles never stabilize under repeated p∘r: each parse/render cycle drifts one
more ulp (verified on 8.0.45: `4.3319172962174079e299` re-renders as `…408`,
`…409`, `…4094`, never settling).

All of Spirit's JSON write paths are text-mediated — the buffered copier
SELECTs rendered text and writes an INSERT literal the target re-parses; the
binlog applier and the move/sync appliers do the same — so for a source
document x the target always stores exactly **y = p(r(x))**, the "text image".

The checksum previously cast JSON symmetrically,
`CAST(CAST(col AS char CHARACTER SET utf8mb4) AS json)` on both sides,
comparing r(p(r(x))) with r(p(r(y))). Since y = p(r(x)), that fails whenever y
is not a fixed point of p∘r: the target side's extra re-parse adds a fresh ulp
the source side doesn't get. In a random sample of 17-significant-digit values
on 8.0.45, ~29% mis-parse and ~10% sit on non-converging orbits — for those,
no retry and no repair of any kind can ever make the symmetric checksum pass.
Tables holding full-precision doubles inside JSON self-mint checksum failures.

## Fix: asymmetric casts (the text-image contract)

For JSON columns only, the cast is now side-dependent:

| side | expression | computes |
|---|---|---|
| source | `CAST(CAST(col AS char CHARACTER SET utf8mb4) AS json)` | r(p(r(x))) — the render of the predicted text image |
| target | `CAST(col AS json)` | r(y) — the stored document, never re-parsed |

The two hash equal **iff y = p(r(x))** — the checksum now asserts exactly the
contract the write paths guarantee. Every real divergence (missed update,
stale row, wrong value) still fails. It also closes a blindness of the
symmetric form: a target holding the source *bytes* where the text image
differs used to compare equal (both sides round-tripped onto the same value);
now it is caught and repaired. Non-JSON columns are unchanged.

## Repair must produce the text image

- `SingleChecker.replaceChunk`'s server-side `REPLACE INTO … SELECT` was
  byte-faithful; under asymmetric casts that re-flags forever (the repaired
  row holds x, the source predicts p(r(x))). Its SELECT list now comes from
  the new `ColumnMapping.RepairExprs`, which wraps JSON columns in the same
  text round-trip — repair converges in one pass even for non-converging
  values, because the stored image is rendered once and never re-parsed.
- The distributed `replaceChunk` and the continuous checker's `MySQLRecopier`
  are *already* text-mediated (SELECT to client, applier re-writes SQL
  literals): exactly one round trip. JSON deliberately stays bare there — a
  wrap would apply p∘r twice and diverge. Documented in comments.

## Scope notes

- **DistributedChecker (moves)** now uses the target-side expressions for
  target queries (it previously reused the source list on both sides), so the
  contract applies uniformly — move writes are text-mediated too. This assumes
  source and target servers parse JSON text identically; the misparse bugs are
  present in all current releases, and if a future version changed parse
  behavior, a cross-version move of affected values would fail the checksum
  safely rather than pass silently.
- **ContinuousChecker** inherits the asymmetric pair through `ChecksumExprs`;
  its recopy path is applier-mediated and correct as-is.
- **Binlog applier leg**: for binlog-applied rows the "rendered text" comes
  from the pinned go-mysql fork, not from the server, so the contract
  additionally depends on the fork's renderer matching the server's r for
  every stored document. The pin's fidelity fixes provide that, and
  `pkg/change/subscription_buffered_jsonfidelity_test.go` gates it: if the
  pin were dropped, upstream's remaining render divergences (huge integral
  doubles, zero-value temporal opaques) would make the applier write a
  *different* text image, which this checksum then correctly flags. In a
  migration that is a repairable checksum failure; under deferred cutover
  the ContinuousChecker treats divergence as fatal (no recopier), so one
  affected binlog write would abort the cutover — failing safe, but the
  dependency chain is worth stating.
- **Unbuffered copy mode** produces byte-exact rows: misparse-affected values
  now false-flag once, the round-trip repair rewrites them as text images, and
  the next attempt converges. Acceptable (this is what checksum retries are
  for), and not worth plumbing the copy mode into the checker.

## Verification

- New e2e test writes a mixed population of misparse-affected doubles through
  one text round trip (the buffered-writer image): the checksum passes on the
  **first attempt**. Reverting to symmetric casts makes it fail (verified on
  8.0.45: the misparse-affected rows flag on every attempt).
- New convergence e2e: target rows hold source bytes plus one genuinely
  different document → flagged with `FixDifferences=false`; `Run` with
  `FixDifferences=true, MaxRetries=2` repairs and passes; a SQL assertion
  proves every target row then stores the text image. Reverting the repair to
  byte-faithful makes it fail (differences on every attempt).
- New distributed e2e (`TestDistributedJSONChecksumTextImage`): same
  population, target table in a separate database holding the one-round-trip
  image, checksummed by the `DistributedChecker` — which builds its checksum
  SQL independently of the single-server checker, so its side-dependent casts
  need their own pin (previously no test exercised them: reverting the
  target side to the source expression left the whole suite green). Passes
  on the first attempt with zero differences; the same reversion now fails it.
- Unit tests cover the side-dependent cast, asymmetric `ChecksumExprs`, and
  `RepairExprs` (including column renames).
- `pkg/table`, `pkg/checksum`, `pkg/datasync` suites green; JSON-related
  `pkg/change` and checksum-related `pkg/migration` tests green; `gofmt`,
  `go vet`, `golangci-lint` clean.

### DIFF
--- a/pkg/change/subscription_buffered_jsonfidelity_test.go
+++ b/pkg/change/subscription_buffered_jsonfidelity_test.go
@@ -22,14 +22,17 @@ import (
 // textual spelling interacts badly with MySQL's JSON parser, plus the
 // zero-value temporal opaques whose spelling go-mysql gets wrong.
 //
-// STATUS: the fail-today cases below are RED BY DESIGN on the current
-// go-mysql pin (v1.15.1-0.20260526024741-088eb1fbf0ea). They assert
-// correct behavior (source == destination); a go-mysql fix for
-// formatMySQLDouble (replication/json_mysql_text.go:102-111) is being
-// prepared and will turn the huge-double class green, with the
-// zero-temporal class covered by a separate follow-up
-// (replication/json_binary.go:539 and :561-563). Do not weaken these
-// assertions to make them pass.
+// STATUS: the go.mod pin (github.com/morgo/go-mysql
+// v1.16.1-0.20260723231236-3aced1dddcf4, the TEMP replace in go.mod)
+// carries the formatMySQLDouble my_gcvt-parity fix
+// (replication/json_mysql_text.go) and the zero-temporal opaque fix
+// (replication/json_binary.go), so the huge-double and zero-temporal
+// classes assert green and gate the pin: dropping the replace for an
+// upstream go-mysql without those fixes turns every CI job red here.
+// seventeen_digit_mantissa stays skipped BY DESIGN — MySQL's own JSON
+// text parser misrounds that class (bugs #116160/#112904), which no
+// client-side renderer can compensate; do not unskip it, and do not
+// weaken any assertions to make them pass.
 //
 // Every case asserts two per-row fingerprints between source and
 // destination:
@@ -71,19 +74,23 @@ type jsonFidelityCase struct {
 
 func jsonFidelityCases() []jsonFidelityCase {
 	return []jsonFidelityCase{
-		// ---- FAIL TODAY: huge integral doubles ------------------------
-		// go-mysql's formatMySQLDouble (replication/json_mysql_text.go:
-		// 102-111 at pin v1.15.1-0.20260526024741-088eb1fbf0ea) renders
-		// whole-number doubles with strconv.FormatFloat(f, 'f', 1, 64):
-		// the FULL decimal expansion, not MySQL's shortest round-trip
-		// spelling. MySQL reparses the applier's text with rapidjson's
+		// ---- PIN-GATED: huge integral doubles -------------------------
+		// Green under the current fork pin; red if the pin is dropped for
+		// an unfixed upstream go-mysql. Upstream's formatMySQLDouble
+		// (replication/json_mysql_text.go:102-111 at
+		// v1.15.1-0.20260526024741-088eb1fbf0ea) rendered whole-number
+		// doubles with strconv.FormatFloat(f, 'f', 1, 64): the FULL
+		// decimal expansion, not MySQL's shortest round-trip spelling.
+		// MySQL reparses the applier's text with rapidjson's
 		// normal-precision path, which is NOT correctly rounded (MySQL
 		// bugs #116160 / #112904), so long expansions can land on a
 		// different double than the source stored. The source values here
 		// are built with SQL DOUBLE constructors, which go through
-		// my_strtod and ARE correctly rounded.
+		// my_strtod and ARE correctly rounded. The pinned fork renders
+		// my_gcvt-parity shortest forms instead, which reparse exactly.
 		//
-		// Verified against MySQL 8.0.45; per-key expectations today:
+		// Verified against MySQL 8.0.45; per-key expectations under the
+		// UNFIXED upstream renderer (what this case guards against):
 		//
 		//   'a' 1e16:    applier writes "10000000000000000.0" where MySQL
 		//        renders "1e16" — but the expansion reparses exactly, so
@@ -115,17 +122,18 @@ func jsonFidelityCases() []jsonFidelityCase {
 			assertStrict: true,
 		},
 
-		// ---- FAIL TODAY: zero-value temporal opaques ------------------
-		// go-mysql renders the zero TIME as "00:00:00"
-		// (replication/json_binary.go:539) and the zero DATETIME as
-		// "0000-00-00 00:00:00" (json_binary.go:563), but MySQL's render
-		// of the temporal opaques always carries the 6-digit fraction:
-		// "00:00:00.000000" / "0000-00-00 00:00:00.000000". The applier's
-		// text write therefore persists a JSON STRING with the shorter
-		// spelling on the destination — a difference no reparse can heal,
-		// so STRICT and RELAXED both fail. NOT fixed by the
-		// formatMySQLDouble change (separate go-mysql follow-up); the
-		// assertions still state the correct behavior.
+		// ---- PIN-GATED: zero-value temporal opaques -------------------
+		// Green under the current fork pin; red if the pin is dropped for
+		// an unfixed upstream go-mysql. Upstream renders the zero TIME as
+		// "00:00:00" (replication/json_binary.go:539) and the zero
+		// DATETIME as "0000-00-00 00:00:00" (json_binary.go:563), but
+		// MySQL's render of the temporal opaques always carries the
+		// 6-digit fraction: "00:00:00.000000" /
+		// "0000-00-00 00:00:00.000000". An applier text write with the
+		// shorter spelling persists a differing JSON STRING on the
+		// destination — a difference no reparse can heal (STRICT and
+		// RELAXED would both fail). The pinned fork renders the fractions
+		// MySQL-identically.
 		{
 			// Midnight TIME is a valid value under strict sql_mode.
 			name:         "zero_time_opaque",
--- a/pkg/checksum/distributed.go
+++ b/pkg/checksum/distributed.go
@@ -60,9 +60,18 @@ func (c *DistributedChecker) ChecksumChunk(ctx context.Context, chunk *table.Chu
 
 	c.logger.Debug("checksumming chunk", "chunk", chunk.String())
 
-	// Build the checksum query fragment. The same WHERE clause and column list
-	// applies to both sources and targets since all schemas are identical.
-	checksumColumns, _, err := chunk.ColumnMapping.ChecksumExprs()
+	// Build the checksum query fragments. The same WHERE clause applies to
+	// both sources and targets since all schemas are identical, but the
+	// column expressions are side-dependent: JSON columns round-trip through
+	// text on the source and render strictly on the target (see
+	// table.castExpr). Move/sync writes are text-mediated — the applier
+	// writes rendered text that the target re-parses — so the same
+	// text-image contract as the single-server checksum applies here. This
+	// does assume source and target servers parse JSON text identically; if
+	// a future server version changed parse behavior (e.g. fixed the double
+	// misrounding bugs), a cross-version move of affected values would fail
+	// the checksum safely rather than pass silently.
+	sourceChecksumCols, targetChecksumCols, err := chunk.ColumnMapping.ChecksumExprs()
 	if err != nil {
 		return err
 	}
@@ -82,7 +91,7 @@ func (c *DistributedChecker) ChecksumChunk(ctx context.Context, chunk *table.Chu
 		defer c.sourcePools[i].trxPool.Put(srcTrx)
 
 		sourceQuery := fmt.Sprintf("SELECT BIT_XOR(CRC32(CONCAT(%s))) as checksum, count(*) as c FROM %s WHERE %s",
-			checksumColumns,
+			sourceChecksumCols,
 			chunk.Table.QuotedTableName,
 			whereClause,
 		)
@@ -108,7 +117,7 @@ func (c *DistributedChecker) ChecksumChunk(ctx context.Context, chunk *table.Chu
 		defer targetTrxPool.Put(targetTrx)
 
 		targetQuery := fmt.Sprintf("SELECT BIT_XOR(CRC32(CONCAT(%s))) as checksum, count(*) as c FROM %s WHERE %s",
-			checksumColumns,
+			targetChecksumCols,
 			chunk.Table.QuotedTableName,
 			whereClause,
 		)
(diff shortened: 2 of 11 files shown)


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1078, 562faec.

**Verdict:** 1 finding — the DistributedChecker's new target-side asymmetry is the one piece of wiring with zero test coverage (proven by deletion experiment); the text-image contract itself verified sound, both by code trace of every JSON write path and empirically against four MySQL versions.

### Findings

**1. CONFIRMED (untested wiring, by deletion experiment) — reverting the DistributedChecker's target side back to the symmetric source expression passes the entire checksum suite.** [distributed.go:74](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/checksum/distributed.go#L74), [:120](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/checksum/distributed.go#L120)

Experiment run live against MySQL 8.0.45: with the target query switched back to `sourceChecksumCols` (exactly the pre-PR shape), `go test ./pkg/checksum/` is fully green — `distributed_test.go`, `recopier_test.go`, `continuous_test.go`, and the move/datasync suites contain no JSON column anywhere (`grep -ci json` is 0 in all of them). The single-server legs are well pinned (see below), but the distributed leg — the scope note the PR body calls out as newly uniform ("DistributedChecker now uses the target-side expressions") — is pinned only by comments. A future refactor that "simplifies" `ChecksumChunk` back to one shared fragment reintroduces the self-minted, unrepairable checksum failures for cross-server moves of misparse-affected documents, and no CI job notices. Same for the "deliberately read bare" contracts in [distributed.go:219-226](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/checksum/distributed.go#L219-L226) and [recopier.go:111-114](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/checksum/recopier.go#L111-L114) — a well-meaning "add the wrap here too" (which the comments explain would double-apply parse∘render and diverge) is also test-invisible. Ask: one distributed/move e2e with a misparse-affected JSON population, mirroring `TestJSONChecksumFullMantissaTextImage`.

The whole PR stands on one factual claim: every JSON write path stores exactly `p(r(x))` — MySQL's render, MySQL's (buggy) parse, once each. We inventoried every write path at this head. Buffered copier (the default; forced for move/sync): bare SELECT, MySQL renders on the wire, the applier writes a quoted literal the target parses once — contract holds. Single-server repair: `REPLACE INTO … SELECT` with the round-trip cast computes `p(r(x))` server-side and stores the JSON value with no further text hop ([single.go:276-286](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/checksum/single.go#L276-L286)). Distributed repair and `MySQLRecopier`: client SELECT + applier literal = exactly one round trip, so leaving JSON bare there is correct, as the new comments argue. The two paths that deviate: the **unbuffered copier** is server-side byte-faithful ([unbuffered.go:61](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/copier/unbuffered.go#L61)) — disclosed in the PR body, and it converges because the migration checksum runs with `FixDifferences: true, MaxRetries: 3` ([runner.go:816-817](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/migration/runner.go#L816-L817)) *before* the fatal continuous phase; and the **binlog applier** never SELECTs at all — it writes binlog row images rendered to text client-side by go-mysql (`RenderJSONAsMySQLText`, [binlog.go:336](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/change/binlog.go#L336), [gtid.go:368](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/change/gtid.go#L368)), so for that leg the contract's `r` is go-mysql's renderer, which matches MySQL's only via the fork pin (first note). Contract math also checked: render collisions are type-tag-only and the old symmetric form was equally blind to every one of them; doubles render injectively, so a target corrupted to a neighboring ulp still hashes differently. Empirically: both deletion experiments fail exactly as the test comments predict (symmetric revert → text-image test flags pks 1, 2, 4; byte-faithful repair revert → "checksum found differences on every attempt (2/2)"), and all five JSON e2e tests pass on 8.0.28, 8.0.45, 8.4.9, and 9.7.1.

### Notes (not blocking)

- **The binlog leg's contract rests on the TEMP-pinned go-mysql fork, and the PR body glosses over it.** "The binlog applier … do[es] the same [SELECTs rendered text]" is mechanically imprecise: it writes client-rendered row images, and `y = p(r(x))` holds for concurrent DML only because [go.mod:39-40](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/go.mod#L39-L40) pins `morgo/go-mysql` (whose `formatMySQLDouble` was validated byte-for-byte against MySQL's render). The pin pre-exists on main and careless removal turns every CI job red via the unskipped strict fidelity assertions (`huge_integral_doubles`), so the hazard is guarded — but worth one clause in the PR body, because in a pin-removed world the asymmetric checksum escalates for deferred-cutover migrations: the sentinel-wait ContinuousChecker runs with `DivergenceIsFatal: true` and no recopier ([runner.go:1751](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/migration/runner.go#L1751)), and one divergent binlog write during the wait aborts the cutover. (Defensible — that world is genuinely corrupting — but it's a behavior change from the old silently-passing symmetric form.) Related: the fidelity suite's STATUS block still calls the superseded v1.15.1 pin "current" (`subscription_buffered_jsonfidelity_test.go:25-31`) — refresh it; and do *not* unskip `seventeen_digit_mantissa`, it fails under the current pin by design.
- **`--unbuffered` migrations now mass-flag and mass-repair affected JSON chunks** the old checksum passed byte-equal. Converges (attempt 1 repairs, attempt 2 clean) and the PR discloses it, but a JSON-dense unbuffered migration turns its checksum phase into a serialized per-chunk DELETE+REPLACE under `recopyLock`, each bounded by the 10-minute `fixChunkTimeout`. Cheap hardening if it ever bites: wrap JSON in `textRoundTripCast` in the unbuffered SELECT so the copier writes the contract image directly.
- **The comment's CRC literals don't reproduce.** [json_test.go:74](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/checksum/json_test.go#L74) quotes "790255375 vs 3468554960 on 8.0.45"; re-running the symmetric revert on 8.0.45 against this exact 7-row population yields 2311333401 vs 3524052335. The substance is right (verified — it does fail), but the literals look like they were measured on an earlier draft population. Drop the numbers or re-measure.
- **The regression teeth are version-dependent, and the comment is already time-stale.** On MySQL 9.7.1 the probe values are parse/render fixed points (verified by direct SQL), so both new e2e tests pass there under *either* cast form — the 8.0.x/8.4 matrix jobs carry all the regression coverage. Fine today; worth a line in the [castExpr comment](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/table/utils.go#L120) ("unfixed through 8.0.45") acknowledging 9.x behaves differently, so nobody trims the older matrix jobs thinking the coverage is redundant.
- **The e2e tests hand-write the contract SQL** (three copies of `CAST(CAST(j AS CHAR CHARACTER SET utf8mb4) AS JSON)`) because [`textRoundTripCast`](https://github.com/block/spirit/blob/562faecb7d0a978d8d03c71fa7b84d8bcbdcd7fb/pkg/table/utils.go#L152) is unexported. The pkg/table unit pins mitigate drift, but exporting the helper (or a test accessor) would couple them.
- `RepairExprs`'s error branch is effectively unreachable via its only caller (ChecksumExprs validates the same type-map keys earlier in the same flow) — correct and tested, just informational.

### Verified correct

- Complete caller sweep: all four `ChecksumExprs` consumers route source exprs to the source table and target exprs to the target (single ChecksumChunk + inspectDifferences, distributed, and `continuous.go:1065-1076`, which already consumed both values and needed no change); no production caller of `castExpr`/`wrapCastType*` exists outside ChecksumExprs; no stray hand-built symmetric casts remain.
- NULL handling is symmetric and untouched (`IFNULL(cast,'')` + `ISNULL(col)` per side, `'#'` separator unchanged); row 6 of the new e2e exercises it.
- Move flow safety: move chunkers build `NewColumnMapping(t, nil, nil)` which coalesces the nil target to the source TableInfo (SetInfo run), and moves validate column-set equality per target — so `targetChecksumCols` are always valid on the target servers and type lookups can't miss.
- Repair convergence mechanism traced: `Run` retries full passes; with byte-faithful repair reverted, `MaxRetries=2` exhausts with "differences on every attempt (2/2)" — the new convergence test has exactly zero slack, as its comment claims. `chunker.Reset()` between phases is the interface-blessed pattern `Run` itself uses; the `differencesFound==0` assertion is belt-and-braces on top of runChecksum's own mismatch error.
- CI matrix enumerated: 8 jobs, all MySQL (8.0.28/8.0.42/8.0.45 ×4 variants/8.4/9.7) — no MariaDB, no 5.7, so `CAST(col AS json)` validity is a non-issue and no new server-version requirement is introduced (the old expression already had an outer JSON cast).
- Comment conventions match repo norms (issue references and past-form narratives are established in this codebase); `chkjsonasym*` table naming and raw-RunSQL setup match the file's four pre-existing tests; no DCO/agent-disclosure norm violated.
- `go build ./...`, `go vet ./...` clean at head; pure unit tests pass; CI fully green (8 builds, lint, govulncheck, Semgrep, zizmor).

Kiran01bm (approved):

morgo:
🤖 Reply from Morgan's AI agent — all points addressed in 985b26cd:

- **Distributed e2e (the finding):** added `TestDistributedJSONChecksumTextImage` — same misparse-affected population as `TestJSONChecksumFullMantissaTextImage`, target table in a separate database holding the one-text-round-trip image, checksummed through the `DistributedChecker` (selected via `config.Applier`). Re-ran your deletion experiment to confirm it's a real pin: with `targetChecksumCols` reverted to `sourceChecksumCols` in `ChecksumChunk`, the new test fails with a checksum mismatch while the rest of the suite stays green; restored, it passes first-attempt with zero differences.
- **Stale CRC literals:** replaced the `790255375 vs 3468554960` comment (and the same claim in the PR body) with what actually reproduces on 8.0.45 — under symmetric casts the misparse-affected rows 1, 2 and 4 flag on every attempt.
- **Fork-pin dependency:** PR body now has a "Binlog applier leg" scope note spelling out the chain — binlog-applied rows get their rendered text from the pinned go-mysql fork, the jsonfidelity tests gate the pin, and in a pin-dropped world a divergent binlog write is a repairable checksum failure in a migration but aborts a deferred cutover (ContinuousChecker treats divergence as fatal, no recopier).
- **Fidelity STATUS block:** rewritten for the current pin (v1.16.1-0.20260723231236); the two "FAIL TODAY" sections are now "PIN-GATED", framed as the upstream behavior they guard against. `seventeen_digit_mantissa` stays skipped by design, per your note.
- **castExpr comment:** now notes MySQL 9.x parses the probe values correctly, so the 8.0.x/8.4 CI legs carry the parse-fidelity regression coverage — don't trim them as redundant.

`pkg/checksum`, `pkg/table`, `pkg/change` suites, gofmt/vet/golangci-lint all green locally. Enabling auto-merge.

PR #1081, opened 2026-07-24
applier: add Stats() pipeline snapshot with queue-wait/write-time percentiles

Kiran01bm:
## Summary
Adds an applier-side `Stats()` snapshot so a write-limited copy pipeline is diagnosable at runtime. Today the copier's dynamic chunker feedback measures end-to-end chunk time (read + queue wait + write), so a saturated applier presents as a read-side/chunker problem — the chunker shrinks chunks while the true bottleneck is invisible.

## What
- New `Stats()` on the `Applier` interface, implemented by `SingleTargetApplier` and `ShardedApplier`, reporting queue depth/capacity, pending work, active write workers, and rolling p50/p90 queue-wait vs write-time (128-entry timing ring, matching the default buffer capacity).
- Queue-wait is stamped before the buffered send, deliberately including send-side backpressure; write time is measured dequeue-to-completion, so the two cleanly separate "waiting for a worker" from "doing the write".
- Sharded stats aggregate capacity/depth/workers across shards with one shared timing ring.

## Why
During a load-test migration on a large table, throughput collapsed ~4x below target with the chunker pinned near its floor. Post-analysis showed the apparent source-side bottleneck was most plausibly the applier write path (write-threads at default 4), but nothing in the status output could confirm it: worker count is now reported (#953) but queue occupancy and wait-vs-write split are not. With this snapshot, a write-limited pipeline shows up unambiguously as queue pegged at capacity and queue-wait ≫ write time.

A follow-up PR will surface these fields in the migration/move status lines and metrics.

```
before — chunker feedback blends read + queue-wait + write

┌─────────┐   end-to-end chunk latency (read + queue + write, blended)
│ chunker │◀─────────────────────────────────────────────────────────┐
└────┬────┘                                                          │
     │ shrinks chunks                                                │
     ▼                                                               │
┌────────┐     ┌───────────────┐     ┌───────────────┐               │
│ reader │────▶│ applier queue │────▶│ write workers │───────────────┘
└────────┘     └───────────────┘     └───────────────┘
                saturation here is invisible — presents as a
                chunker/read-side problem
```

```
after — applier exposes its own pipeline snapshot

┌────────┐     ┌───────────────┐     ┌───────────────┐
│ reader │────▶│ applier queue │────▶│ write workers │
└────────┘     └───────┬───────┘     └───────┬───────┘
                       │                     │
                       ▼                     ▼
               ┌─────────────────────────────────────────┐
               │ Stats(): queue depth/capacity, pending, │
               │ active workers, queue-wait p50/p90,     │
               │ write-time p50/p90                      │
               └─────────────────────────────────────────┘
                write-limited pipeline is unambiguous:
                queue pegged at capacity, queue-wait ≫ write-time
```

### DIFF
--- a/pkg/applier/README.md
+++ b/pkg/applier/README.md
@@ -119,6 +119,10 @@ The applier tracks all pending work internally and invokes the callback only whe
 
 This allows the copier to continue reading and queuing more work without blocking, while still maintaining correctness by only advancing the watermark after writes complete.
 
+### Pipeline observability (`Stats()`)
+
+Both appliers expose a point-in-time `Stats()` snapshot (see `stats.go`): queue depth/capacity, pending work, live write workers, and rolling p50/p90 of per-chunklet **queue wait** (time between `Apply()` offering a chunklet to the buffer and a worker dequeueing it, including send-side backpressure) and **write time**. This exists because the copier's chunk feedback is end-to-end — read + queue wait + write — so a saturated write side otherwise presents as a read/chunker problem. A queue pegged at capacity with queue-wait far above write time means the pipeline is write-limited; a near-empty queue means it is read-limited.
+
 
 ## Implementation Details
 
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -67,6 +67,12 @@ type Applier interface {
 	// For the subscription: callback will update binlog coordinates
 	Apply(ctx context.Context, chunk *table.Chunk, rows [][]any, callback ApplyCallback) error
 
+	// Stats returns a point-in-time snapshot of the write pipeline: queue
+	// occupancy, pending work, live workers, and rolling queue-wait /
+	// write-time percentiles. Safe to call concurrently with Apply; values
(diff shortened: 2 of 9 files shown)


Kiran01bm:
**Heads-up: this is PR 1 of 2.** A follow-up PR will surface these applier stats in the `migrate`/`move` status lines and as metrics gauges (queue depth/capacity, pending work, active workers, queue-wait p50/p90, write-time p50/p90).

**Why the split:** this PR is instrumentation-only — it adds `Stats()` to the `Applier` interface and its implementations with zero user-visible behavior change, so it can be reviewed purely on the correctness of the measurement (where queue-wait is stamped, what the timing ring captures, mutex/race safety). The follow-up touches the presentation surface (status formatting, metric names/cardinality), which tends to invite iteration; splitting keeps that churn from blocking the core instrumentation and lets us pick the instrumentation up early in our qaload write-threads testing.

morgo (approved):

PR #1087, opened 2026-07-27
checksum: pace the checksum phase against the throttler, and let it scale

morgo:
Follows #1083 and #1086, which brought read-worker scaling into autoscaling. The checksum was the last major phase of a schema change that ignored throttling entirely: it read at a fixed thread count regardless of replica lag or server load, and could not be adjusted once a pass started.

This makes it a participating phase. It is PR 1 of 3 — see [Out of scope](#out-of-scope).

## What it does

Three capabilities. Only the first is unconditional, so the table is keyed on the opt-in rather than on the server — `AutoscaleConfig.Enabled` (`--enable-experimental-autoscaling`) is what constructs the scaler, and *both* shedding and growth live in the scaler:

|                                       | hard stop | shed on backlog | grow |
| ------------------------------------- | --------- | --------------- | ---- |
| scaling disabled (default), any server | yes       | no              | no   |
| scaling enabled, stock MySQL          | yes       | yes             | no (recovers to configured count only) |
| scaling enabled, Aurora               | yes       | yes             | yes (utilization law) |

**1. A hard stop, everywhere.** The dispatch loop calls `Throttler.BlockWait` before each chunk, so a checksum now pauses under lag or load regardless of the flag. Chunks already in flight are never interrupted — an abandoned chunk is wasted I/O that has to be redone from the same watermark, so the checksum stops *dispatching* rather than cancelling. Runners wire this through a new `ThrottleAware` capability, because they build the checker before the throttler is open.

**2. Shedding on change-feed backlog, on any server — but only under the flag.** The feed flushes concurrently with the checksum and its backlog gates cut-over, so if the feed is losing ground our reads are winning a race against writes that actually have to finish. Unlike utilization this signal needs nothing from the throttler, so it is *available* everywhere; it is *active* only where the scaler exists. It is the only shedding lever on stock MySQL, since the continuous utilization signal is Aurora-only. Without the flag a checksum has the hard stop and nothing else, and never moves its own worker count in either direction.

**3. Growth on the utilization zone law**, shared with the copier via a new `pkg/autoscale`. Aurora-only, and likewise under the flag.

## The constraint that shaped the design

The `REPEATABLE READ` transaction pool **cannot grow** once the table lock is released. Every transaction takes its snapshot under that lock, so they all see one point in time; a transaction started later would read a newer snapshot and could compare a chunk against changes its siblings cannot see.

So the pool is provisioned at the autoscale ceiling up front, whether or not scaling is enabled, and a resizable `autoscale.Limiter` gates the live worker count — `errgroup.SetLimit` may not be resized while goroutines are active. Over-provisioning costs one connection per idle transaction and **no** extra history retention, since every read view pins from the same instant.

Lowering the limit never interrupts in-flight work; the reduction is absorbed by subsequent releases.

## Measuring the backlog correctly

Worth calling out because the obvious formulations are wrong, and I shipped one of them first (fixed over the later commits — the description below is the shipped behaviour).

The pending-change count is a sawtooth: it climbs monotonically between flushes and drops when one lands. At the production numbers — 5s control tick, 30s `DefaultFlushInterval` — **the rising edge alone is six samples long**, so "the backlog is large and has been rising for N ticks" is the normal state of any table busy enough to matter. It carries no information about whether the feed is coping.

Window minima don't recover it either, which is the subtler trap. A poll lands some offset φ after the flush and therefore reads `residual + writeRate·φ`; because the flush interval is an exact multiple of the tick, φ is fixed for the whole pass by the arbitrary phase between two independent tickers. On a busy table the sampling term can clear the threshold on its own, and a *rising write rate* on a fully-draining feed produces rising apparent minima indistinguishable from a feed falling behind.

What a healthy feed guarantees is that each flush drains it to near zero, and only the feed can say what a flush left behind. So the quantity is now **the post-flush residual, recorded by the feed at flush completion** and exposed as `change.Source.FlushResidual() (residual, flushCount)` — not polled. That removes the write rate from the signal entirely. Successive residuals are compared across distinct flushes (keyed on the flush counter, so a residual is evaluated once), with hysteresis of `csBacklogHysteresisFlushes` in **both** directions. The exit condition matters as much as the entry one: shedding is one step per flush while growth is one step per two ticks, so a single favourable flush clearing the verdict would let the grows outpace the sheds and the controller would drift up while the feed fell further behind. While a verdict holds it suppresses growth as well as driving shedding.

Because the signal is keyed on the flush counter, silence needs handling rather than latching. `binlogClient.flush` returns early on error *before* recording a residual, and the periodic flusher logs and carries on rather than failing the migration — so the counter can freeze while the backlog grows without bound (a flush that merely takes minutes freezes it too). After `csStaleFlushTicks` ticks with no new flush the scaler stops trusting the standing verdict and freezes **increases** — growth and recovery alike — logging once per episode and again when flushes resume. It deliberately does not shed on it: a frozen counter says the signal stopped, not which way it was heading, and shedding on no evidence would walk a healthy pass down to one worker. This also covers `DistributedChecker`, whose aggregate flush counter is the minimum across feeds, so one stuck feed freezes the signal for all of them instead of silently blinding it.

The unit tests drive the scaler at the real tick:flush ratio rather than with hand-set backlog values, which is what let the original mistake through.

## Incidental fix: a latent false PASS

If a pass stopped dispatching before the chunker was drained — cancellation landing while parked in `BlockWait` with no chunk in flight — `g.Wait()` returned nil, `runChecksum` returned nil, and the runner logged **"checksum passed"** for a partially verified table. Both checkers now fail unless the chunker is fully read, ordered after the existing yield check so a legitimate yield is still classified as a yield. `TestChecksumCancelledWhileThrottledNeverReportsPass` guards it, and is non-vacuous — the same scenario returned nil before the fix.

This was reachable before this PR via cutover-time cancellation; adding `BlockWait` to the dispatch loop widened the window enough that a test caught it.

## Groundwork for chunk sizing

Each pass now logs a `checksum chunk size distribution` line: chunk count, duration p50/p90/max, rows p50/max, and how many chunks hit `table.MaxDynamicRowSize`.

The row-capped count is the useful one. The checksum aggregates server-side and returns one row per chunk, so its chunks are far cheaper than the copier's — but if most of them are pinned at the row ceiling well inside the time target, then raising `--target-chunk-time` alone would change nothing. Better to decide that from data than by guessing. (It was: PR 3 in the stack found 10 of 12 chunks row-capped on the benchmark table.)

## Docs

`docs/migrate.md` is updated in the same sweep: the connection-pool formula (now `threads + write-threads + control-plane + checksum-off-pool`, with what each term covers), and a **Checksum scaling** subsection under `--enable-experimental-autoscaling` separating the three behaviours above plus the frozen-signal freeze.

## Out of scope

Deliberately not in this PR, in this order:

1. **Repair-path chunk subdivision** — #1089. Prerequisite for the next one: without it, one mismatch in a large chunk means recopying the whole thing.
2. **Actually growing checksum chunk sizes**, driven by the measurements above — #1090. 500ms chunks don't make much sense for a phase that returns one aggregated row per chunk and would benefit from read-ahead; ~10s is the target worth aiming at.

Throttle-triggered yield was considered and dropped: yielding re-acquires a table lock and flushes under it, which is the wrong move during server overload, and the 24h default yield timeout suggests long snapshot holds are already tolerated.

`ContinuousChecker` is not covered by any of this — it manages its own pacing through `MinPassInterval` and its retry queue, and takes no table lock or snapshot pool.

## Deployment note

`checksum_threads` is a **new metric** and needs allowlisting in the tern sink (`apps/tern/metrics/{spirit_sink.go,metrics.go}`), or it will Sentry-log as an unknown metric. Non-fatal.

## Testing

- New unit tests for the zone law and `Limiter` invariants in `pkg/autoscale`, including a 32-goroutine test asserting peak concurrency never exceeds the maximum limit ever set while resizing concurrently.
- New scaler tests covering growth/shed/hold/halve, bounds, the backlog veto with and without a continuous signal, a healthy feed never vetoing, a flat-but-large residual never vetoing, recovery on both paths, panic-zone precedence, and the frozen-flush-counter freeze (freezes growth, freezes recovery, does not shed, and a never-flushed feed is not penalised).
- `chunkstats` tests now pin the nearest-rank percentile semantics, including small-`n` cases, so the logged distribution can't quietly change meaning.
- New integration tests against real MySQL: the throttler is consulted, throttling blocks dispatch and completes on release, a throttled pass is cancellable, mid-pass scale-up succeeds (proving the over-provisioned pool works), and the false-PASS regression.
- `pkg/checksum`, `pkg/autoscale` and `pkg/migration` suites pass, including under `-race`. `golangci-lint` clean.


morgo:
🤖 Pushed 5da25049, which changes the design of the backlog signal — flagging it since it touches `pkg/change`'s `Source` interface, which is more than the original PR description implied.

**The signal was wrong twice, in the same way.** Both attempts inferred the post-flush residual by polling `GetDeltaLen`, and both spuriously shed workers on a completely healthy feed:

1. First: *"the count is large and has risen for 2 consecutive ticks."* At a 5s tick and 30s flush the rising edge alone is six samples long, so this is the normal state of any busy table.
2. Then: *"take the minimum over a window spanning two flush intervals,"* on the theory that such a window contains a flush trough. It does — but the poll doesn't land on it. A sample taken φ after a flush reads `residual + writeRate·φ`, and since 30s is an exact multiple of 5s, **φ never rotates**: it's fixed for the whole pass by the arbitrary phase between two independent tickers. At 4000 keys/s with φ=3s, a fully-draining feed reports a 12,000 residual — over threshold, every flush. Worse, a *rising write rate* makes the apparent residual rise on a feed that drains completely, which reproduced the Aurora ratchet-to-one-worker through a different door.

Both slipped through because the test fixture sampled the backlog exactly at the bottom of each cycle — i.e. modelled a poller synchronised to the flush, the one case that never happens in production.

**So the residual is now reported by the feed instead of inferred.** `change.Source` gains:

```go
FlushResidual() (residual, flushes int)
```

recorded at the end of `flush()` in both clients. The write rate leaves the signal entirely, and the flush counter lets the controller compare residuals once per flush rather than once per poll. The window, its phase sensitivity, and a `csTick`-derived package var all disappear.

Also in this push: the verdict now has **hysteresis in both directions**. Shedding is one step per flush (30s) while growth is one per two ticks (10s), so a single favourable flush clearing the verdict let grows outpace sheds — a feed losing ground on balance but not monotonically would net-grow. And the utilization panic zone now outranks the backlog veto, since both shed but panic sheds multiplicatively; evaluating the veto first meant a tick inside its cooldown consumed the panic response.

Happy to split the `pkg/change` change out into its own PR if you'd rather review it separately — it's self-contained (interface method + two implementations + a test).

Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1087, 496deca.

**Verdict:** 1 finding — the code held up under a dedicated control-law review and a hand-walk of the new Limiter, but the PR description no longer describes the shipped design in two load-bearing places. Six non-blocking notes.

Head-movement note: the review ran against `2b2f83d8` and was re-based onto `496decad` after the author pushed a merge that **removes the chunk-subdivision feature from this PR** (`subdivide.go`, `table/chunk_split.go`, both test files, and the repair-flow hooks — ~1,200 lines). The delta was audited: apart from that removal, only a main-side `runner.go` hunk (from the #1088 series) changed; the scaler, `pkg/change`, `pkg/autoscale`, copier, and metrics files are byte-identical between the two heads, so all analysis carries over. Subdivision-specific findings (a `fixChunkTimeout` budget regression, split-boundary edge cases) are moot at this head and were dropped; the repair path is back to its pre-PR shape (budget opens after `recopyLock`, verified). CI is fully green on `496decad` (all 8 MySQL matrix jobs); the checksum/table suites also passed locally under `-race` at the prior head, whose checksum code differs only by the removal.

### Findings

- **The PR description describes a superseded control law and overstates where shedding applies — for a control-law PR, the body is what reviewers review, and both errors mislead.** (1) The "Measuring the backlog correctly" section says the controller "tracks the **trough** of a window spanning two flush intervals and compares successive troughs." The shipped code has no windows and no troughs: it records the residual at each flush completion, keyed on a monotonic flush counter, and compares successive flush-keyed residuals ([autoscaler.go#L284-L296](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/autoscaler.go#L284-L296)) — and the PR's own in-code comment explicitly refutes window-minima polling ([autoscaler.go#L66](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/autoscaler.go#L66)). The paragraph describes the intermediate revision that commit f9113f03 replaced. (2) "Shedding on change-feed backlog, **everywhere**" (and the capability table's stock-MySQL "shed on backlog: yes") — the scaler, shedding included, is constructed only under `if c.autoscale` ([single.go#L648-L656](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/single.go#L648-L656)), and that flows from `EnableExperimentalAutoscaling` ([runner.go#L714](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/migration/runner.go#L714), [#L869](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/migration/runner.go#L869)). Without the opt-in, stock MySQL gets the hard stop only. What's true: *with* the flag on stock MySQL, shed works and growth recovers only to the configured count. Rewrite the two sections (the "Out of scope" section is accurate again after the subdivision removal, and "PR 1 of 3" holds).

The design constraint the whole PR bends around: the `REPEATABLE READ` snapshot pool **cannot grow** after the table lock is released, so live concurrency is gated by a resizable limiter over a ceiling-provisioned pool — replacing `errgroup.SetLimit`. Verified end to end:

- **Pool ≥ limiter, always.** The transaction pool is created at `maxConcurrency = max(Autoscale.MaxThreads, concurrency)` ([checksum.go#L226](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/checksum.go#L226), [single.go#L465](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/single.go#L465)) at the same under-lock call site as before, and the scaler is constructed with the same `c.maxConcurrency` as its ceiling — the limiter can never admit more workers than there are pinned snapshots. No path mints a transaction after the lock: repair reads reuse the mismatching chunk's own pooled transaction, and recopies deliberately run on `c.db` (current data, off-pool, covered by the new `checksumOffPoolConns` reserve).
- **The Limiter state machine survives a hand-walk** ([autoscale.go#L110-L204](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/autoscale/autoscale.go#L110-L204)): the generation channel is captured under the same mutex that guards the admit check, so a release between capture and select closes the already-captured generation — no lost wakeup; raises broadcast and wake blocked acquirers; a cancelled waiter has not yet incremented `inFlight`, so nothing leaks; `inFlight++` happens only under the lock while `< limit`, so peak concurrency can never exceed the highest limit ever set (also pinned by the PR's 32-goroutine test). The 4→1-with-4-in-flight→releases→8 sequence converges with capacity exactly equal to the set limit — no drift.
- **Dispatch can't lose a chunk.** `BlockWait` runs on the dispatcher *before* a permit is taken (a throttled checksum holds no capacity), and the chunk is claimed by `chunker.Next()` *inside* the worker after a successful `Acquire` ([single.go#L658-L672](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/single.go#L658-L672)) — a cancellation parked in `BlockWait`/`Acquire` breaks the loop with no chunk claimed, and the new drained guard ([single.go#L723-L728](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/single.go#L723-L728)), ordered after the yield classification at [#L703](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/single.go#L703), converts that into a failure rather than a false PASS. An independent verifier confirmed the guard is genuine and correctly ordered, with one framing nuance: the wide-open version of that hazard is *introduced by this PR's own dispatch changes* and guarded in the same PR (pre-PR it existed only as a narrow cutover-cancellation race), which the PR body fairly acknowledges.

### Notes (not blocking)

- **A persistently failing flush freezes the veto and the controller fails open — in exactly the losing-ground regime the law exists for (CONFIRMED, low).** `flush()` returns on a subscription error before `recordFlushResidual()` ([binlog.go#L1032](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/change/binlog.go#L1032)), the periodic loop logs and keeps ticking ([#L1147](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/change/binlog.go#L1147)), and `observeBacklog` early-returns the latched verdict when the counter hasn't advanced ([autoscaler.go#L284-L286](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/autoscaler.go#L284-L286)) — default false, so Aurora growth continues against an unflushable backlog; in the distributed checker one stuck feed blinds the veto for all feeds via the `min()` aggregation ([distributed.go#L151](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/distributed.go#L151)). Independent verification downgraded this to low: persistent flush failure dooms the run regardless (cutover's flush-under-lock propagates the same error), transient windows self-heal in ~60s, and a verdict already latched true stays latched. Cheap hardening if wanted: treat a counter frozen for >N flush intervals as losing (fail closed), or at least log it from the scaler.
- **The flush residual is contaminated by during-flush arrivals only in multi-feed flows (REFUTED for the common case).** The verifier traced the delta map's single writer through the subscription mutex, which is held for the *entire* flush ([subscription_buffered.go#L471-L473](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/change/subscription_buffered.go#L471-L473)) — so for single-subscription `migrate` the contamination window is microseconds and a false veto is essentially impossible. With N>1 subscriptions (distributed/move-shaped flows), earlier-flushed subscriptions re-accumulate while later ones flush and the sum inflates by ~writeRate × the later flushes' duration; worst case is a spurious single-worker shed that recovers in two clean flushes. Fixable later by summing per-subscription residuals captured at each drain point instead of one post-loop `GetDeltaLen()` ([binlog.go#L1043](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/change/binlog.go#L1043)).
- **Hysteresis edge cases, deliberate but worth owning:** a strict period-2 residual alternation (rise, dip, rise, dip…) never produces two consecutive rises, so a feed climbing on average never latches the losing verdict ([autoscaler.go#L290-L296](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/autoscaler.go#L290-L296)) — a measure-zero trajectory under real jitter, and symmetric (the same pattern can't *clear* a latched verdict either); and a large-but-flat residual plateau is classified healthy by design (test-pinned), meaning "pinned at max depth, holding ground" never sheds even though cutover's backlog gate is unreachable until traffic drops.
- **Commit `b9bedee0` is a raw Copilot-Autofix artifact: message "Potential fix for pull request finding".** The change itself is correct — the old `percentileIndex` returned the max for p90 at small n; the fix implements nearest-rank exactly matching the applier's percentiles, which the function's own doc comment already claimed ([chunkstats.go#L88](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/chunkstats.go#L88)) — but the message violates the repo's `pkg: summary` convention and describes nothing. Squash it into f179bc0a or reword (e.g. `checksum: use nearest-rank percentiles in chunk stats`); the changed semantics are also not pinned by `chunkstats_test.go` (asserts only max/count fields).
- **The subdivision feature was removed via the merge commit itself** — `496decad` is nominally "Merge branch 'main'" but also deletes ~1,200 lines the branch added, which is easy for a reviewer to miss (a merge diff is rarely read). Since the branch history now contains a full feature added-then-removed plus the Copilot commit, consider rebasing the branch clean (or squash-merging) so main's history doesn't carry a phantom feature; and the peeled-out subdivision work deserves its follow-up PR while it's fresh.
- **User-facing docs don't mention the new behavior.** `docs/migrate.md` still documents the pool as `threads + write-threads + 1` ([#L401](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/docs/migrate.md#L401) — now `+ controlPlaneConns() + checksumOffPoolConns`), and the `--enable-experimental-autoscaling` description ([#L405](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/docs/migrate.md#L405)) says "read- and write-thread scaling" with no mention that the checksum now pauses under throttle for everyone or scales under the flag. Same docs-lag pattern flagged on #1085; worth one sweep before the stack ships.

### Verified correct

- **Copier scope is provably zero-change:** its constants became aliases (`acLowWatermark = autoscale.LowWatermark`, [copier/autoscaler.go#L58](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/copier/autoscaler.go#L58) etc.) with identical values (0.4 / 0.7 / 1.0 / 2 ticks / 5s), so the shared law cannot drift and the copier's behavior is untouched.
- **The scaler's truth table was pinned by the control-law finder and held under verification:** panic-halve outranks the backlog verdict; shed is at most one worker per flush with a floor of 1; growth (both the Aurora Grow zone toward max and the recovery path toward the configured start) is suppressed while the verdict is latched; verdict changes need two consecutive agreeing flushes in either direction (no flap); startup is baseline-only (first flush never vetoes); recovery after a healed feed is ~60s + one step per 2 ticks. All test-pinned at the real 6:1 tick:flush ratio.
- **Runner wiring is complete for migrate and deliberate elsewhere:** `setThrottlerOnPhases` ([runner.go#L946](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/migration/runner.go#L946)) hands the real multi-throttler to copier + checker on the production path; the always-throttled test mock stays copier-only with a documented rationale ([#L974](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/migration/runner.go#L974)); `move` passes an explicit `Noop` to its checker consistent with its copier (move is unpaced by design, pre-existing); a checker lacking the capability degrades to unthrottled with a debug log, never a crash ([checksum.go#L229](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/checksum/checksum.go#L229)).
- Connection accounting: the `checksumOffPoolConns = 2` reserve is real (repair statements and chunker prefetch both run off-pool and serialized — one connection each), and the checksum's ceiling reuses the copier's `maxRead` allocation since the phases don't overlap; both pool-sizing sites updated consistently.
- `checksum_threads` gauge named consistently with `read_threads`/`write_threads` ([metrics.go#L31](https://github.com/block/spirit/blob/496decadea5a191b436a59c9e20c66ff90050c39/pkg/metrics/metrics.go#L31)); the PR's deployment note about tern-sink allowlisting is accurate and downstream.
- `ContinuousChecker` untouched as claimed — its only diff is a test fake implementing the widened `change.Source` interface; the checksum status line gains a throttled/scaled suffix mirroring the copier's.
- Yield semantics preserved: `ErrYieldTimeout` classification runs before the new drained guard, including the subtle all-dispatched-but-cancelled-mid-final-chunk case.
- CI fully green on `496decad` (8-job MySQL matrix, lint, govulncheck); `pkg/checksum` and `pkg/table` suites pass locally under `-race` at the analyzed head; `go build`/`go vet` clean.

Kiran01bm (approved):

morgo:
🤖 Thanks — addressed. Pushed 63d22164 and fad9bafc; `checksum-autoscale` is re-mirrored on upstream so #1089's diff stays honest.

**Finding 1 — fixed, and it was worse than reported.** Both errors were also in the shipped `pkg/checksum/README.md`, which matters more than the body since it outlives the PR. Fixed in both places:

- The backlog section now describes what ships: the post-flush residual, recorded by the feed at flush completion and read via `change.Source.FlushResidual()`, compared across flush-keyed samples with two-sided hysteresis. The window/trough language is gone. I kept the *reason* window minima fail (the fixed sampling offset φ, so a rising write rate on a healthy feed mimics a losing one) because that's the part worth not re-learning.
- The capability table is now keyed on the opt-in rather than the server, since `AutoscaleConfig.Enabled` is what constructs the scaler and both shedding and growth live there. Three rows: scaling off (hard stop only, any server) / on + stock MySQL / on + Aurora.

**Note 1 (fail-open) — fixed rather than logged.** You're right that a persistent failure dooms the run anyway, but the freeze also happens when a single flush merely takes minutes, which is survivable and is exactly the losing-ground regime. After `csStaleFlushTicks` (12 ticks = two flush intervals) with no new flush the scaler stops trusting the standing verdict and freezes **increases** — growth *and* recovery — logging once per episode plus a line when flushes resume. It deliberately does not fail closed to *losing*: a frozen counter says the signal stopped, not which way it was heading, and shedding on no evidence walks a healthy pass down to one worker. Staleness is only counted after a first flush is observed, so a pass whose feed hasn't flushed yet isn't penalised. This also covers the distributed `min()` case you flagged — one stuck feed now freezes the signal visibly instead of blinding it. Four new tests.

**Note 3 (hysteresis edges) — owned in comments.** The period-2 alternation and the large-but-flat plateau are now written down as deliberate, including that the plateau means cut-over's gate is out of reach until traffic drops, which worker count can't fix.

**Note 2 (multi-feed contamination) — recorded, not fixed.** Your refutation for single-feed `migrate` matches what I read. The N>1 inflation and the per-subscription-residual fix are now in the `DistributedChecker.flushResidual` doc comment so the next person meets it there.

**Note 4 — test added; commit message left to squash.** `chunkstats_test.go` now pins nearest-rank explicitly, including the small-`n` cases where definitions disagree (`percentileIndex` table test plus exact p50/p90 assertions on the summary line). I'd rather not reword `b9bedee0` in place: that branch is #1089's base on upstream, so rewriting it force-pushes three branches and re-mirrors two upstream bases mid-review. Squash-merging #1087 drops the message entirely, which gets the same outcome for main's history — happy to reword if you'd prefer to merge unsquashed.

**Note 6 (docs) — swept.** `docs/migrate.md` now documents the real pool formula with what each headroom term covers, and has a **Checksum scaling** subsection under `--enable-experimental-autoscaling` separating the three behaviours (hard stop unconditional, shedding under the flag on any server, growth under the flag on Aurora only) plus the frozen-signal freeze.

**Note 5 — the premise is a misreading; no rebase warranted on that ground.** `496decad` is a clean merge of main: `git log --format='%h %p' -1 496decad` → parents `b9bedee0 ce04c8df`, and `git diff --stat ce04c8df..496decad` is purely main's #1088 series. The subdivision files were never in this branch's reachable history — `git ls-tree 496decad pkg/checksum/subdivide.go pkg/table/chunk_split.go` is empty and `git log 496decad -- <those paths>` returns nothing. What the re-base saw was my earlier force-push splitting the subdivision work out into its own PR, not a deletion inside the merge. So there is no phantom feature to rebase away. The follow-ups it asks for are already open and stacked: #1089 (repair-path subdivision) → #1090 (grow checksum chunks toward ~10s; it found 10/12 chunks row-capped, so the row cap — not the time target — is what binds).

Suites green: `pkg/checksum` and `pkg/migration` pass, `pkg/checksum` also under `-race`. `golangci-lint` clean.

PR #1089, opened 2026-07-27
checksum: narrow a mismatching chunk before recopying it

morgo:
> **Stacked on #1087** — this targets the `checksum-autoscale` branch, so the diff shows only this PR's own change. Retarget to `main` once #1087 merges.

This is PR 2 of 3 in the checksum-autoscaling series. PR 1 (#1087) made the checksum phase throttler-aware and scalable, and started measuring chunk read times. PR 3 will use that measurement to grow checksum chunks toward ~10s of read time. This PR is the prerequisite for that: it makes a mismatch repair proportional to the mismatch rather than to the chunk.

## Why

A checksum chunk is sized by how long it takes to *read*. That says nothing about how long it takes to rewrite — and on a mismatch we `DELETE` and re-write the chunk's entire key range. `replaceChunk`'s own comment has flagged this for a while:

> Note that the chunk is dynamically sized based on the target-time that it took to *read* data in the checksum. This could be substantially longer than the time that it takes to copy the data. Maybe in future we could consider splitting the chunk here, but this is expected to be a very rare situation, so a small stall from an XL sized chunk is considered acceptable.

Once chunks are sized for read-ahead rather than for latency, "a small stall from an XL sized chunk" stops being small: fixing a handful of rows becomes a large write burst, a long lock hold, and a wide window in which the target range is deleted but not yet rewritten (the `DELETE` and `REPLACE` are deliberately separate transactions, to avoid a documented deadlock on a unique secondary index).

## What

On mismatch, cut the range into 8, re-checksum each piece inside the same snapshot that observed the mismatch, and repair only the pieces that differ — recursing twice, or until a piece is under 2000 rows. On a 19,600-row chunk with one bad row that rewrites **306 rows instead of 19,600**.

New: `table.Chunk.Split` (the range arithmetic, in `pkg/table` because `datumTp` and `expandRowConstructorComparison` are unexported and both checkers need it) and `checksum.narrowRepair` (the depth-bounded driver). Both checkers gained a `checksumRange` helper so the whole-chunk and sub-range comparisons are the same code.

## Why skipping a clean piece is safe

`BIT_XOR` is associative and the row counts are additive, so over an exact partition read in one snapshot the pieces must reproduce the whole-range result — meaning a mismatching range must contain a mismatching piece.

Both halves of that matter, and both are enforced:

- **Same snapshot.** A piece read from a different snapshot could look clean because a concurrent write happened to fix it, and the repair would be skipped on the strength of an observation that never applied to the chunk being fixed. There is a test that fails if either checker is moved off the snapshot transaction.
- **Exact partition.** `Split` cannot promise this for every key type, so it is *checked* rather than assumed: the pieces' row counts must sum to the parent's on both sides. MySQL orders `ENUM` by declaration ordinal but compares it against a string literal lexically; a case-insensitive collation orders by weight rather than bytes; a `NULL` key value satisfies neither `>=` nor `<`. A gap would let a differing row escape repair entirely, so a shortfall or excess means recopy the whole chunk.

## Fallbacks

Everything that makes narrowing unusable falls back to repairing the whole range, because declining to fix a known difference is not an option: too few distinct key values to cut, a failed boundary query, no piece differing, and — importantly — *every* piece differing. That last one is the systematic case (a lossy `ALTER`, a wrong column mapping, a truncated target) where narrowing buys nothing; without it, one chunk fanned out into 64 repairs and 128 transactions for the same write volume.

Relatedly, `fixChunkTimeout` and its detachment from the caller's cancellation are now owned once per chunk and shared across its narrowed repairs. The narrowed repairs together rewrite no more than the single repair they replace, so one budget covers them — and giving each its own would multiply the span in which a fix ignores cancellation by the number of pieces.

A verification *failure* is propagated rather than swallowed: it means the snapshot is gone, so the pass should fail and be retried. `differencesFound` is incremented before narrowing starts, so a chunk left unrepaired this way cannot be reported as a pass.

## Sharded sources

Boundaries come from one shard, since offsets into one shard's rows are the only row-position information a single query gives. The shard is chosen **per range**, not once per chunk: a sub-range's rows can sit almost entirely on a different shard than its parent's did, and cutting it with a shard that holds nothing in it would find no boundaries and silently give up on narrowing.

## Testing

`pkg/table` and `pkg/checksum` pass, both under `-race`; `golangci-lint` clean.

The strongest assertion uses a target-only `marker` column — the shape an "add column" migration has. It is invisible to the checksum, which compares intersecting columns only, but the repair resets it on every row it rewrites, so counting cleared markers measures exactly how much of the chunk was recopied. Covered for both the single-server (`REPLACE INTO ... SELECT`) and distributed (`DELETE` + applier re-apply) repair paths.

Also covered: exact partitioning for single-column, composite, bounded, unbounded and filtered ranges; key ties on a non-unique key; the `ENUM` ordinal-vs-lexical case; the count-validation fallback in both the gap and overlap directions; the all-differ and none-differ fallbacks; per-range shard selection; and verification-error propagation.

### DIFF
--- a/pkg/checksum/README.md
+++ b/pkg/checksum/README.md
@@ -62,6 +62,28 @@ The actual implementation includes additional handling:
 
 The CRC32 + XOR aggregate technique for table checksumming was pioneered by **pt-table-checksum** from Percona Toolkit, which established this as a reliable method for verifying data consistency in MySQL. This same approach has since been adopted by other database tools, including TiDB's data migration and verification utilities, demonstrating its effectiveness for distributed database scenarios.
 
+## Repairing a mismatch
+
+When a chunk fails verification and `FixDifferences` is on, the checker recopies the chunk's key range: `DELETE` then `REPLACE INTO ... SELECT` for the single-server case, or `DELETE` on every target followed by a re-apply through the applier for the distributed case.
+
+Chunks are sized by how long they take to *read* while checksumming, which says nothing about how long they take to rewrite. Recopying a whole XL chunk to fix a handful of rows is a large write burst, a long lock hold, and a wide window in which the target range is deleted but not yet rewritten. So the repair is narrowed first (`subdivide.go`):
+
+1. Cut the chunk's key range into `csRepairSplitParts` contiguous pieces (`table.Chunk.Split`, using `ORDER BY key LIMIT 1 OFFSET n` boundary lookups). The first and last inherit the original's bounds — inclusivity and unboundedness included — and each interior cut is an exclusive upper bound on one piece and an inclusive lower bound on the next.
+2. Re-checksum each piece **inside the same snapshot transaction that observed the mismatch**, and keep only the pieces that differ.
+3. Recurse on each differing piece, stopping at `csRepairMaxDepth` rounds or once a piece holds fewer than `csRepairMinSplitRows` rows — below that, recopying is cheaper than analysing.
+
+Step 2 is sound because `BIT_XOR` is associative and the row counts are additive: over an exact partition read in one snapshot, the pieces must reproduce the whole-range result, so a mismatching range must contain a mismatching piece. Reading a piece from a *different* snapshot would break that — a piece could look clean because a concurrent write happened to fix it, and the repair would be skipped on the strength of an observation that never applied to the chunk being fixed.
+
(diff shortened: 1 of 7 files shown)


morgo:
I think this might be the wrong direction. We are narrowing checksum chunks primarily because the single server checksum still uses `REPLACE INTO`. What we should do instead is ensure the checksum always uses appliers so the recopy phase is lock less.

PR #1090, opened 2026-07-27
checksum: size chunks for the checksum, not for the copier

morgo:
> **Stacked on #1089** — this targets the `checksum-repair-subdivide` branch, so the diff shows only this PR's own change. Retarget once #1089 merges.

This is PR 3 of 3 in the checksum-autoscaling series. PR 1 (#1087) made the checksum throttler-aware and scalable, and added a per-pass measurement of chunk read times specifically to answer one question: *which ceiling is actually binding on checksum chunk size?* PR 2 (#1089) made a mismatch repair proportional to the mismatch rather than to the chunk, so that growing chunks does not grow the repair. This PR reads the measurement and acts on it.

## What the measurement said

A 1M-row, 6-column table (194 MB against a 128 MB buffer pool), `--threads 4`:

```
checksum chunk size distribution chunks=126 target=500ms duration-p50=8ms
  duration-p90=30ms duration-max=44ms rows-p50=5062 rows-max=25624 row-capped=0/126
```

Neither ceiling was binding. The p90 chunk was using **6% of its time budget**, and **not one chunk in 126** reached the row cap. The chunk size ended the pass at 25,624 rows, still climbing.

That is not a tuning problem, it is a convergence problem. The sizer's growth is capped at `MaxDynamicStepFactor` (1.5×) per feedback window, and a window is 10 chunks — so climbing from `StartingChunkSize` (1000 rows) to `MaxDynamicRowSize` (100k) takes roughly **130 chunks**. This table had 126. The checksum spent its entire life converging and never once used the chunk size it had measured as correct. Any table under a few million rows behaves the same way, and raising the time target alone makes it *worse*, not better: a bigger target means a longer climb.

## What this changes

The checksum gets its own two settings, because a chunk's duration bounds different things in the two phases. A copy chunk's time is a write transaction's lifetime — a latency budget that decides how long a row lock is held and how far a replica appears behind. A checksum chunk's time bounds neither: the CRC is aggregated server-side so exactly one row per chunk crosses the wire, and it runs inside a `REPEATABLE READ` snapshot that is held for the whole pass either way.

- **Starting size** — `checksum.ChunkStartRows` (the row cap) instead of `table.StartingChunkSize`. New `ChunkerConfig.StartingChunkSize`, honoured by both chunker implementations and across `Reset` and the optimistic chunker's algorithm switches. This is the change that mattered.
- **Target time** — `--checksum-target-chunk-time` (10s) instead of the copier's `--target-chunk-time` (500ms). For `move`, whose `--target-chunk-time` is already documented as checksum-only, the default moves 5s → 10s instead of adding a flag.

Starting at the cap is safe because the sizer's asymmetry runs the other way from the copier's: it shrinks with **no** per-step cap, and panic-shrinks on any single chunk exceeding `DynamicPanicFactor` × the target. Overshooting costs one slow chunk. Undershooting cost the whole pass.

Same table, after:

| | chunks | rows p50 | duration p90 | at the row cap |
| --- | --- | --- | --- | --- |
| before | 126 | 5,062 | 30ms | 0/126 |
| after | 12 | 100,000 | 162ms | 10/12 |

## What I am *not* claiming

Per-row cost at p50 is unchanged on this box (~1.6µs either way). Read-ahead has little left to win when the storage is a local SSD, and "larger chunks engage InnoDB linear read-ahead and Aurora's batched prefetch" is a claim about network-attached storage that a laptop cannot demonstrate. What the numbers do show is p90 per-row cost falling with the chunk count (5.9µs → 1.6µs) as per-chunk overhead amortises over 20× the rows, and the sizing reaching its bound in two chunks instead of never.

I also did **not** raise `MaxDynamicRowSize`, so this does not deliver literal 10s chunks — at 100k rows the cap binds first, and 10s becomes a safety valve rather than a goal. That is deliberate. The row cap is a bound that can be reasoned about: it is what bounds a repair (PR 2), and what `inspectDifferences` holds in memory on a mismatch. A binding *time* target would instead let chunk size follow load. Raising the cap is a separate decision, and it is now a decision with data behind it — `row-capped` reads 10/12 rather than 0/126, so the instrument PR 1 added will say when the cap is costing something.

## Costs, both bounded by the row cap

- **Throttle latency.** A throttle decision takes effect between chunks (`BlockWait` before dispatch; chunks in flight are never abandoned), so back-off now tracks a chunk's duration rather than 500ms.
- **Small-table parallelism.** A table with fewer than `threads` × 100k rows yields fewer chunks than there are workers. Affordable precisely because it is a small table.
- **Gap prefetch fires sooner.** On an auto-increment key, being at the row cap from chunk one means `maybeSwitchToPrefetch` now fires on the first feedback window rather than never. That is a net gain — in prefetch mode a chunk holds exactly `chunkSize` *rows* rather than `chunkSize` *key values*, so chunks stay uniform across a key space pitted with deletes — at the cost of one index-only `OFFSET` query per chunk.

The continuous (during-copy) checksum is deliberately left on the copier's budget: it shares the copy phase, so it is held to copy-phase latency. `--target-chunk-time`'s help text and the docs now say so, since it no longer covers the cutover checksum.

## Testing

`pkg/table`, `pkg/checksum`, `pkg/copier` and `pkg/migration` pass, `pkg/table` and `pkg/checksum` also under `-race`; `golangci-lint` clean.

New tests cover the starting size on both chunker implementations (including that it survives `Reset`, and that it is clamped to the dynamic bounds without going through the growth-capping path), and the runner wiring that gives the checksum a different profile from the copier — which is the thing a future refactor is most likely to undo silently.

One existing test needed its fixture resized: `advanceRunnerToChecksumWatermarks` seeded 4096 rows to get "more than one chunk on the runway", which is now less than one checksum chunk. It now sizes the runway against `checksum.ChunkStartRows`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -191,7 +191,7 @@ scripts/      → Build and run helper scripts
 
 ### Key design decisions
 
-- **Dynamic chunking**: chunk size auto-adjusts against a target based on the 90th percentile of the last 10 chunks, rather than a fixed row count. The default buffered copier targets an in-memory *byte budget* (`--target-chunk-size`, default `table.DefaultTargetChunkBytes` = 16 MiB); the checksum and legacy `--unbuffered` copier target a *chunk time* (`--target-chunk-time`, default 500ms for migrate).
+- **Dynamic chunking**: chunk size auto-adjusts against a target based on the 90th percentile of the last 10 chunks, rather than a fixed row count. The default buffered copier targets an in-memory *byte budget* (`--target-chunk-size`, default `table.DefaultTargetChunkBytes` = 16 MiB); the legacy `--unbuffered` copier and the continuous checksum target a *chunk time* (`--target-chunk-time`, default 500ms for migrate). The cutover checksum targets its own much larger chunk time (`--checksum-target-chunk-time`, default `checksum.DefaultTargetChunkTime` = 10s) and starts at the row ceiling rather than ramping up to it (`checksum.ChunkStartRows`), because growth is capped per feedback window and the ramp is longer than many whole tables — see `pkg/checksum/README.md`.
 - **Change row map**: binlog changes are deduplicated in a map before flushing, so a row updated 10 times is only copied once.
 - **High watermark optimization**: binlog changes above the copier's current position are discarded (only for auto-increment PKs).
 - **Checkpoint/resume**: progress is saved periodically; interrupted migrations resume automatically with ~1 minute of lost progress.
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ The following are some of the optimizations that make Spirit faster than gh-ost:
 Rather than accept a fixed chunk size (such as 1000 rows), Spirit dynamically adjusts the chunk size against a target. This is both safer for very wide tables with a lot of indexes and faster for smaller tables. The target depends on the copier:
 
 - The **default buffered copier** reads full rows into memory, so it sizes each chunk against an in-memory **byte budget** (`--target-chunk-size`, default 16 MiB). Time is a poor signal here — the buffered copier's measured chunk time includes waiting behind the write queue, which inflates under load independently of chunk size — whereas a byte budget is a stable property of the data and keeps chunks large enough to engage InnoDB/Aurora read-ahead.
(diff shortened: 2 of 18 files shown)


morgo:
Similar to https://github.com/block/spirit/pull/1089 I think this is the wrong direction. If we use the appliers for checksum recopy then we can just set the checksum chunk time target to 5s and be done with it.

PR #1096, opened 2026-07-29
autoscale: own the thread counts; drop --write-threads auto-sizing

morgo:
## Problem

On a 96-vCPU Aurora instance the checksum autoscaler converged at **4 workers** and stopped:

```
autoscaler adjusting checksum threads from=2 max=4 min=1 reason="utilization below low watermark" to=3
autoscaler adjusting checksum threads from=3 max=4 min=1 reason="utilization below low watermark" to=4
(silence)
```

The cap was binding, not the signal — `max=4` is `2 × --threads`, and `--threads` was at its default. In the same migration the write pool had auto-sized itself to 94 from the instance (`applier-workers=94`), so we were reading with 4 threads and writing with 94.

The flags and the controller were both trying to size the pool.

## Change

When autoscaling engages, it owns the thread counts: `--threads` and `--write-threads` are ignored and each pool is sized from the instance instead. A controller told to find the right size should not also be told where to stop, least of all by a number the caller left at its default.

| Pool | Starts at | Ceiling |
| --- | --- | --- |
| Apply (write) | `vCPUs - 2` (min 1) | `2 × start` |
| Copy read | `ceil((vCPUs - 2) / 4)` (min 2) | `ceil(vCPUs / 2)` |
| Checksum | same as copy read | `ceil(vCPUs / 2)` |

On a 96-vCPU instance: 94 write threads growing to at most 188, and 24 read (then checksum) threads growing to at most 48.

### Why the read side is shaped differently

Write threads spend most of their life parked on a redo-log flush, so a count above the vCPU count is not oversubscription — it is what keeps the log busy, and it is why the redo-aware signal excludes those waiters. A read thread scanning a table that is already in the buffer pool is pure CPU, so the same count really does take cores from the application; oversubscribing readers is how a checksum ends up degrading the workload it was supposed to be invisible to. Readers therefore start at about a quarter of the instance and earn their way up.

### Why the ceiling is a share of the box, not a multiple of the start

The checksum's snapshot pool **cannot grow**: every transaction must take its read view at the same instant, so the entire pool is created serially while the brief table lock is held. Its ceiling is not a hypothesis the controller might reach — it is lock time spent up front, on every run, whether scaling reaches it or not.

That rules out a ceiling of `vCPUs`: 96 transactions started serially under a cutover-class lock, to reserve capacity the load signal may well never hand out. `ceil(vCPUs / 2)` is the compromise — it bounds both the cores taken from the workload and the lock window, and neither of those has anything to do with where the ramp began, which is why it stays anchored to the instance rather than tracking the start.

(At exactly `MinVCPUs` the two bounds meet — start 2, ceiling 2 — so the read side can shed but not grow on a 4-vCPU instance. That is the intended reading of a box that small.)

### `--write-threads 0 = auto` is gone

Auto-sizing existed to reach the instance vCPU count without a flag. `--enable-experimental-autoscaling` now does that and more, and ignores `--write-threads` when it engages — so the sentinel was a second, quieter way to ask for the same thing, hidden inside a count.

`--write-threads` is now always the literal number of apply threads, on both `migrate` and `move`, defaulting to 4. A zero from a programmatic caller is filled in with that default (migrate in `normalizeOptions` beside `Threads`, move in `NewRunner` beside the other zero-means-default fields) rather than triggering a probe. `throttler.ResolveWriteThreads`, `DefaultWriteThreads` and `WriteThreadVCPUReserve` are removed; the reserve lives on as `autoscale.VCPUReserve`, where the write start is now derived, and its pinning test moves with it. `AuroraVCPUs` stays — the throttler and the autoscaler both read it.

**This is a behaviour change for anyone passing `--write-threads 0` today**: they now get 4 rather than an instance-sized pool, and should pass `--enable-experimental-autoscaling` instead.

## Scope of the override

Flags are only overridden when the controller can actually steer. Nothing changes with a non-Aurora target, with `--unbuffered`, or below `autoscale.MinVCPUs` — the counts stay exactly as configured, and one line is logged either way. `spirit move` has no autoscaling wiring at all, so it always honors both flags.

## Worth reviewing

- **Nominal oversubscription at copy start.** 94 write + 24 read = 118 threads nominal against a 97-thread hard stop on a 96-vCPU instance. The dual-pool law and the throttler arbitrate that down, and the write side starts *above* the band precisely so it sheds — but it is the largest behavioural change here.
- **Connection budget.** `maxRead + maxWrite + …` becomes `48 + 188 + …` ≈ 240 on a 24xlarge, up from `8 + 188 + …`. Worth a look if you have proxy connection limits.
- **Checksum lock window.** Still grows with the ceiling — 48 serial transactions rather than 8. Halving the ceiling halves that, but it is the number to watch in the staging run.
- **The two formulas coincide at most real sizes.** For any vCPU count that is a multiple of 4 above `MinVCPUs`, `ceil(vCPUs / 2)` happens to equal `2 × start`. They are different rules for different reasons (one bounds a share of the box, the other scales a ramp) and the code keeps them separate, but the arithmetic agreeing is worth knowing when reading logs.

## Testing

- `autoscale.ReadBounds` covers the sizing table across 4–96 vCPUs, the never-above-half-the-instance invariant, and the below-`MinVCPUs` coherence cases.
- `autoscale.VCPUReserve` pins the reserve and the write-side vCPU→threads mapping (moved from `pkg/throttler`).
- `copier.resolveReadCeiling` covers supplied-vs-derived ceilings and the floor-at-start clamp.
- `TestAutoscalingLeavesThreadFlagsAloneWhenItCannotEngage` runs a real migration with `--enable-experimental-autoscaling --threads 3 --write-threads 5` against the (non-Aurora) test server and asserts both flags survive.
- Full suite green apart from the privilege tests that need compose root credentials locally.
- The engaged path still needs a staging run against a real instance — it reads a vCPU count no local target reports.

Docs updated: `docs/migrate.md` (threads, write-threads, the autoscaling section and its checksum subsection), `docs/move.md`, `pkg/autoscale/README.md`, `pkg/copier/README.md`, `pkg/checksum/README.md`.


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1096, 9a01bae.

**Verdict:** 4 findings — one real edge-case in the engage gate (transient `IsAurora` probe error silently bypasses two safety guards), one silent behavior break for explicit `--write-threads 0`, and two doc-vs-code mismatches this PR introduced. None is a data-correctness risk; the core mechanism is sound.

### Findings

1. **A transient `IsAurora` probe error engages the autoscaler with flag-relative bounds and silently skips two safety guards.** [pkg/migration/runner.go#L733](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L733): `if isAurora, err := throttler.IsAurora(ctx, r.db); err == nil && isAurora {` treats a probe *error* the same as non-Aurora — the override block is skipped but `autoscaleEnabled` stays true. `IsAurora` is an uncached per-call query, and the throttler is built later by an independent probe ([pkg/throttler/aurora.go#L104](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/throttler/aurora.go#L104), invoked from `setupThrottler` after `setupCopierCheckerAndReplClient`). So on a real Aurora target, one transient connection blip at the first probe yields: a `GradualThrottler` is built, `autoscalerIfEnabled` fully engages, but with flag-relative bounds — the configuration the PR's own docs say no longer exists ("the flags are only overridden when the controller can actually steer"). Concretely, that also bypasses (a) the `MinVCPUs` guard at [runner.go#L738](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L738) — a <4-vCPU instance scales and oscillates, exactly what the guard prevents — and (b) the unguarded-redo cap: `redoAware` stays false ([runner.go#L745](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L745) is inside the success branch), so with `--max-commit-latency=0` the write ceiling is 2×start instead of the safety-capped start, while the throttler independently selects redo-aware mode — growth against a signal that deliberately ignores redo-log waiters, with no commit-latency backstop. The comment at [runner.go#L729](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L729) ("an IsAurora probe failure is benign here … the autoscaler stays dormant regardless") is only true for *persistent* failures; nothing enforces the correlated-failure assumption. Minimal fix: treat a probe error as `autoscaleEnabled = false` (fail-static), or share one probe result between the two sites.

2. **Explicit `--write-threads 0` is silently coerced to 4 with nothing logged.** [pkg/migration/migration.go#L165](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/migration.go#L165) and [pkg/move/runner.go#L189](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/move/runner.go#L189). The PR description acknowledges the break, but at runtime it is invisible: an operator who adopted the previously documented `--write-threads 0 = auto` opt-in reruns the same command after upgrading and their apply pool drops from e.g. 94 (vCPUs−2 on a 24xlarge) to 4 — a ~24× write-parallelism regression with no error and no log line pointing at the flag. Since the kong default is 4, a literal 0 can only appear when explicitly passed (or from a programmatic caller), so a one-line warning in `normalizeOptions`/`NewRunner` ("write-threads 0 no longer means auto; using 4 — use --enable-experimental-autoscaling for instance-derived sizing") is precisely targeted and cheap.

3. **Docs: "On stock MySQL the flag does not engage at all" is wrong and contradicts the bullet two lines above it.** [docs/migrate.md#L479](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/docs/migrate.md#L479). On a non-Aurora target `autoscaleEnabled` stays true and `checksum.AutoscaleConfig.Enabled=true`, so the checksum scaler runs: backlog shedding is active on any server (the code's own capability matrix in `checksum/autoscaler.go` says "stock MySQL: shed yes, grow no"), which is what [L478](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/docs/migrate.md#L478) ("**Shedding** happens only with the flag, on any server") correctly states. The sentence even contradicts itself ("does not engage at all, so the checksum can shed"). Accurate wording: only *growth* never engages on stock MySQL because the utilization signal is absent.

4. **Docs: the checksum-pool provisioning dichotomy misses the flag-on-stock-MySQL case, where the pool is 2×`--threads` and the ceiling is unreachable.** [docs/migrate.md#L485](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/docs/migrate.md#L485) says the snapshot pool is provisioned at "`ceil(vCPUs / 2)` with this flag, plain `threads` without it". With the flag on a non-Aurora target, `readCeiling` stays 0 and `maxRead = ResolveMaxReadThreads(Threads, true)` = 2×`--threads` ([pkg/migration/runner.go#L783](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L783)), and [pkg/checksum/checksum.go#L254](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/checksum/checksum.go#L254) provisions that many `START TRANSACTION WITH CONSISTENT SNAPSHOT` serially under the table lock — capacity growth can never use, since growth requires a `GradualThrottler` (Aurora-only). The pre-PR wording ("2 × threads with this flag") was correct for this case; the doc's own rationale says the lock-window cost matters. Worth fixing the sentence to three cases — and arguably worth passing `MaxThreads: Threads` to the checksum when no gradual signal will exist, since half the provisioned transactions are pure lock-window overhead there (pre-existing behavior, but this PR is what documents the cost).

The riskiest mechanism is the in-place override — `r.migration.Threads = readStart; r.migration.WriteThreads = writeStart` at [runner.go#L763](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L763) — mutating the flag values mid-setup while the rest of the runner keeps reading them. Traced every consumer: the pool seed in `Run` ([runner.go#L276](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L276)) reads pre-override values but the pool only ever grows and is re-finalized post-override at [runner.go#L797](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L797); `maxWrite` is computed *after* the override ([runner.go#L775](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L775)); the applier `Threads`, copier `Concurrency`, checksum `Concurrency`/`MaxThreads`, and the throttler build all happen downstream of the override in the same setup call; the resume-from-checkpoint path goes through the same function; and preflight checks read the configured values by design (they run before setup). No consumer reads a stale count. `ReadBounds` arithmetic was independently checked against the PR's doc table at every listed size (4→2/2, 8→2/4, 16→4/8, 96→24/48, 192→48/96) — all correct, including the start-floor/ceiling meet at 4 vCPUs.

### Notes (not blocking)

- **Dropped runtime rejection of negative `WriteThreads` for library callers.** Deleted `ResolveWriteThreads` rejected negatives on every path; `Migration.Validate`/`Move.Validate` are Kong-only hooks never invoked programmatically, and the new fill-ins only fix `== 0`. A programmatic `WriteThreads: -5` now silently runs with 2 clamped write workers (migrate), and on the move path `MaxOpenConnections = Threads + WriteThreads + 2` can go negative — `SetMaxOpenConns(-2)` means *unlimited*, and the `setupUnderLocks` growth guard recomputes the identical formula so it never corrects ([pkg/move/runner.go#L1009](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/move/runner.go#L1009)). Degenerate misuse, no in-repo caller exposed — but normalizing `<= 0` instead of `== 0` restores fail-safe for free.
- **The write-side formula has no shared helper and its test is tautological.** `writeStart := max(1, vCPUs-autoscale.VCPUReserve)` is inlined at [runner.go#L756](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L756) while `TestVCPUReserve` re-computes the same expression inside the test body — it pins the constant, not the production mapping. A `WriteBounds` sibling to `ReadBounds` would be tested for real and keep both instance-derived formulas in one package.
- **Runner-triplet drift on the zero→default fill-in** (AGENTS.md "Keeping the runner triplet in sync"): three styles in three runners — bare `4` in [migration.go#L166](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/migration.go#L166), a documented `defaultWriteThreads` const in [move/runner.go#L38](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/move/runner.go#L38), pre-existing bare `4` in [datasync/runner.go#L147](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/datasync/runner.go#L147). The PR deleted `throttler.DefaultWriteThreads`, which was exactly the shared constant.
- **Preflight `settings` check still enforces `--threads` ∈ 1–64 on a flag autoscaling will ignore** ([pkg/migration/check/settings.go#L17](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/check/settings.go#L17)): `--enable-experimental-autoscaling --threads 100` fails preflight for a value that would never be used, while the derived counts (up to 48 read / 190 write on a 48xlarge) are never validated against any bound.
- **Test hygiene in the new integration test** ([runner_pool_test.go#L37](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner_pool_test.go#L37)): raw `RunSQL` with a partial manual drop (misses `_old`/`_chkpnt`) and no table cleanup, where AGENTS.md documents `testutils.NewTestTable` as the preferred helper for exactly this lifecycle; `require.Positive(t, autoscale.MinVCPUs)` asserts a compile-time constant and can never fail.
- **Observability:** the startup log ([runner.go#L238](https://github.com/block/spirit/blob/9a01baef610ab85985b59a1800912ecd9f1871b0/pkg/migration/runner.go#L238)) reports `concurrency=` from the flag value autoscaling later discards; and on the resume-fails-then-fresh path `setupCopierCheckerAndReplClient` runs twice, so the second "autoscaling engaged" line reports the first run's *derived* values as `configured_threads`/`configured_write_threads` (the deleted `ResolveWriteThreads` carried an explicit idempotency contract for this double-call path).

### Verified correct

- `ReadBounds` start/ceiling arithmetic matches the doc table at every listed instance size, clamps coherently below `MinVCPUs` (start ≤ ceiling, start > 0), and the never-above-half-the-instance invariant holds.
- The override happens before every downstream consumer (applier threads, copier concurrency, checksum concurrency/ceiling, connection-pool finalization, throttler build), on both the fresh and resume paths.
- All removed symbols (`ResolveWriteThreads`, `DefaultWriteThreads`, `WriteThreadVCPUReserve`) are fully swept — zero remaining references in code or docs; the reserve constant and its pinning test moved intact to `pkg/autoscale` (with a new 96→94 case).
- Non-engage paths (non-Aurora, `--unbuffered`, Aurora < `MinVCPUs`) genuinely leave both flags untouched, pinned by the new `TestAutoscalingLeavesThreadFlagsAloneWhenItCannotEngage` integration test; the connection budget there is still sized from the flag-implied ceilings, which only over-budgets (harmless).
- The checksum's snapshot pool is provisioned at exactly the same `maxRead` the connection pool is sized for, so the lock-window transactions never starve on connections; shedding-signal wiring (change-feed residual) is independent of throttler type as documented.
- `spirit move` honors both flags everywhere (no autoscale wiring), and the zero→default fill-in lands in `NewRunner` beside the other zero-means-default fields as the PR claims.
- CI is fully green on 9a01bae (8 build matrix jobs, lint, govulncheck, Semgrep).

Kiran01bm (commented):
stamping it but needs agent review findings to be reviewed - https://github.com/block/spirit/pull/1096#issuecomment-5126596525

morgo:
🤖 Thanks Kiran — went through all ten. Four fixed, plus two of the notes; three declined with reasons. Pushed as 6bc05516.

**Fixed**

1. **`IsAurora` probe error fell through with autoscaling still on.** You were right that the old "a probe failure is benign, matching `AuroraSetup.Build`" comment didn't hold: that's a *second, independent, uncached* query, so a blip here doesn't stop the throttler from selecting a `GradualThrottler` later. The copier would then have scaled against exactly the flag-relative bounds this PR exists to replace, with both guards skipped — the `MinVCPUs` check, and `redoAware` staying `false` so an unguarded redo log gets the 2x ceiling instead of the capped one. The gate is now a `switch` and a probe error disables autoscaling and logs why. Non-Aurora still stays *enabled* on purpose — the checksum's backlog veto works on any server, so the flag still buys shedding there; only growth needs Aurora.

2. **`maxRead` fallback over-provisioned the checksum's lock window.** It was `ResolveMaxReadThreads(Threads, autoscaleEnabled)` → 2x. A zero `readCeiling` means the gate derived nothing, which is precisely the case where neither read pool can grow (growth needs the Aurora signal). The checksum turns that ceiling into transactions started *serially under the table lock* whether scaling reaches it or not, so the 2x was lock time bought for capacity nothing could use. Now the configured count and no more.

3. **`--write-threads 0`.** Zero used to mean auto-size, so anyone who'd adopted that opt-in would silently drop from the instance vCPU count to 4. Both runners now normalize `<= 0` and warn on a literal `0`, naming the flag that replaced it. Non-positive rather than `== 0` matters in move: `MaxOpenConnections` is `Threads + WriteThreads + 2`, and a negative reaches `SetMaxOpenConns` as *unlimited*.

4. **Both doc claims.** "the flag does not engage at all" on stock MySQL contradicted the shedding it does buy — the capability matrix in `pkg/checksum/autoscaler.go` says hard stop yes / shed yes / grow no. Reworded, and the checksum-pool provisioning sentence now says which targets get `ceil(vCPUs/2)` versus plain `threads`.

**Also took**

- Hoisted the write start into `autoscale.WriteStart(vCPUs)` next to `ReadBounds`, so both derivations live in one file, and `TestWriteStart` now calls it instead of restating the expression (it could not previously fail if the production formula changed). Documented why there's no `WriteCeiling`: the write bound isn't purely a function of the instance, it also depends on the throttler's signal mode and the commit-latency backstop, so it stays in `ResolveMaxWriteThreads`.
- Dropped the `configured_threads`/`configured_write_threads` log keys. Good catch — `setupCopierCheckerAndReplClient` runs twice when a resume attempt fails and falls back to a fresh migration, so on the second pass those fields hold the *first* pass's derived values, not anything the caller configured.
- Test hygiene: the new integration test is on `testutils.NewTestTable` now (so `_old`/`_chkpnt` get cleaned up too), and the tautological `require.Positive(t, autoscale.MinVCPUs)` is gone. The pool assertion changed with fix 2 — read term is 1x, write term still 2x — with a comment on why the two pools budget differently.

**Declined**

- **Preflight `--threads` 1–64 range check** (`pkg/migration/check/settings.go`): it runs before we know whether autoscaling will engage — engaging needs a live `IsAurora` + vCPU read against the target. Validating the flag the user actually passed is the only thing it can do at that point, and it's the right thing when the gate doesn't engage.
- **`pkg/datasync/runner.go:147`'s bare `4`**: already a `<= 0` fill-in with correct behaviour. Datasync doesn't autoscale, so hoisting a shared constant across the package boundary is churn with a nonzero chance of changing something, for no behaviour change.
- **The `concurrency=` line at `runner.go:238`**: it's the startup banner, logged before setup runs, so it reports the flag value by construction. The engage line then says explicitly that the flags are ignored and prints the derived counts. Making the banner conditional would mean deferring it past setup.

One open question for you, not from this review — at every real instance size (a multiple of 4 above `MinVCPUs`), `ceil(vCPUs/2)` is exactly `2 × start`, so the `MaxReadThreads` plumbing through `AutoscaleConfig` is numerically a no-op on real hardware and `autoscale.Ceiling(start, true)` would give the same answers. Happy to delete it if you'd rather have less surface; I left it in because it makes the "half the instance" bound explicit rather than emergent, and it holds if the start formula ever moves.

aparajon (approved):

PR #1106, opened 2026-07-31
update readme

morgo:

This updates some of the performance talk in the readme:

- I updated the spirit performance numbers since they were pre-buffered applies, and do not faithfully represent the current experience. I used a 1.43TiB table not a 10TiB because it was easier to test, and is more faithful to a typical case.
- Similarly I removed gh-ost comparisons. I know they've made at least some performance enhancements. To be fair we'd need to re-run the comparisons, and keep them current, which we're not in an easy position to do.
- I documented autoscaling here. It's a massive win.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ It is similar to gh-ost except:
 - It only supports MySQL 8.0 and higher
 - It is multi-threaded in both the row-copying and the binlog applying phase
 
-The goal of Spirit is to apply schema changes as fast as possible. This makes it unsuitable in the following scenarios:
+The goal of Spirit is to apply schema changes as fast as possible, while also preserving safety. This makes it unsuitable in the following scenarios:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,7 @@ It is similar to gh-ost except:
 - It only supports MySQL 8.0 and higher
 - It is multi-threaded in both the row-copying and the binlog applying phase
 
-The goal of Spirit is to apply schema changes much faster than gh-ost. This makes it unsuitable in the following scenarios:
+The goal of Spirit is to apply schema changes as fast as possible. This makes it unsuitable in the following scenarios:
```
aparajon: I wonder if we should word this as "as fast as possible while also preserving safety" or something along those lines

aparajon (approved):

README.md
morgo: Love it, I'll add it.

PR #1112, opened 2026-08-06
fix(dbconn): return committed retry row count

jonwinton:
## Summary

- Return `RowsAffected` from the committed `RetryableTransaction` attempt only.
- Discard row counts from rolled-back attempts.
- Return zero rows when a fatal error or retry exhaustion prevents a commit.

First correctness slice extracted from #1111.

## Verification

- `MYSQL_DSN=tsandbox:msandbox@tcp(127.0.0.1:8043)/test go test -race -count=1 -run '^(TestRetryableTransactionRowsAffectedExcludesRolledBackAttempts|TestRetryableTrx)$' ./pkg/dbconn`

## Review order

### DIFF
--- a/pkg/dbconn/dbconn.go
+++ b/pkg/dbconn/dbconn.go
@@ -181,6 +181,7 @@ func RetryableTransaction(ctx context.Context, db *sql.DB, dupKeyHandling DupKey
 	)
 	for i := range config.MaxRetries {
 		func() {
+			var attemptRowsAffected int64
 			// Start a transaction
 			if trx, err = db.BeginTx(ctx, nil); err != nil {
 				return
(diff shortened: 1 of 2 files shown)


morgo (approved):

jonwinton:
CodeQL infrastructure note: dynamic run [31125495568](https://github.com/block/spirit/actions/runs/31125495568) was cancelled after roughly one hour without executing any job steps. `gh run rerun 31125495568 --failed` reports that the workflow cannot be retried, and a direct check-suite rerequest is unavailable (404). All non-CodeQL checks pass; no code-level CodeQL finding was produced.

aparajon (approved):
🤖 Verified: the per-attempt accumulator is assigned to the result only after a successful commit, and only one commit can ever happen per call (the loop exits on success), so rolled-back attempts can no longer contribute affected rows. The fatal-error path now correctly reports 0 with the error. Note the Go test workflow never ran at this head — `go test -race ./pkg/dbconn` passes locally against MySQL (53s), including both new assertions.

PR #1113, opened 2026-08-06
feat(copier): expose completed work totals

jonwinton:
## Summary

- Add an optional `CompletedWorkReporter` capability without changing the `Copier` interface.
- Track exact completed rows and chunks for buffered and unbuffered copiers.
- Reset totals per run, count zero-row completed chunks, and record only successful durable application.

Exporter-neutral prerequisite extracted from #1111.

## Verification

- `go test -race -count=1 -run '^TestCompletedWork' ./pkg/copier`

## Review order

### DIFF
--- a/pkg/copier/buffered.go
+++ b/pkg/copier/buffered.go
@@ -37,6 +37,8 @@ type buffered struct {
 	isInvalid        atomic.Bool
 	errMu            sync.Mutex // guards firstErr
 	firstErr         error      // first error that invalidated the copy (any goroutine)
+	completedRows    atomic.Uint64
+	completedChunks  atomic.Uint64
 	startTime        time.Time
 	throttler        throttler.Throttler
 	dbConfig         *dbconn.DBConfig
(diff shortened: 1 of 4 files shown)


morgo (commented):
> * Add an optional `CompletedWorkReporter` capability without changing the `Copier` interface.

I'm happy for you to change/extend the interface. We don't have any external users of this interface to my knowledge.

In the migration caller there is some overlap with this, where we get copied rows from the chunker. I actually wonder if this should move to the copier here. They technically mean different things, but they are close enough we can standardize on the copier version.

aparajon (approved):
🤖 Sound, and +1 to Morgan's suggestion to standardize on the copier's version (extending `Copier` directly is fine by us too). Two doc-level notes on the "authoritative aggregate" contract:

1. `Run()` resets the counters, so after a resume-from-checkpoint `CompletedWork()` reports this-process work, not migration-lifetime totals. Consumers building terminal accounting from it (strata events) should know which they're getting — worth a sentence on `CompletedWorkReporter`.
2. In buffered mode the rows are recorded at read/`Feedback` time, before the applier has durably landed them — on a failure path the count can overstate applied work. Fine for success-terminal accounting; worth stating.

Test workflow never ran at this head; `go test -race ./pkg/copier` passes locally.

PR #1114, opened 2026-08-06
feat(move): return cutover mutation results

jonwinton:
## Summary

- Add a result-bearing forward cutover callback that reports durable external mutation separately from callback error.
- Stop retrying once a callback reports a durable ownership or topology change.
- Preserve the legacy callback path and use one result-bearing helper for empty and full moves.
- Keep legacy and result-bearing setters mutually exclusive, including nil reset behavior.

Third independent correctness/API slice extracted from #1111.

## Verification

- `MYSQL_DSN=tsandbox:msandbox@tcp(127.0.0.1:8043)/test go test -race -count=1 ./pkg/move -run '^(TestCutOverResultControlsRetry|TestForwardCutoverCallbackConfiguration|TestEmptyDatabaseMoveNoTables)$'`

## Review order

### DIFF
--- a/pkg/move/cutover.go
+++ b/pkg/move/cutover.go
@@ -27,6 +27,15 @@ var renameRetryWait = 1 * time.Second
 // that state cannot converge, so the retry loop aborts immediately.
 var errRenameRollbackFailed = errors.New("rename rollback failed")
 
+// CutoverResult reports whether a caller-owned forward cutover callback made a
+// durable ownership or topology mutation, even when it also returned an error.
+type CutoverResult struct {
+	DurableMutation bool
+}
+
+// CutoverResultCallback is the result-bearing form of a forward cutover callback.
+type CutoverResultCallback func(context.Context) (CutoverResult, error)
+
 // CutOverSource holds per-source state needed for the cutover.
 type CutOverSource struct {
 	DB         *sql.DB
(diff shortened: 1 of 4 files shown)


morgo (approved):
I'm happy with this if you want to re-target main and merge it.

jonwinton:
CodeQL infrastructure note: dynamic run [31125806730](https://github.com/block/spirit/actions/runs/31125806730) was cancelled after roughly one hour without executing any job steps. `gh run rerun 31125806730 --failed` reports that the workflow cannot be retried, and a direct check-suite rerequest is unavailable (404). All non-CodeQL checks pass; no code-level CodeQL finding was produced.

aparajon (approved):
🤖 Traced the retry interleavings and they hold: legacy callback semantics unchanged; callback-succeeded-then-rename-failed keeps the existing retry-under-lock behavior (`mutated && !succeeded` is false); ctx-cancel after a mutated failure still returns the manual-intervention abort rather than `ctx.Err()`. Fail-closed on durable-mutation-then-error is exactly the right shape — it's the move-side answer to the evidence asymmetry from #1111.

Two notes:

1. The `DurableMutation` doc should tell callback authors to set it on **ambiguous** outcomes too (routing RPC timed out — did it land?): true whenever a mutation *may* have occurred, false only when provably nothing mutated. That fail-safe discipline is the flag's whole point, and the current wording ("made a durable ... mutation") reads as certain-knowledge-only.
2. Heads-up: this overlaps `pkg/move/runner.go` with #1118, so whichever lands second rebases.

CI at this head: both CodeQL Analyze jobs cancelled and no test workflow ran — locally `go build ./...` + the targeted cutover/move tests pass with `-race`.

PR #1115, opened 2026-08-06
feat(status): own workflow lifecycle events

jonwinton:
## Summary

- Make the existing `status.State` the only workflow phase vocabulary carried by typed events.
- Add a zero-value-safe `status.Lifecycle` that owns state writes, exact attempts, outcomes, completed-work payloads, durable mutation, and terminal evidence.
- Preserve exact parent contexts and synchronously isolate observer panics from application behavior.
- Use opaque attempt handles so concurrent state changes cannot mislabel completions.

This replaces the duplicate `WorkflowStage` direction from #1111. Runner adoption is isolated in #1116.

## Verification

- `go test -race -count=1 ./pkg/status`
- `go build ./...`
- `make lint`

## Review order

### DIFF
--- a/pkg/status/lifecycle.go
+++ b/pkg/status/lifecycle.go
@@ -1,0 +1,168 @@
+package status
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"sync/atomic"
+)
+
+// Lifecycle owns the current workflow state, event observer, and terminal
+// evidence delivery. Its zero value is ready for use.
+type Lifecycle struct {
+	state State
+
+	observerMu sync.RWMutex
+	observer   WorkflowObserver
+
+	durableMutationObserved atomic.Bool
+	terminalObserved        atomic.Bool
+}
+
+// Get returns the current workflow state.
+func (l *Lifecycle) Get() State {
+	if l == nil {
+		return Initial
+	}
+	return l.state.Get()
+}
+
+// Set updates the current workflow state without emitting an event.
+func (l *Lifecycle) Set(state State) {
+	if l == nil {
+		return
+	}
+	l.state.Set(state)
+}
+
+// SetObserver replaces the workflow observer. A nil observer disables delivery.
+func (l *Lifecycle) SetObserver(observer WorkflowObserver) {
+	if l == nil {
+		return
+	}
+	l.observerMu.Lock()
+	l.observer = observer
+	l.observerMu.Unlock()
+}
+
+// HasObserver reports whether workflow event delivery is enabled.
+func (l *Lifecycle) HasObserver() bool {
+	if l == nil {
+		return false
+	}
+	l.observerMu.RLock()
+	hasObserver := l.observer != nil
+	l.observerMu.RUnlock()
+	return hasObserver
+}
+
+// Start records state before synchronously delivering its started event. Each
+// call returns a distinct attempt, including repeated starts of the same state.
+func (l *Lifecycle) Start(parent context.Context, state State) WorkflowAttempt {
+	if l == nil {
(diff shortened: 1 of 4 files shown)


morgo:
I have an alternative proposal to this in https://github.com/block/spirit/pull/1118

I believe it's similar enough in intention.

aparajon (commented):
🤖 Reviewed side-by-side with #1118, from the SchemaBot/Strata consumer seat. Local: `go build ./...` clean, `go test -race ./pkg/status` green (note: the Go test workflow never ran at this head — only DCO/Semgrep/zizmor).

First, credit where due: this fixes the thing I cared most about from #1111 — `WorkflowEvent.State` is literally `status.State`, so there is no second *phase* vocabulary. The payload/outcome/evidence layer here is genuinely good and worth keeping whichever shape lands.

Findings, if this merges as-is:

| # | Tier | Item |
|---|---|---|
| 1 | Fix before merge | `emit` swallows observer panics silently — `_ = recover()` with no log (lifecycle.go:113-118). Isolation is right; silence is not. A panicking span emitter drops every subsequent event with zero operator signal. Log the recovered value with the event's state/transition. |
| 2 | Fix before merge | A started event has no structural guarantee of a finished counterpart — panic or early return between `Start` and `Finish` leaves a permanently open attempt/span. #1116 hand-protects exactly one of nine sites (reversewindow). This is the hole `Do(state, fn)` closes by construction. |
| 3 | Fix before merge | The "authoritative order" doc claim (workflow.go:102-104) is false under concurrency: `Start` does `state.Set` then `emit` with no serialization, and observers run outside any lock — concurrent emitters (e.g. an ErrCleanup event from the binlog-error callback) interleave. Serialize emission or weaken the doc to single-goroutine callers. |
| 4 | Small | Outcome misclassification: any `errors.Is(err, context.DeadlineExceeded)` → `Cancelled` (lifecycle.go:157-168), so a lock-wait/query timeout — a genuine failure — records as an operator cancel and nobody investigates. Suggest: `Cancelled` only when `runCtx.Err() != nil`; bare DeadlineExceeded from sub-operations is `Failed`. The table test pins the current semantics. |
| 5 | Small | `Totals` silently stripped for non-CopyRows states (lifecycle.go:150-153) — hard-codes a CopyRows-only policy inside pkg/status; a future Checksum/ApplyChangeset totals needs an API change. Move the policy to the caller. |
| 6 | Follow-up | Evidence delivery is at-most-once per process (crash-after-cutover-before-emit, or resume in a new process, emits zero or re-emits). Doc that consumers must never treat *absence* of DurableMutation as evidence the swap didn't happen. |

**Recommendation:** steer to #1118's bracketed shape and port the good parts of this PR into it. Concrete salvage list: the `WorkflowEvent` payload struct + closed `String()` vocabulary and its test; the `workflowOutcome` err→outcome derivation (after deciding item 4); the DurableMutation/Terminal CAS-once + `valid()` guard + `ResetEvidence` + dedupe tests, wholesale; panic isolation (with logging per item 1); the CopyRows `WorkflowTotals` payload from #1113; the zero-value-safety tests. What the Tracker obsoletes: `WorkflowAttempt` handles + finished-CAS, `WorkflowResult`, parent-context capture in the attempt struct, `HasObserver`, and all the Start/Finish split plumbing #1116 demonstrates the cost of.

Verified solid: attempt idempotence (double-Finish and copied-handle Finish emit exactly one finished event — CAS on the shared pointee), invalid `Terminal` values don't burn the once-guard, state written before the started event so `Get()` in a callback is consistent single-goroutine, `-race` clean.

PR #1131, opened 2026-08-14
migrate: remove --lint and --lint-only

### LINKED ISSUE #1129: remove --lint and --lint-only from `spirit migrate`
I propose we remove --lint and --lint-only from `spirit migrate`. My rationale is:

- We have duplicate functionality in `spirit lint`. It's always better to recommend one way of doing things.
- The pkg/migration/runner.go has probably 100-200 lines of code dedicated to this. It's a god-mode file and it's not straight forward to reduce its size, but we can start here and show we have some discipline.

morgo:
Fixes #1129

`spirit lint` and `spirit diff` already expose the linting engine, so carrying a second entry point inside `spirit migrate` duplicated functionality and gave users two ways to do the same thing. Removing it also trims ~120 lines out of `pkg/migration/runner.go`, which is the package's god-mode file.

**This is a backward incompatible change.** `spirit migrate --lint` and `--lint-only` are gone; use `spirit lint` or `spirit diff` instead. We don't commit to compatibility on these flags.

One behavioural difference on the replacement path is worth calling out: `migrate --lint` demoted `invisible_index_before_drop` to a warning via `defaultLinterSettings`, so a `DROP INDEX` printed a warning and the migration proceeded. `spirit lint` / `spirit diff` build a bare `Config{}`, so the linter's own default (`raiseError: true`) applies and the same drop is an **error** — exit code 1. That is now documented in `docs/lint.md` under Violation Severity.

## Removed

- `Migration.Lint` / `Migration.LintOnly` and the `--lint`/`--lint-only` mutual-exclusion check in `Validate()`
- The lint dispatch block in `Runner.Run`
- `Runner.lint()` and `Runner.getCreateTable()` — the latter was only ever called by `lint()`
- `pkg/migration/lint.go` (`defaultLinterSettings`, `printLinters`)
- `pkg/migration/lint_test.go`, the `WithLint()` / `WithLintOnly()` test helpers, and the two `TestMigrationValidate` lint cases

The `lint`, `statement`, and `sqlescape` imports all drop out of `runner.go` with it.

## Test wiring

The single-version suite selects tests with `-run 'Resume|Checkpoint|UniqueOnNonUniqueData|ChunkerPrefetching|Unparsable|Lint'`. With `lint_test.go` gone, `Lint` matches nothing, so it's dropped from the regex and the accompanying comments in `compose/compose.yml`, `.github/workflows/mysql8.0.45-singleversion-docker.yml`, and the `singleversion_test.go` header.

## Docs

- `docs/migrate.md` — `lint` / `lint-only` sections and TOC entries
- `docs/lint.md`, `docs/diff.md` — the `spirit migrate --lint` cross-references in "See Also"
- `AGENTS.md` — `WithLint()` / `WithLintOnly()` in the test-helper option list

## Verification

Kept off the shared local MySQL, so the DB-backed migration suite was not run locally — CI covers it. Nothing outside the deleted lint path changes behavior.

- `go build ./...` and `go vet ./...` clean, including `-tags singleversion`
- Test binaries compile for `pkg/migration` (both tags) and `pkg/lint`
- `golangci-lint run ./pkg/migration/...` — 0 issues
- `TestMigrationValidate` passes
- `spirit migrate --help` no longer mentions lint; `spirit lint` / `spirit diff` still register normally


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1131, bb48f07e.

**Verdict:** 4 findings — 1 non-blocking (undocumented lint-severity flip on the replacement path), 3 suggestions; CI 11/11 pass at head bb48f07e (DCO Check, 8 build matrix jobs, govulncheck, lint — verified via gh pr checks at review time; head still bb48f07e9c57f8066deca1799faf0bfa530b338b, state OPEN).

### Non-blocking

1. **The recommended replacement path is not behavior-equivalent: `invisible_index_before_drop` flips from warning to error.** The deleted `defaultLinterSettings` demoted this linter to a warning under `migrate --lint`, but `spirit lint`/`spirit diff` build a bare `Config{}` ([cmd_lint.go:70](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/cmd/spirit/cmd_lint.go#L70)), so the linter's own default `raiseError: "true"` ([lint_invisible_index.go:53](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/lint/lint_invisible_index.go#L53)) makes it SeverityError and the command exits 1. A user whose `DROP INDEX` previously warned and proceeded (exit 0) now gets a hard failure with no CLI knob and no mention in the PR body or docs; a docs note (or a settings knob) is a safe follow-up.

### General suggestions

2. **Stale doc comment on `Validate()`.** [migration.go:114](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/migration/migration.go#L114) still says the hook checks "invalid flag combinations", but the lint/lint-only mutual exclusion was the only combination check — the body now contains only five negative-value checks. Update the comment so future readers don't assume combination checks live here.

3. **Stale `TestMigrationValidate` header comment.** [migration_test.go:937](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/migration/migration_test.go#L937) claims the test covers "invalid flag combinations", but the only combination case (lint + lint-only) was removed; the table now holds only zero/typical/negative single-flag cases. Someone adding a future combination check may wrongly assume coverage exists.

4. **Pre-existing: AGENTS.md's workflow list omits the singleversion workflow this PR edits.** The CI/CD list at [AGENTS.md:325](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/AGENTS.md#L325) names 8 of 11 workflows, missing `mysql8.0.45-singleversion-docker.yml`, `mysql-semisync-docker.yml`, and `govulncheck.yml`. An agent following it when adding a singleversion-tagged test won't learn the dedicated workflow's `-run` regex needs updating — the exact hazard this PR's regex edit works around; follow-up material.

The pruned test-selection regex in the singleversion CI job: dropping `|Lint` from `-run 'Resume|Checkpoint|UniqueOnNonUniqueData|ChunkerPrefetching|Unparsable'` ([compose.yml:49](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/compose/compose.yml#L49)) could silently stop running still-live tests, since `go test` exits 0 when a `-run` pattern matches nothing. Proven safe by enumerating every file carrying `//go:build singleversion` — only [resume_test.go](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/migration/resume_test.go) and [singleversion_test.go](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/migration/singleversion_test.go) remain — and every `func Test*` in them: all 19 names match one of the five retained alternatives. A grep for `func Test.*Lint` in pkg/migration returns zero hits, so the suite lost exactly the deleted lint tests and nothing else. `go vet -tags singleversion ./pkg/migration/...` confirms the tagged tree still compiles.

### Verified correct

- No dangling references to the deleted code: `go build ./...` and `go vet -tags singleversion ./pkg/migration/...` pass on the PR head worktree.
- Removal sweep is complete: repo-wide grep for `--lint`/`WithLint`/`printLinters`/`defaultLinterSettings`/`getCreateTable` hits only unrelated pre-existing code, e.g. [create_table.go:248](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/statement/create_table.go#L248).
- The documented replacement path exists and is untouched: [main.go:26-27](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/cmd/spirit/main.go#L26-L27) still registers `lint.LintCmd`/`lint.DiffCmd`; pkg/lint is unmodified.
- Surviving `TestMigrationValidate` cases match `Validate()`'s five negative-value checks one-for-one ([migration_test.go:947-964](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/migration/migration_test.go#L947-L964)).
- Safe composition with sibling #1132 in either merge order: its runner.go hunks (connection-budget comment, RepairApplier field) are disjoint from this PR's deletions.
- AGENTS.md test-helper option list no longer mentions `WithLint()`/`WithLintOnly()` (grep returns nothing), matching the surviving helpers in [helpers_test.go](https://github.com/block/spirit/blob/bb48f07e9c57f8066deca1799faf0bfa530b338b/pkg/migration/helpers_test.go).
- CI: 11/11 checks pass at head bb48f07e (DCO Check, 8-job build matrix, govulncheck, lint).

Kiran01bm (approved):

morgo:
🤖 Thanks — all four findings addressed in 075a382c.

**1. `invisible_index_before_drop` severity flip (non-blocking).** Confirmed and worth stating outright. `migrate --lint` demoted it to a warning via `defaultLinterSettings`, so a `DROP INDEX` printed a warning and the migration proceeded; `spirit lint`/`diff` build a bare `Config{}` ([cmd_lint.go:70](https://github.com/block/spirit/blob/main/pkg/lint/cmd_lint.go#L70)), so the linter's own `raiseError: "true"` default applies, `HasErrors` trips, and the command exits 1. Documented under Violation Severity in `docs/lint.md` and called out in the PR body. Went with the docs note over a settings knob — reintroducing a config surface here would partly undo the consolidation this PR is doing.

Worth noting the blast radius is narrower than it first looks: the linter only fires on `ALTER ... DROP INDEX`, and `spirit lint` operates on a schema with no changes, so in practice this is a `spirit diff` behaviour change.

**2. Stale `Validate()` godoc** and **3. stale `TestMigrationValidate` header** — both fixed. Each now states there are no cross-flag combination checks today and points at where to add the first one, so the next person doesn't assume coverage that isn't there.

**4. AGENTS.md workflow list.** Fixed rather than deferred — it's cheap, and the gap is exactly the hazard this PR had to navigate. All three missing workflows are now listed, and the singleversion entry spells out that its `-run` regex lives in `compose/compose.yml` and that `go test` exits 0 when the pattern matches nothing, so a name mismatch silently skips the test.

Separately, the two CI failures on the first run were an unrelated flake in `TestPeriodicFlushPrioritizesParkedSubscription` — filed as #1133 with the root cause (nondeterministic `c.subs.Snapshot()` order lets the all-subscription pass drain table A twice before B, satisfying the test's `len(recorded) >= 2` gate too early). Both jobs passed on rerun.

PR #1132, opened 2026-08-14
checksum: repair mismatched chunks through the applier

### LINKED ISSUE #1130: checksum chunk replacements should go via the applier
Currently if a chunk needs replacing it using an unbuffered MySQL statement directly, causing locking.

We should convert this to use the applier interface. This will allow us to also use bigger checksum time chunks and not fear the consequences, or not need to divide chunks for recopying.

morgo:
Fixes #1130.

## The problem

`SingleChecker.replaceChunk` rewrote a mismatched chunk with a single `REPLACE INTO _new (...) SELECT ... FROM original`. Under `REPEATABLE READ` that `SELECT` is a **locking** read — it takes shared next-key locks on every source row it touches — so application writes to the *original* table block behind a repair for as long as the statement runs, and the write half holds its `_new` row locks for the same duration. Both footprints scale with the chunk, which is what made large checksum chunks risky to repair.

Confirmed against a live 8.0 server: with an uncommitted `UPDATE` holding one source row, the old statement dies with `ERROR 1205 Lock wait timeout exceeded`, while the plain `SELECT` the new path uses returns immediately.

## The change

Repair now does what `DistributedChecker.replaceChunk` and `MySQLRecopier` already do:

1. `DELETE` the chunk's key range on the target (unchanged — this is what removes rows the source no longer has).
2. Read the source rows into Spirit with a plain, non-locking consistent read.
3. Rewrite them through the **applier** — the same buffered write path the copier and the binlog apply use, which splits the write into bounded statements.

Rows are read and submitted in batches (`repairBatchRows` / `repairBatchBytes`), so client memory is bounded no matter how big the chunk is, and reads pipeline with writes.

The applier is started and stopped **around each repair** rather than for the checker's lifetime. Its workers then live under the repair's own cancellation-detached, time-bounded context, so a parent cancellation mid-repair cannot strand queued inserts, and `Stop()` joins them before `replaceChunk` returns. (This is tighter than `DistributedChecker`, which starts the applier for the whole of `Run` and documents that queued inserts may be dropped if `Run` returns early.)

## Two consequences of the applier being the write path

Both are already true of the other two repair implementations, so this makes the three consistent:

- **`INSERT IGNORE`, not `REPLACE`.** Rows inside the just-deleted range have nothing to conflict with; a row colliding on a UNIQUE *secondary* key with a row **outside** the range is now skipped rather than clobbering that row. The chunk then stays diverged, the next attempt re-flags it, and retries exhaust into the existing `checksum found differences on every attempt` error — the right outcome for a lossy ALTER such as adding a unique index to non-unique data (`TestUnfixableUniqueChecksum` still passes unchanged). Rows skipped this way are logged with counts.
- **JSON columns are read bare.** The read/write pair is already text-mediated, so a repaired row lands as exactly the one-text-round-trip image the checksum's source side predicts. Adding a cast on top would apply `parse∘render` twice, which does not converge for the doubles MySQL's JSON text parser misrounds. `ColumnMapping.RepairExprs` existed only to add that cast server-side and is removed. `TestJSONChecksumMisparsedDoubleRepairConverges` is the convergence proof and still passes.

## Wiring

New `CheckerConfig.RepairApplier`, required by the single-server checker: `NewChecker` returns `repair applier must be non-nil` rather than quietly building a second write path, so a checker that could not repair fails at construction instead of at the first mismatch. The migration runner passes the applier the copy phase used, so a repair inherits the configured write concurrency (the copier has stopped it by the time the checksum runs). The distributed checker is unaffected — it repairs through its own `Applier` — so there are no move/sync wiring changes.

## Tests

New `pkg/checksum/repair_test.go`:

- `TestRepairStreamsChunkThroughApplier` — 2048-row chunk (> `repairBatchRows`) diverged in all three ways (missing rows, wrong row, extra row); asserts an exact rewrite.
- `TestRepairDoesNotLockSourceRows` — the #1130 regression: an uncommitted `UPDATE` on a source row in the chunk must not delay the repair, and the repair must read the committed pre-image.
- `TestRepairWithColumnRename` — pins the positional source→target mapping the applier relies on.
- `TestRepairEmptySourceRange` — the DELETE alone is the repair; the applier is never started.

Full `go test ./...` green, `-race` green on the repair/JSON/fix paths, `golangci-lint ./pkg/...` clean. Docs: new "Chunk repair" section in `pkg/checksum/README.md`, and `pkg/migration/README.md`'s "no data locks on the source table" claim is now true of the repair path too.


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1132, 4d2bff4a.

**Verdict:** 8 findings — 3 non-blocking, 5 general suggestions; none blocking. CI is 11/11 pass (DCO, 8 build matrix jobs, govulncheck, lint) at head 4d2bff4a; PR still OPEN with head unchanged.

### Non-blocking

1. **Repair can resurrect a source-deleted row under delete churn.** The new plain non-locking read at [pkg/checksum/single.go:465](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L465) replaces a REPLACE INTO...SELECT whose SELECT half held shared row locks, so a row deleted on the source after the repair's read — whose binlog DELETE lands on the target before the repair's async INSERT IGNORE — is re-inserted stale, leaving the chunk diverged and burning a checksum retry. It fails safe (retries converge; no bad cutover) and the same design pre-exists in DistributedChecker, but sustained delete churn on one chunk can exhaust MaxRetries into "checksum found differences on every attempt" — worth a code comment or a follow-up mitigation.

2. **The skip-detection warning can be masked by retry-inflated rowsAffected.** RetryableTransaction accumulates `rowsAffected` across attempts without reset ([pkg/dbconn/dbconn.go:190](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/dbconn/dbconn.go#L190), `+=` at [:270](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/dbconn/dbconn.go#L270) before Commit), so a statement-success/Commit-retry double-counts and can push appliedRows past sourceRows, suppressing the "did not rewrite every source row" warning at [pkg/checksum/single.go:559](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L559). The accumulator quirk is pre-existing, but this PR adds its first consumer that compares it to an expected count — reset the counter per attempt or note the caveat.

3. **No test exercises any error branch of the new repair path.** Applier Start failure, Apply submit failure, Wait failure, and the `firstApplyErr` propagation at [pkg/checksum/single.go:556](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L556) are all untested (repair_test.go's 4 tests are happy-path only, no error injection or mid-repair cancellation). A refactor that drops the firstApplyErr check compiles and passes the suite, turning a surfaced write error into a slower retry-exhaustion failure — add at least one failure-injection test.

### General suggestions

4. **The memory-bound doc comment overstates the bound.** [pkg/checksum/single.go:323](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L323) says memory is bounded by repairBatchRows/repairBatchBytes, but Apply() is async and up to 128 chunklets of up to 1 MiB can queue in the applier's buffer ([pkg/applier/single_target.go:114](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/applier/single_target.go#L114)) — still bounded, but at ~128 MiB, not one batch. Reword here and in pkg/checksum/README.md.

5. **TestRepairEmptySourceRange doesn't assert its stated purpose.** The comment claims "the applier is never started" but the test body only asserts COUNT(*)=0 ([pkg/checksum/repair_test.go:189](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/repair_test.go#L189)), so a regression that starts the applier on an empty range stays green — a spy applier recording Start calls would pin it.

6. **No repair test exercises the restart lifecycle production uses.** The fixture comment ([pkg/checksum/repair_test.go:37](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/repair_test.go#L37)) claims it mirrors the runner's wiring, yet every test starts a fresh never-started applier, so the Start-after-Stop restart branch ([pkg/applier/single_target.go:143](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/applier/single_target.go#L143)) and two-sequential-repairs sequencing are never entered. Coverage note only — the branch pre-exists and is exercised elsewhere.

7. **Repair SELECT omits FORCE INDEX (PRIMARY) where the copier's equivalent read forces it.** [pkg/checksum/single.go:460](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L460) has no hint while [pkg/copier/buffered.go:73](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/copier/buffered.go#L73) forces PRIMARY on the same key-range predicate shape, so misleading optimizer stats could pick a worse plan and push an XL-chunk repair toward the 10-minute fixChunkTimeout. Perf/consistency only.

8. **Commit subject's trailing "(#1130)" mimics the squash-merge PR autolink but points at the issue.** Head commit 2f2cde8c ends "…through the applier (#1130)", whereas main's precedent reserves that form for the PR number GitHub appends on squash (after merge it would read "(#1130) (#1132)"). The body's "Fixes #1130" already carries the issue ref — drop it from the subject.

The riskiest mechanism is reusing the migration runner's shared applier — the instance the copier used and the binlog feed still writes through — restarted around each repair under a cancellation-detached 10-minute fixCtx. Lifecycle checks out: the copier calls applier.Stop() before the checksum phase ([pkg/copier/buffered.go:249](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/copier/buffered.go#L249)), Start's restart branch reinitializes channels and worker bookkeeping ([pkg/applier/single_target.go:143](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/applier/single_target.go#L143)), and repairs are serialized under recopyLock with defer-LIFO guaranteeing Stop completes before the lock releases ([pkg/checksum/single.go:352](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L352)). Every Apply/Start/Stop caller repo-wide was traced: the binlog feed uses only the synchronous UpsertRows/DeleteKeys ([pkg/change/subscription_buffered.go:1019](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/change/subscription_buffered.go#L1019)), which Stop does not control and Wait does not track, so a repair can neither kill nor block on feed writes. Failure paths funnel through the deferred Stop and the exactly-once callback protocol ([pkg/applier/single_target.go:268](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/applier/single_target.go#L268)), surfacing as a chunk retry with cutover gated on a fully clean pass, so the deleted-then-not-rewritten window converges or aborts — never silently loses rows. The residual risks that survive are availability/diagnostic-only: findings 1 and 2.

### Verified correct

- Cancellation-detached repair covers all writes: DELETE + read + Start + Apply + Wait run under WithoutCancel + 10-min timeout, deferred Stop joins workers on every return path ([pkg/checksum/single.go:427](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/single.go#L427)).
- Callback/result sync is race-free: Wait returns only when pendingWork is empty and callbacksInFlight is 0, so reading firstApplyErr/appliedRows after Wait races nothing ([pkg/applier/single_target.go:281](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/applier/single_target.go#L281)).
- Fail-closed constructor swept completely: "repair applier must be non-nil" ([pkg/checksum/checksum.go:301](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/checksum.go#L301)); both non-test callers set an applier, and move takes the exempted DistributedChecker branch.
- INSERT IGNORE does not swallow real errors: SHOW WARNINGS after every statement makes non-dup-key codes fatal ([pkg/dbconn/dbconn.go:242](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/dbconn/dbconn.go#L242)).
- Positional column mapping is consistent by construction (same ColumnMapping builds both lists), pinned by TestRepairWithColumnRename ([pkg/checksum/repair_test.go:163](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/repair_test.go#L163)).
- RepairExprs removal is a complete sweep; its JSON round-trip invariant is re-established and pinned by TestJSONChecksumMisparsedDoubleRepairConverges ([pkg/checksum/json_test.go:219](https://github.com/block/spirit/blob/4d2bff4a752c935eda38dfe78632e73788518e07/pkg/checksum/json_test.go#L219)).
- Composition with sibling #1136 is safe: overlapping files but fully disjoint hunks/symbols, pairwise merge-tree clean in either order; both READMEs match the implementation.
- Build/vet/lint clean per finders; CI 11/11 pass (DCO, 8 build matrix jobs, govulncheck, lint) at head 4d2bff4a.

Kiran01bm (approved):

PR #1135, opened 2026-08-14
change: de-flake TestPeriodicFlushPrioritizesParkedSubscription

### LINKED ISSUE #1133: flaky test: TestPeriodicFlushPrioritizesParkedSubscription (all-subscription pass drains A twice before B)
`TestPeriodicFlushPrioritizesParkedSubscription` fails roughly half the time on a nondeterministic map-iteration order. Seen on PR #1131 (a docs/flags cleanup that touches nothing in `pkg/change`), where it took down both the MySQL 8.4 GA and MySQL 8.0.42 jobs at the same assertion:

```
--- FAIL: TestPeriodicFlushPrioritizesParkedSubscription (0.43s)
    Error Trace: /app/pkg/change/subscription_buffered_concurrency_test.go:612
    Error:       []string{"..._caacd0_a", "..._caacd0_a"} does not contain "..._caacd0_b"
```

Runs: [8.4 GA](https://github.com/block/spirit/actions/runs/31811038658/job/94801452104), [8.0.42](https://github.com/block/spirit/actions/runs/31811038706/job/94801452209).

## Cause

On a park signal, `runPeriodicFlush` drains the parked subscription first, then runs the all-subscription pass (`pkg/change/binlog.go:1245`). That pass iterates `c.subs.Snapshot()` **serially in nondeterministic map order** (`pkg/change/binlog.go:1102`).

Table A still has a buffered row after the priority flush — the `'parked'` insert is buffered once the park releases — so the pass drains A a second time. The test gates on a bare count:

```go
require.Eventually(t, func() bool {
    return len(rec.recorded()) >= 2
}, 15*time.Second, 50*time.Millisecond, "park-requested flush did not reach the applier")
order := rec.recorded()
require.Equal(t, srcA.TableName, order[0], "...")
require.Contains(t, order, srcB.TableName, "...")
```

When the snapshot order is `[A, B]`, the recorder hits two entries — both `A` — the gate fires, and the snapshot is taken before B is ever drained, so the trailing `Contains` fails. When the order is `[B, A]`, it passes. Nothing about the production behaviour under test is wrong; only the test's stopping condition is.

## Suggested fix

Gate on B appearing rather than on a count:

```go
require.Eventually(t, func() bool {
    return slices.Contains(rec.recorded(), srcB.TableName)
}, 15*time.Second, 50*time.Millisecond, "the all-subscription pass must still drain the other subscription")
order := rec.recorded()
require.Equal(t, srcA.TableName, order[0], "the parked subscription must be flushed before the all-subscription pass")
```

That keeps both properties the test cares about — A is flushed first (the priority), and B is eventually drained — and makes the trailing `require.Contains` redundant.

Introduced with the test in #1124.

morgo:
Fixes #1133.

`TestPeriodicFlushPrioritizesParkedSubscription` gated on `len(rec.recorded()) >= 2`, which assumed the two recorded flushes were A (the priority flush on the park signal) and B (the follow-up all-subscription pass). They need not be.

`flush` walks `c.subs.Snapshot()` (`pkg/change/binlog.go:1102`), and `Snapshot` builds its slice by ranging a map (`pkg/change/subscription_registry.go:56`), so the pass visits the registry in nondeterministic order. Table A still has a buffered row when the pass starts — the `'parked'` insert buffers once the park releases — so when the order is `[A, B]` the pass drains A a second time, the count reaches two, and `rec.recorded()` is snapshotted before B is ever visited. The trailing `require.Contains` then fails on `[A, A]`:

```
Error: []string{"..._caacd0_a", "..._caacd0_a"} does not contain "..._caacd0_b"
```

Roughly half the runs. It took down both the 8.4 GA and 8.0.42 jobs on #1131, a docs/flags change that touches nothing in `pkg/change`.

## Fix

Gate on B appearing rather than on a bare count:

```go
require.Eventually(t, func() bool {
    return slices.Contains(rec.recorded(), srcB.TableName)
}, 15*time.Second, 50*time.Millisecond, "the all-subscription pass must still drain the other subscription")
order := rec.recorded()
require.Equal(t, srcA.TableName, order[0], "the parked subscription must be flushed before the all-subscription pass")
```

That still pins both properties the test cares about — A is flushed first (the priority), and B is eventually drained by the all-subscription pass — and makes the old trailing `require.Contains` redundant. Test-only; production behaviour is unchanged.

## Testing

`go vet` clean. The fixed test passes 10/10 under `-count=10`.

I also ran the pre-fix version 20 times locally to reproduce the flake and it passed 20/20 — the failure needs A's `'parked'` row to still be buffered when the pass runs, which local timing doesn't produce. So the local runs confirm the new gate is stable but don't independently reproduce the CI failure; the diagnosis rests on the code reading above plus the two CI runs linked in #1133.

### DIFF
--- a/pkg/change/subscription_buffered_concurrency_test.go
+++ b/pkg/change/subscription_buffered_concurrency_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -603,13 +604,16 @@ func TestPeriodicFlushPrioritizesParkedSubscription(t *testing.T) {
 		return subA.timesParked.Load() >= 1
 	}, 10*time.Second, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
 
-	// The priority flush drains A; the follow-up pass drains B.
+	// The priority flush drains A; the follow-up pass drains B. Gate on B
+	// specifically rather than on a bare count: the all-subscription pass
+	// walks the registry in map order, so it can drain A a second time
+	// (the 'parked' insert buffers once the park releases) before it ever
+	// reaches B, and a count of two would then snapshot [A, A].
 	require.Eventually(t, func() bool {
-		return len(rec.recorded()) >= 2
-	}, 15*time.Second, 50*time.Millisecond, "park-requested flush did not reach the applier")
+		return slices.Contains(rec.recorded(), srcB.TableName)
+	}, 15*time.Second, 50*time.Millisecond, "the all-subscription pass must still drain the other subscription")
 	order := rec.recorded()
 	require.Equal(t, srcA.TableName, order[0], "the parked subscription must be flushed before the all-subscription pass")
-	require.Contains(t, order, srcB.TableName, "the all-subscription pass must still drain the other subscription")
 }
 
 // totalUpsertedRows sums the rows across all recorded UpsertRows calls.


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1135, e8f7e01f.

**Verdict:** 1 finding — suggestion only, no blocking or non-blocking issues; CI 11/11 pass at head e8f7e01f (DCO Check, 8 build-matrix jobs, govulncheck, lint), PR still OPEN at the reviewed SHA.

### General suggestions

1. **Detection of a removed priority flush stays probabilistic (~50% per run) — pre-existing, not introduced here.** If someone deletes the parked-first flush at [binlog.go:1255](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/binlog.go#L1255), the fallthrough `c.flush` drains in `Snapshot` map order ([subscription_registry.go:60](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_registry.go#L60)), so `order[0] == srcA.TableName` at [subscription_buffered_concurrency_test.go:616](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L616) still passes ~half the time by luck — the test's own comment at [:547](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L547) acknowledges the coin flip, and the 8 build-matrix jobs amplify per-CI-run detection to ~99.6%. A deterministic detector would need a seeded/ordered snapshot — reasonable follow-up, out of scope for a de-flake PR.

The risk was that dropping the `len >= 2` count gate leaves the surviving `require.Equal(order[0], srcA.TableName)` merely likely rather than deterministic — trading one flake for another. Verified deterministic four ways: every `c.subs.Snapshot()` call site in [binlog.go](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/binlog.go#L1102) was enumerated, and with the 1-hour ticker ([test:587](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L587)) never firing, the only flush drivers in the assertion window are the parked-priority flush ([binlog.go:1255](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/binlog.go#L1255)) then the all-pass ([:1265](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/binlog.go#L1265)), sequential on one goroutine. A's buffer is provably non-empty at priority-flush time (`require.Positive(subA.sizeBytes)` at [test:598](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L598); the park via `requestFlush` at [subscription_buffered.go:526](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered.go#L526) precedes the triggering row's buffering). `orderRecordingApplier` appends before delegating ([test:529](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L529)), so even an applier error records A first, and the `require.Eventually` gate on Contains(B) at [:613](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L613) FailNows on timeout, so `order` is non-empty before indexing. All four hold: the new assertions are deterministic given a healthy applier, and the de-flake is sound.

### Verified correct

- PR-body diagnosis matches the code: `c.flush` walks `Snapshot()` which ranges a Go map (nondeterminism documented as intentional at [subscription_registry.go:51-55](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_registry.go#L51-L55)) — the flake is real and the fix is at the right altitude (test gate, not flush-order determinism).
- Both original invariants preserved: the deleted `require.Contains(order, srcB.TableName)` is exactly subsumed by the new Eventually gate at [test:613](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L613), and priority is still proven by `order[0]` at [:616](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L616).
- No empty-slice panic and no data race: Eventually guarantees `len(recorded()) >= 1` before indexing, and `recorded()` returns a mutex-guarded copy ([test:534-538](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L534-L538)).
- No false-positive Contains match: srcA/srcB table names carry distinct unique `_a`/`_b` suffixes ([test:551](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L551)) with exact-string equality in `slices.Contains`.
- New Eventually bounds (15s / 50ms) match the neighboring single-subscription test's convention ([test:510-515](https://github.com/block/spirit/blob/e8f7e01f7775e2e8198a6212b50708b86bb1862a/pkg/change/subscription_buffered_concurrency_test.go#L510-L515)).
- Composition safe: test-only change, zero file overlap with sibling PRs, merge-tree clean vs block/main, no stale references to the old gate or failure message remain (grep clean).
- Build/test/CI: `go build`/`go vet` clean and golangci-lint 0 issues at head; CI 11/11 pass at e8f7e01f (DCO Check, 8 build-matrix jobs, govulncheck, lint).

Kiran01bm (approved):

PR #1136, opened 2026-08-14
status: consolidate the recurring log lines into one status report

### LINKED ISSUE #329: Consolidate common log lines
When spirit runs, there are 3 common log lines that keep being printed:
```
2024-07-08T23:51:17.639 info finished periodic flush of binary log: total-duration=2.854617ms batch-size=5583
2024-07-08T23:51:47.636 info migration status: state=copyRows copy-progress=273732111/3306757995 8.28% binlog-deltas=0 total-time=10h57m32s copier-time=10h57m30s copier-remaining-time=125h32m28s (-18h6m from 10h56m ago) copier-is-throttled=false conns-in-use=2
2024-07-08T23:51:47.636 info checkpoint: low-watermark={"Key":["id"],"ChunkSize":6452,"LowerBound":{"Value": ["2876218631"],"Inclusive":true},"UpperBound":{"Value": ["2876225083"],"Inclusive":false}} log-file=mysql-bin-changelog.210394 log-pos=1852984531 rows-copied=273688043 rows-copied-logical=273732111
```

To give these tasks names, they are roughly:
- The replClient status
- The copier or main task status
- The checkpoint status

These all run on different intervals, which is why they currently print separately.

What I suggest is saving the last replClient update and last checkpoint update to the migration runner state, and then printing the output in the same log line as the main task status.

The main task can then say:
- `lastCheckpoint: 10s ago` (we don't need the additional info)
- `lastChangelogFlush: 10s ago; took 2.5s; 5512 batch size`
- the `chunkSize` should also be written to the copier's status, since currently it's part of the checkpoint output.

I think `conns-in-use` can also be removed.

We might change the output interval from every 30s to every 1 minute too.

morgo:
Fixes #329.

A run used to emit three recurring INFO lines on three different schedules — the status line, the checkpoint dumper, and the change feed's periodic flush — plus a `rotate to next binlog` line from go-mysql on every rotation. This folds all of it into one report, and then makes that report readable.

### The report

```
2026/08/14 15:07:24 INFO migration status: state=copyRows total-time=2m0s copier-time=2m0s
  copier   30.27%  4954686/16370180  chunk-size=92220  eta=3m54s  throttled=false
  applier queue=34/128  workers=4  wait-p50=260ms  write-p50=28ms  write-p90=47ms
  binlog  deltas=0  rotations=67 (0 forced)  flushed 0s ago (took 3µs, 0 rows)
  ckpt    20s ago  BLKD5234QXY40-bin.057295:125584236

2026/08/14 15:08:24 INFO migration status: state=checksum total-time=3m0s checksum-time=5s
  checksum   6.75%  1104410/16370180  chunk-size=17083  threads=4  throttled=false
  binlog   deltas=0  rotations=76 (0 forced)  flushed 5s ago (took 2µs, 0 rows)
  ckpt     30s ago  BLKD5234QXY40-bin.057312:1030034417
```

A single flat line of twenty-odd `key=value` fields is dense but not readable: a value that changes width — `chunk-size` 8097 → 92220, `write-p50` 31ms → 115ms — shifts every field after it, so the eye can't follow one number down a scrollback, which is what a 30-second heartbeat is for. Grouping by subsystem and padding the labels fixes the columns.

Which rows appear depends on the state; `checksum` replaces `copier` during that phase, and `waitingOnSentinelTable` gets a `sentinel` row. All three runners (migrate, move, sync) build the same block.

### What moved onto it

| Was | Now |
| --- | --- |
| `INFO checkpoint low-watermark=… position=…` every 50s | `ckpt` row (age + position); the detail is still logged at DEBUG |
| `INFO finished periodic flush of binary log` every 30s | `binlog` row's `flushed X ago (took Y, n rows)`; demoted to DEBUG |
| `INFO rotate to next binlog` from go-mysql, per rotation | `binlog` row's `rotations=n (m forced)`; demoted to DEBUG via a log wrapper |

The rotation counter deliberately preserves what that noisy line was good for. `rotations` is what the source is doing; `(n forced)` is the subset spirit caused itself by issuing `FLUSH BINARY LOGS` from `BlockWait`, which is the "is cutover-time waiting churning through binlogs?" signal. Duplicate rotate events (the server sends a real one and an artificial one at the same position) are counted once.

`checkpoint-position` is new: `since-checkpoint` alone tells you how stale the resume point is but not whether it is still *usable*. Resuming only works while that position is still on the server, so a run with heavy rotation and short binlog retention can become unresumable long before anything fails. Read the two together, against `rotations`.

`chunk-size` was previously visible only inside the checkpoint line's watermark JSON, which is no longer logged at INFO. The checksum phase now reports its own — it sizes chunks dynamically too.

### What came off it

- **`conns-in-use`** — reported `sql.DB` pool occupancy, which tracks the configured thread count and says nothing an operator acts on. Also removed from the `apply complete` summary.
- **The relative ETA comparison** (`copier-remaining-time=13m (-2m from 1m ago)`) — more confusing than useful in practice. `copier_eta_history.go` deleted.
- **Four applier fields** — `pending`, `rows-per-chunklet`, `queue-wait-p90`, and (conditionally) `build-p50` / `handoff-p50`. The last two now render only when they carry a diagnosis: build ≥25% of write means client-CPU bound, handoff ≥1ms means blocked behind the completion path (#1097). Both are silent on a healthy run, so their *presence* is the signal. Every field stays on `Stats` and in the metrics sink, which is what dashboards should read.

### Notable in review

- **This is one log record containing newlines.** The default slog handler (what the CLI uses) prints it as written; a handler that quotes messages — `TextHandler`, JSON — will escape them onto one line. That is the main trade-off of the format.
- **Field prefixes are gone** (`applier-queue=` → `queue=` under the `applier` label), so anything grepping the log text needs updating. Metrics are unchanged.
- **`change.Source` is untouched.** Feed stats arrive through a new optional `StatsReporter` interface, so out-of-tree sources (strata's `pkg/vstream`) keep compiling and simply contribute nothing to the row.
- **`copier.Copier` gains `CopyProgress()` and `ChunkSize()`**, and `checksum.Paced` gains `ChunkSize()`. Chunk size is sampled from the claimed chunk rather than read off the chunker, so it works for every `Chunker` implementation including the multi-table one.
- `WatchTask` no longer logs a bare `INFO` with an empty message for states whose `Status()` is empty.
- The commits are split so the format is revertible on its own: consolidation → checkpoint position + applier trim → block layout → bars removed.

### Testing

`go build ./...`, `make lint` (0 issues), and the full test suites for pkg/status, applier, change, copier, checksum, migration (including `-tags singleversion`), move and datasync. New unit tests cover the block renderer, the log-demoting handler, `LastCheckpoint`, `FeedStats` (including against a live feed: real flushes, real rotations), and the applier exception fields. Also exercised end-to-end against a local MySQL with a 16M-row `ALTER` — the output above is from that run.

Docs: a new "Reading the status output" section in `docs/migrate.md` explaining every row, plus `pkg/status/README.md`, `pkg/applier/README.md` and `pkg/copier/README.md`.


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/1136, ce9c43f4.

**Verdict:** 8 findings — 1 blocking (doc phase order), 3 non-blocking, 4 suggestions; CI 11/11 pass at ce9c43f4 (8 build jobs, DCO Check, govulncheck, lint).

### Blocking

1. **New status doc states the wrong migrate phase order.** [docs/migrate.md:685](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/docs/migrate.md#L685) places `waitingOnSentinelTable` right after `copyRows` and includes `restoreSecondaryIndexes`, but the enum and runner put the sentinel wait after `postChecksum`, immediately before cutover ([pkg/status/state.go:20-33](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/status/state.go#L20-L33), [pkg/migration/runner.go:445](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/migration/runner.go#L445)), and migrate never enters `restoreSecondaryIndexes` (move-only). An operator using `--defer-cutover` will mistime dropping the sentinel table — their manual gate before cutover; fix the ordering, drop the phantom phase, and note the state fires whenever sentinel respect is on (default), not only with `--defer-cutover`.

### Non-blocking

2. **gtidClient's rotation counter over-counts by one at stream connect.** [pkg/change/gtid.go:534](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/gtid.go#L534) initializes `currentLogName := ""`, so the server's artificial RotateEvent at dump start always counts, whereas binlogClient deliberately seeds from `flushedPos.Name` ([pkg/change/binlog.go:567](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/binlog.go#L567)) to dedup exactly this event. A GTID-mode run on an idle server shows `rotations=1 (0 forced)` with zero real rotations, permanently one high; seed the same way.

3. **The ApplyChangeset/PostChecksum status blocks omit the ckpt row while checkpoints keep being written.** [pkg/migration/runner.go:1956](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/migration/runner.go#L1956) (and [pkg/move/runner.go:1443](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/move/runner.go#L1443)) render only applier and binlog rows, yet the dumper still writes every 50s in those states and the per-dump line is now DEBUG. During a long delta-drain under heavy rotation the operator can no longer see the resume position falling toward the binlog-retention cliff — the PR's own resumability signal; add the ckpt row to that case.

4. **gtidClient's rotation counting, FeedStats plumbing, and BlockWait log gating have zero test coverage.** `gtid_test.go` is untouched and the live-feed test exercises binlogClient only, with the rotation assertion just a lower bound ([pkg/change/stats_test.go:189](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/stats_test.go#L189)). If the dedup at [pkg/change/gtid.go:683](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/gtid.go#L683) breaks, GTID-mode rotations silently double-count or read 0; a unit-level replay of a real+artificial rotate pair would pin it.

### General suggestions

5. **A migration is now completely silent at INFO during analyzeTable, and the empty-Status guard is untested.** The new skip at [pkg/status/task.go:54](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/status/task.go#L54) plus the missing AnalyzeTable case in the Status switch ([pkg/migration/runner.go:1986](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/migration/runner.go#L1986)) and the checkpoint demotion mean a blocked ANALYZE TABLE emits nothing, tripping log-based liveness checks; consider a minimal `state=analyzeTable` line, and a test for the guard.

6. **FeedStats.Flushes/Residual are populated, merged, and documented but never rendered or consumed.** [pkg/change/stats.go:31](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/stats.go#L31) declares them and [:98](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/stats.go#L98) sums them, but `String()` never prints them and the autoscaler reads `Source.FlushResidual` directly; drop them or render them.

7. **build-p50 lacks an absolute noise floor, so a healthy fast pipeline renders the self-contradictory `build-p50=0s`.** [pkg/applier/stats.go:136](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/applier/stats.go#L136) gates only on share-of-write-p50, so sub-millisecond values (e.g. 300µs build vs 800µs write) trigger the "client-CPU bound" exception and round to 0s; add a floor like handoff's at [:139](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/applier/stats.go#L139).

8. **datasync's copyRows block omits the applier row and `throttled=`, diverging from migration/move.** [pkg/datasync/runner.go:1507](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/datasync/runner.go#L1507) skips both while migration renders them ([pkg/migration/runner.go:1936-1938](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/migration/runner.go#L1936-L1938)), so write-side saturation is invisible for exactly one runner; if deliberate, a comment citing the runner-triplet rule would help.

The demoting slog handler wrapped around go-mysql's syncer logger ([pkg/change/loghandler.go](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/loghandler.go#L63)) fronts every line the replication feed emits, so a bug there either floods the log or silently swallows reconnect/error WARNs. Demotion is an exact-message map lookup against the single literal `rotate to next binlog` ([:22](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/loghandler.go#L22)), matching the pinned dependency's only emission site, so nothing else can be demoted. The slog contract holds: `Enabled` returns the union so demotable INFO records still reach `Handle`, which rewrites the level only on a match and re-gates on `inner.Enabled` ([:59-70](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/loghandler.go#L59-L70)), meaning unmapped WARN/ERROR pass untouched and demoted records survive at DEBUG rather than vanishing; `WithAttrs`/`WithGroup` rewrap so derived loggers keep the behavior. All loghandler tests plus `TestSyncerConfigUsesDemotingLogger` (proving both clients install the wrapper) pass under `-race` at head.

### Verified correct

- binlogClient rotation dedup: seeded from `flushedPos.Name` under the mutex, so connect-time and synthetic rotates count zero ([pkg/change/binlog.go:566-568](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/binlog.go#L566-L568)).
- StatusRow multi-source merge: counters summed, stalest flush wins, never-flushed dominates ([pkg/change/stats.go:93-119](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/stats.go#L93-L119)); order-independence tested.
- Forced-rotation attribution is exact: `flushedBinlogs` increments only after spirit's own `FLUSH BINARY LOGS` ([pkg/change/binlog.go:1370-1373](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/change/binlog.go#L1370-L1373)), so forced ≤ rotations by construction.
- LastCheckpoint stores position before timestamp, so a fresh age never pairs with a stale position ([pkg/status/lastcheckpoint.go:36-40](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/status/lastcheckpoint.go#L36-L40)).
- Checkpoint-write-failure-is-fatal invariant preserved: errors still wrap `ErrCouldNotWriteCheckpoint` and the dumper still cancels ([pkg/status/task.go:101-103](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/status/task.go#L101-L103)).
- Zero-guards on the new renderers: build-share requires `WriteTimeP50 > 0` and progress guards `total==0` ([pkg/applier/stats.go:136](https://github.com/block/spirit/blob/ce9c43f47b8cb65cc5a34f76cf85d54747c8fa03/pkg/applier/stats.go#L136)); pinned by `TestStatsStringExceptionFields`.
- Removed-behavior sweep complete: no stray references to `conns-in-use`, `copy-progress=`, `binlog-deltas`, or `copier-remaining-time` remain outside deliberate doc updates.
- Local unit runs at head (`pkg/status`, loghandler, FeedStats/StatusRow, applier stats, all `-race`) pass; CI 11/11 green at ce9c43f4.

Kiran01bm (approved):

morgo:
🤖 Thanks — all eight addressed in cbee8299.

**1. Doc phase order (blocking)** — fixed. The table now reads `copyRows → applyChangeset → analyzeTable → checksum → postChecksum → waitingOnSentinelTable → cutOver`, `restoreSecondaryIndexes` is gone (move/sync only), and the row notes the sentinel wait is entered whenever the sentinel is respected — the default — with `--defer-cutover` deciding only whether it blocks.

**2. GTID rotation over-count** — fixed. `binlogClient` seeds from `flushedPos.Name`; the GTID client resumes by GTID and has no file name to seed from, so the dedup now lives in `gtidClient.countRotation`: an empty `currentLogName` means "nothing seen yet" and the first rotate only seeds the comparison. That also covers `recreateStreamer` re-opening the current file.

**3. Missing ckpt row during the drain** — fixed in both runners. `applyChangeset`/`postChecksum` now render the ckpt row, since the dumper keeps writing there and that is exactly when the resume position can walk toward the retention cliff.

**4. GTID coverage** — added `TestGTIDCountRotation` (seed event, real+artificial pair, streamer re-open, empty name) and `TestGTIDFeedStats` (plumbing plus `ForcedRotations` staying zero). Both are unit-level, no server needed.

**5. Silent analyzeTable** — fixed. `analyzeTable` now renders a minimal block (header + binlog + ckpt) in migrate, and move does the same for `analyzeTable`/`restoreSecondaryIndexes`. Also added `TestContinuallyDumpStatusSkipsEmpty`, which pins both halves of the guard: an empty status logs nothing, a non-empty one still logs.

**6. Unrendered FeedStats fields** — dropped `Flushes`/`Residual`. Nothing rendered them and the autoscaler reads `Source.FlushResidual` directly; the test that used them now asserts through that accessor.

**7. build-p50 with no floor** — added `buildNoiseFloor = 1ms` alongside the share test, so 300µs-build/800µs-write no longer renders `build-p50=0s`. Test and docs updated.

**8. datasync divergence** — the applier row was missing because the field was assigned outside `progMu`; it is now published under it and rendered in both `copyRows` and `applyChangeset`. `throttled=` stays off, with a comment: a sync copies through a Noop throttler, so the field would be a constant `false`.

`go build`, `go vet`, `make lint` (0 issues) and the status/applier/change/copier/checksum/migration/move/datasync suites are green locally, including `-tags singleversion`.

PR #1137, opened 2026-08-14
feat(status): expose authoritative workflow events

jonwinton:
## Summary

Replace runner-specific timing and inferred progress with one lifecycle contract owned by `status.Tracker`.

- `Tracker.Do(ctx, state, fn)` brackets every real migration, move, and datasync attempt and emits synchronous started/finished events.
- Observers receive explicit durable-mutation and terminal-ownership evidence independently from phase completion.
- Copy completion reports settled rows and chunks once from authoritative copier counters.
- Result-bearing forward-cutover callbacks preserve mutation truth even when a later error is returned.
- Ambiguous direct DDL, cutover callbacks, source renames, and reverse renames stop retrying and require manual ownership recovery.
- Reverse-window ownership is persisted so resumed runs retain authoritative terminal and durable-mutation semantics.

Observers are optional. With no observer attached, runners retain their existing behavior and do not query completed-work totals. The tracker remains zero-value ready; observer delivery is synchronous and panic-isolated, and callback state is copied before invocation so observers can safely replace or remove themselves.

## Design

This deliberately does not add a second workflow engine beside Spirit's status machine. `status.Tracker` remains the sole source of current state and timing; observation is a thin view of transitions and durable facts. Terminal evidence has its own event because reverse-finalized or ownership-ambiguous outcomes can be established after the final phase bracket closes.

Mutation and ownership ambiguity remain distinct facts. A result can require manual recovery without claiming that a durable write was confirmed.

## Rebased on main (2026-08-16)

This branch has been rebased onto `main`; two slices are no longer part of it because they landed independently:

- The retry row-accounting fix (`fix(dbconn): return committed retry row count`) merged as #1145 — commit dropped.
- The unbuffered copier was removed in #1140, so `feat(copier): expose completed work totals` now only touches the buffered copier. Its counters are also recorded from `CopyChunk`, the incremental copy entry point added since this branch was written, so the totals stay authoritative on both paths.

`copier.CompletedWorkReporter` is still an optional capability rather than a method on `Copier`. With one built-in implementation left it could be folded into the interface — happy to do that if you prefer it.

## Verification

- `go build ./... && go vet ./...`
- `go test -race -count=1 ./pkg/status ./pkg/copier ./pkg/sentinel`
- `MYSQL_DSN=... go test -race -count=1 ./pkg/migration ./pkg/move ./pkg/datasync` — green except `TestBufferedMigrationFailsGracefullyWithMinimalRBR` / `TestMoveFailsGracefullyWithMinimalRBR`, which fail identically on unmodified `main` locally (the sandbox user lacks `SESSION_VARIABLES_ADMIN`).
- `golangci-lint run` on the touched packages: 0 issues.

Supersedes the observer/lifecycle approach in #1111 and the runner-local timing design merged in #1118. It also consolidates the correctness slices previously proposed in #1112 through #1116.

### DIFF
--- a/pkg/copier/buffered.go
+++ b/pkg/copier/buffered.go
@@ -28,20 +28,22 @@ import (
 type buffered struct {
 	sync.Mutex
 
-	applier       applier.Applier
-	chunker       table.Chunker
-	concurrency   int
-	rowsPerSecond atomic.Uint64
-	chunkSize     atomic.Uint64 // size of the most recently claimed chunk; see Copier.ChunkSize
-	isInvalid     atomic.Bool
-	errMu         sync.Mutex // guards firstErr
-	firstErr      error      // first error that invalidated the copy (any goroutine)
-	startTime     time.Time
-	throttler     throttler.Throttler
-	dbConfig      *dbconn.DBConfig
-	logger        *slog.Logger
-	metricsSink   metrics.Sink
-	autoscale     AutoscaleConfig
+	applier         applier.Applier
+	chunker         table.Chunker
+	concurrency     int
+	rowsPerSecond   atomic.Uint64
+	chunkSize       atomic.Uint64 // size of the most recently claimed chunk; see Copier.ChunkSize
+	isInvalid       atomic.Bool
+	errMu           sync.Mutex // guards firstErr
+	firstErr        error      // first error that invalidated the copy (any goroutine)
+	completedRows   atomic.Uint64
+	completedChunks atomic.Uint64
+	startTime       time.Time
+	throttler       throttler.Throttler
+	dbConfig        *dbconn.DBConfig
+	logger          *slog.Logger
+	metricsSink     metrics.Sink
+	autoscale       AutoscaleConfig
 
 	// Read-worker pool management, symmetric with the applier's write-worker
 	// pool (SetWriteWorkers/ActiveWriteWorkers). concurrency above is the
@@ -111,6 +113,7 @@ func (c *buffered) CopyChunk(ctx context.Context, chunk *table.Chunk) error {
 		}
 		totalTime := time.Since(startTime)
 		c.chunker.Feedback(chunk, totalTime, uint64(affectedRows))
+		c.recordCompletedChunk(uint64(affectedRows))
 		if metricsErr := c.sendMetrics(ctx, totalTime, chunk.ChunkSize, uint64(affectedRows)); metricsErr != nil {
 			// Metrics failures don't fail the copy; log and continue.
 			c.logger.Error("error sending metrics from copier", "error", metricsErr)
@@ -254,6 +257,8 @@ func (c *buffered) Run(ctx context.Context) error {
 	c.Lock()
 	c.startTime = time.Now()
 	c.Unlock()
+	c.completedRows.Store(0)
+	c.completedChunks.Store(0)
 	go c.estimateRowsPerSecondLoop(ctx) // estimate rows while copying
 
 	// Start the applier
@@ -426,6 +431,7 @@ func (c *buffered) readWorker(ctx context.Context, quit <-chan struct{}) error {
 			totalTime := time.Since(chunkStartTime)
 			c.logger.Debug("readWorker chunk is empty, sending immediate feedback", "chunk", chunk.String())
 			c.chunker.Feedback(chunk, totalTime, 0)
+			c.recordCompletedChunk(0)
 
 			// Send metrics for empty chunk
 			err := c.sendMetrics(ctx, totalTime, chunk.ChunkSize, 0)
@@ -467,6 +473,7 @@ func (c *buffered) readWorker(ctx context.Context, quit <-chan struct{}) error {
 
 			// Send feedback to chunker with total processing time
 			c.chunker.Feedback(capturedChunk, totalTime, uint64(affectedRows))
+			c.recordCompletedChunk(uint64(affectedRows))
 
 			// Send metrics with total processing time
 			metricsErr := c.sendMetrics(ctx, totalTime, capturedChunk.ChunkSize, uint64(affectedRows))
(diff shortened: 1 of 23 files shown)


morgo:
I have an alternative proposal for the workflow state recording here:
https://github.com/block/spirit/pull/1150

morgo (commented):
🤖 Review posted by Morgan's AI agent.

Reviewed at `1bdf59f2` (rebased onto current main). Focusing on the workflow/observer half; the ownership fixes are addressed separately at the bottom.

## The observer model

The PR bundles two kinds of information that have different consumers, and I think that's what makes the machinery feel heavier than the problem:

1. **Phase telemetry** — which phase, when, how long, how much was copied. The consumer is a dashboard.
2. **Ownership evidence** — durable mutation happened, ownership is ambiguous, reverse finalized. The consumer is the caller deciding whether it may retry.

Both are delivered through the same `WorkflowEvent` stream, so the stream has to be general enough for both, and every runner has to be wired for it whether or not anyone is listening.

**Telemetry already has a home.** `metrics.Sink` is always present, defaults to `NoopSink`, and is caller-supplied. #1150 reports phases there: entry emits the phase, close emits the phase plus its duration. The concrete difference in invasiveness is that `Tracker.Do` needs no context, so #1150 changes **zero** `Do` call sites, against **20** in this PR (7 migration, 8 move, 1 reversewindow, 4 datasync). The context is threaded through `Do` purely so an observer can be handed one.

**Ownership evidence doesn't need an event at all.** It is the answer to "may I retry?", and the caller is already holding an error. `errors.Is(err, status.ErrOwnershipAmbiguous)` delivers it with no delivery mechanism, no ordering contract, and no lifecycle. The happy path needs nothing: if the move succeeded, ownership isn't ambiguous. And a reverse cutover already calls the caller's own `reverseCutoverFunc`, so "we rolled back" is not news to them.

That split removes `WorkflowEvent`, `WorkflowObserver`, `SetObserver`/`HasObserver`, the `atomic.Pointer` slot, panic isolation, the three once-per-run latches, and the `Do` signature change — while keeping every correctness fix in this PR.

### Specific things in the current implementation

- **`pkg/status/tracker.go:248` swallows observer panics silently.** `defer func() { _ = recover() }()` with no logging means a panic in an observer — or in spirit's own code reached through one — disappears without a trace. If panic isolation is kept, it should at least log. (Also, `notifyWorkflowObserver(observer, ctx, event)` takes the context second.)

- **`pkg/status/tracker.go:102` emits *started* between `enter` and `fn`,** so a slow observer's latency is attributed to the phase it is reporting on. Delivery is documented as synchronous, so this is a real coupling: observation changes the numbers being observed.

- **`WorkflowEvent` is an untagged union.** State events set `State`+`Transition`; completed-work events set `State`+`TotalsAvailable` and leave `Transition` zero; evidence events set neither. The doc comment has to spell out which field combinations are legal, and an observer has to reconstruct the discriminator. The completed-work event is the awkward case: it carries `State: CopyRows` but arrives *after* `CopyRows` finished (the defer in `runCopy` is registered before `Do`), so a state machine keyed on `State` has to special-case it.

- **`pkg/migration/runner.go:248` and `pkg/move/runner.go:226` fork behavior on `HasObserver()`.** The runner queries completed work only when someone is watching. That's a deliberate cost saving, but it means the observed and unobserved paths aren't the same code, which is the property you most want from an optional feature.

- **Layering:** `WorkflowTerminalOwnershipReverseFinalized` is a move-only concept, and it now lives in `pkg/status`. The tracker is otherwise a state-and-timing primitive; teaching it move's rollback vocabulary makes it the place where every runner's domain concepts accumulate.

- **Unrelated change riding along:** the `ctx.Err()`/`!chunker.IsRead()` check inside `runCopy` (both runners, added in `444b8476`) converts "copier returned nil while the context was cancelled and the table isn't fully read" into an error. That looks like a genuine fix, but it changes migration and move failure behavior, isn't mentioned in the description, and has nothing to do with observation. It deserves its own commit and its own test.

## Copy totals: the copier shouldn't keep this ledger

`buffered.completedRows`/`completedChunks` duplicate a count the chunker already keeps. `chunkerComposite.Feedback` (`pkg/table/chunker_composite.go:345`) already does exactly what `recordCompletedChunk` does:

```go
atomic.AddUint64(&t.rowsCopied, actualRows)
t.chunksCopied.Add(1)
```

and `Chunker.Progress()` already exposes it, with `multiChunker` summing across children. That deletes the counters, `recordCompletedChunk`, `CompletedWorkReporter`, and the `CompletedWork(c Copier)` capability assert — which is now a type switch over a single implementation anyway, since the unbuffered copier was removed in #1140.

One catch worth knowing, because it bit me when I tried it: **`Progress()` is not a row count on auto-increment tables.** `chunkerOptimistic.Feedback` (`pkg/table/chunker_optimistic.go:359`) *discards* `actualRows` and adds `chunk.ChunkSize` instead — distance travelled through the key space, so it can be compared against the auto-inc max. A 4-row test table reports 3000. So "read it from the chunker" needs the chunker to keep the number it is already being handed: #1150 adds `Chunker.RowsCopied()` summing `actualRows` (5 implementers, 2 of them fakes). The ledger stays in the chunker; it just becomes an honest one.

Also worth noting: a sink already receives this incrementally as `chunk_num_affected_rows` per chunk. The run-level total is a convenience, not new information.

## The ownership fixes

These are the real content of the PR and I'd keep them close to as written — I ported them to `ownership-ambiguity-errors` with very little change beyond dropping the observer plumbing in favour of one exported `status.ErrOwnershipAmbiguous`. The direct-DDL case is the one I'd prioritise: today a connection loss during `ALGORITHM=INSTANT/INPLACE` is indistinguishable from "MySQL declined the DDL", so the runner falls through and copies from a table that may already carry the ALTER.

One refinement I'd suggest. In the migration cutover this PR marks ambiguity whenever `renameMayHaveCommitted` is set, i.e. on any connection loss. But main already verifies server state afterwards, and that verification has three outcomes, not two: committed, definitively *not* committed, and couldn't check. Only the third is ambiguous — a verification that proves the rename didn't happen is a plain retryable failure and shouldn't be escalated to a human. Having `confirmRenameCompleted` return `(completed, verified)` keeps that distinction.

The `CutoverResultCallback` work is separable and has one decision in it that I don't think should be made silently: on main today a *failed* legacy `SetCutover` callback is retried (only `cutoverFuncSucceeded` aborts the loop). This PR makes any legacy-callback failure fatal, on the grounds that an error-only signature can't say whether it mutated first. Defensible, but it turns a transient Vitess-switch blip into "manual intervention required" for every existing caller — worth being explicit about, since the result-bearing callback is the escape hatch.

## Suggested shape

- **#1150** — phase telemetry through the metrics sink, copy totals from the chunker.
- **`ownership-ambiguity-errors`** — the correctness fixes, as typed errors.
- **A third PR** — `CutoverResultCallback`, with the legacy-path decision stated.

morgo:
Obsolete by https://github.com/block/spirit/pull/1155

PR #1138, opened 2026-08-15
migration: convert the test suite to the --statement path

morgo:
## Summary

First step toward removing `--table`/`--alter` from `spirit migrate` in favor of `--statement` only. This PR is **test-only** — no production code changes — so the eventual flag removal becomes a small mechanical change (delete the two struct fields and the else-branch in `normalizeOptions`).

- `NewTestRunner(t, table, alter)` keeps its signature but now composes ``ALTER TABLE `<table>` <alter>`` into `Migration.Statement`, so all ~109 call sites exercise the same `--statement` path as production callers (tern and strata already construct `Migration{Statement: ...}`).
- `WithTable`/`WithAlter` option helpers are removed; call sites converted to `WithStatement`.
- Direct `Migration{Table:, Alter:}` literals (resume, TLS, checksum-invalidation, conf-param tests) become `Statement:`.
- Tests that step through runner internals read the table name from `changes[0].stmt.Table` instead of `Migration.Table`, which is no longer populated by the helpers.

## Deliberately left on the legacy path

These exercise the `--table`/`--alter` path itself and die with it in the follow-up PR:

- `TestMissingAlter` and `TestBadOptions` — validation errors of the legacy path.
- The `TestUnparsableStatements` cases that rely on the alter clause executing **verbatim**: the `--statement` path re-renders DDL through the parser's `Restore()`, which currently drops parens on expression defaults (`DEFAULT ('{}')` → `DEFAULT '{}'`), see #542. That fidelity gap should be closed before the flags are removed, since `--alter` is the documented workaround.

## Coverage note

This flips the bulk of the integration suite from verbatim-alter execution to restored-statement execution, which is the point: restore-infidelity bugs in the `--statement` path now surface in the ordinary test suite instead of hiding behind the legacy path.

## Testing

- `go test ./pkg/migration/` — pass (34s, local MySQL 8.0.43)
- `go test -tags singleversion -run 'Resume|Checkpoint|UniqueOnNonUniqueData|ChunkerPrefetching|Unparsable' ./pkg/migration/` — pass (21s)
- `go vet` both tag variants, `golangci-lint run ./pkg/migration/...` — clean


aparajon (approved):
🤖 Approving. I went looking for ways the conversion could weaken the suite and didn't find one. The thing that makes this safe is easy to miss from the diff, and it's worth stating outright: after `normalizeOptions`, nothing in the production tree reads `Migration.Table` — the parsed statement is the only source of truth — so composing `ALTER TABLE <table> <alter>` and letting the parser re-derive the table is not an approximation of the old path, it *is* the path. The suite was previously the odd one out.

Two things worth pinning down, neither blocking:

**1 — the sequencing between this and the flag removal is load-bearing and nothing enforces it.** After this PR, the two `TestUnparsableStatements` cases that set `Table`/`Alter` directly are the only remaining coverage that the alter clause executes verbatim, i.e. the only thing that would notice #542 regressing. The body correctly says the fidelity gap "should be closed before the flags are removed" — but if the removal PR is written from the flag's perspective it will delete those two cases as legacy-path cruft and the gap closes silently instead. Worth an explicit link on #542 saying the removal is blocked on it, so the ordering survives a context switch.

**2 — the coverage claim is stronger than the body makes it.** This isn't only "restore-infidelity bugs now surface in the ordinary suite"; it's that the suite now runs the code path every library consumer already uses exclusively. `block/schemabot` builds `Migration{Statement: ...}` and never sets `Table`/`Alter`, so before this PR the ~109 runner tests were exercising a path with no remaining production consumer while the one that matters was covered by a handful of statement tests. That inversion is the real argument for the change and I'd put it in the commit message — it's what a reader in six months will want to know about why the suite changed shape.

<details>
<summary>Verified</summary>

I checked the composition for the ways it could be malformed rather than trusting the green run: every `NewTestRunner` call site passes a bare table identifier (none qualified, quoted, or containing whitespace), so `sqlescape.EscapeIdentifier` produces exactly one backtick-quoted name and never a `` `db.tbl` `` that would silently target the wrong table, and no call site's alter argument already contains an `ALTER TABLE` prefix that would compose into a double statement; `TestMissingAlter` and `TestBadOptions` are genuinely untouched on the legacy path (the `bot1`/`badalter` conversion in the diff is `TestBadAlter`, a different test, whose parse-error assertion holds for the whole statement as it did for the clause); the `r.migration.Table` → `r.changes[0].stmt.Table` swaps in the checkpoint and resume tests read the same value the runner itself now uses; and the constructs that go through `Restore()` in the converted call sites (`CONVERT TO CHARACTER SET`, `AUTO_INCREMENT=`, `CHANGE b b <type>`, `ADD UNIQUE`) differ only in rendering, with the expression-default cases correctly held back on the legacy path.

</details>

PR #1139, opened 2026-08-15
change: auto-detect GTIDs, remove --gtid / --enable-experimental-gtid

morgo:
## Summary

Removes `--enable-experimental-gtid` (migrate/move) and `--gtid` (sync). The change-source coordinate scheme is now selected automatically, per server, by `change.NewAutoClient` in the new `pkg/change/auto.go`, shared by all three programs:

- **Fresh runs** probe the server: `gtid_mode=ON` + `enforce_gtid_consistency=ON` selects the GTID client; anything else gets the binlog file+offset client.
- **Resumes** classify the checkpointed position instead of probing: a `file:offset` checkpoint keeps the binlog client even on a GTID-enabled server (e.g. checkpoints from older Spirit versions, or GTIDs enabled mid-run), and a GTID-set checkpoint on a server that no longer has GTIDs is a hard error — surfaced with run state preserved (not `resumeErrorIsDefinitive`), so it does not silently restart the copy from scratch. The two encodings cannot collide: a binlog filename always carries a `.NNNNNN` index and never parses as a GTID set, which the classifier tests pin down (including a UUID-lookalike binlog basename).

The shared code lives in `pkg/change` rather than `pkg/utils` because `pkg/change` imports `pkg/utils` (cycle), and everything it touches — the two client constructors, `ClientConfig` — is already there.

## Per-program notes

- **migrate**: the checkpoint's position is threaded into setup (`setupCopierCheckerAndReplClient(ctx, resumePosition)`); the `gtid_mode` preflight block is deleted (the probe replaces it).
- **move**: repl clients are no longer built in `setupUnderLocks` — they're built by the new `buildReplClients` from `newCopy` (fresh) or `resumeFromCheckpoint` (after the positions JSON is read), so selection happens once fresh-vs-resume is known. Selection is **per source**: N:M shards that disagree on GTID support simply mix schemes, since each source's coordinate is stored and classified independently. The reverse feed classifies each stored reverse position via `change.IsGTIDPosition` (its `GTID` config field is gone), and `targetCurrentPosition` auto-detects from each target.
- **sync**: `setup` reads the checkpoint before wiring the change source; injected `change.Source` (strata's VStream) is untouched, and a `--copy-only` checkpoint stores no position, so a later full sync probes fresh.

## Docs

The three flag sections are replaced by a `## GTID auto-detection` section in migrate.md (behavior + resume rules), with shorter cross-linking sections in move.md (per-source mixing; the reverse window detects from the *targets*) and sync.md (`RELOAD` only needed without GTIDs; copy-only interaction).

## CI

All stacks previously booted GTID-on, so the binlog client would have lost all runner coverage once selection became automatic:

- New `compose/nogtid.yml` overlay flips `gtid-mode`/`enforce-gtid-consistency` off.
- The 8.0.28 "Aurora version" runner now uses it (renamed "…no GTID") — it doubles as full-suite file+offset coverage and matches real Aurora, where GTIDs default to off.
- The 8.0.45 singleversion workflow is now a two-entry matrix (`gtid` / `nogtid`), so the resume/checkpoint suite exercises both coordinate schemes.
- Replication stacks (mysql8/8.0.42/8.4/9.7, semisync) stay GTID-on: their replicas connect with `SOURCE_AUTO_POSITION=1`.
- 19 live-GTID-stream tests in `pkg/change` now `skipUnlessGTIDEnabled(t)` (plus the `gtid` subtest of the decode-filter test).

## Testing

- `go build ./...`, `go vet ./...`, `golangci-lint run` — clean.
- Full `pkg/change` suite with `-race` against **both** a GTID-enabled server and a no-GTID 8.0.45 container: green (the no-GTID run is what caught the one missing skip, `TestGTIDFlushUnderTableLockErrors`).
- E2E resume paths in both modes: migrate (`TestEnumDrop`, checksum-invalidation setup), move (`TestBasicMove`, `TestResumeFromCheckpointE2E`, both reverse-window revert tests incl. N:M), sync (`TestSyncE2E`, `TestSyncResume`, `TestSyncCopyOnly`), and the singleversion `TestResumeFromCheckpoint*` subset on the no-GTID server.
- Compose overlay merge verified with `docker compose config` (command resolves to `--gtid-mode=off --enforce-gtid-consistency=off`).

## Heads-up

- **strata**: removing `ReverseFeedConfig.GTID` and `Sync.GTID` breaks strata's compile on its next spirit dep bump (`pkg/vstream` sets `Sync.GTID` today), and `NewReverseFeed` now takes a `ctx` parameter — downstream PR needed alongside the bump.
- **block/schemabot**: also breaks on the dep bump — it sets the removed `Migration.EnableExperimentalGTID` (`pkg/engine/spirit/execution.go`). Its fix is a deletion: it already probes `gtid_mode`/`enforce_gtid_consistency` itself with the same predicate as the new `GTIDEnabled`, so the probe and the field assignment both go. One behavioural difference to know: schemabot's probe fails *soft* (probe error → warn + binlog client) where `NewAutoClient` fails *hard* — hard is right by that point (`r.db` is an established pool, so a failure is real connectivity trouble, not a reason to silently downgrade the coordinate scheme), but the bump changes behaviour on that path as well as deleting code.
- **Branch protection**: the singleversion workflow's check name changes from `build` to `build (gtid)` / `build (nogtid)` — required-checks config may need updating.


aparajon (approved):
🤖 Approving. The mechanism is sound and the part I most expected to be hand-wavy — resume classification — is the part that's most carefully done: making the checkpoint's own encoding authoritative means a run always finishes in the scheme it started in, and routing the GTID-checkpoint-on-a-server-that-lost-GTIDs case to a non-definitive error is exactly right, since the alternative silently discards resumable copy progress. Removing the flag outright rather than leaving a compatibility override behind is the correct call — a second selection path is a second thing to maintain and reason about, and auto-detection is the behaviour you'd want anyone to get anyway. Two findings, neither blocking.

**1 — the PR description's "Heads-up" section is missing `block/schemabot`.** It lists strata as the downstream that breaks on the dep bump; `block/schemabot` also sets `Migration.EnableExperimentalGTID` (`pkg/engine/spirit/execution.go:54`) and will fail to compile on the dep bump. The good news is that its fix is a deletion rather than a port: it already does this probe itself, and its predicate is `gtid_mode == "ON" && enforce_gtid_consistency == "ON"` — semantically identical to the new `GTIDEnabled`, including treating the permissive modes as off. The one behavioural difference is worth knowing before it lands: its probe fails *soft* (any connection or query error logs a warning and falls back to the binlog client) where `NewAutoClient` fails *hard*. I think hard is the better choice here — by that point `r.db` is an established pool, so a failure is a real connectivity problem rather than a reason to silently downgrade the coordinate scheme — but it does mean the bump changes behaviour on that path as well as deleting code.

**2 — the reverse feed's capture/classify round-trip assumes the captured position is non-empty.** `targetCurrentPosition` probes the target and captures a position in that server's scheme; `NewReverseFeed` then re-derives the scheme from the captured string. Those agree for every non-empty position, but `IsGTIDPosition("")` is false by construction, so a GTID-enabled target that returned an empty `CurrentPosition` would get a *binlog* client handed a GTID-scheme (empty) start position. In a real move the target has just been written by the copy, so `gtid_executed` is non-empty and this is unreachable — which is precisely why it's worth an explicit guard rather than an implicit invariant: an empty position at that point means the capture failed, and erroring in `NewReverseFeed` says so at the site that knows, instead of producing a feed in the wrong scheme.

Action items, both follow-ups:

1. Add a bullet for `block/schemabot` to the "Heads-up" section of this PR's description, alongside the strata one — it sets the removed field in `pkg/engine/spirit/execution.go` and its own probe becomes redundant. Description edit only; no code change here.
2. Reject an empty captured position in `NewReverseFeed` instead of letting it classify as file+offset.

<details>
<summary>Verified</summary>

The claims I could check against the tree hold: `resumeErrorIsDefinitive` is an allowlist of sentinels and type assertions, so the new GTID-mismatch error is correctly non-definitive and the run fails with checkpoint and `_new` table intact rather than restarting the copy; deleting the `gtid_mode` preflight loses nothing, since it only ever guarded the opt-in that no longer exists and the probe now happens unconditionally in `NewAutoClient`; injected `change.Source` implementations bypass the classifier entirely in `datasync`, so a non-MySQL position encoding never reaches `IsGTIDPosition`; moving `readCheckpoint` ahead of the change-source wiring in `datasync.setup` is safe and strictly better — nothing between the two positions touches the target, and a checkpoint-read failure no longer strands a just-constructed client; the two encodings genuinely cannot collide because a binlog coordinate always carries a `.NNNNNN` index that defeats UUID parsing, which the classifier tests pin including the UUID-lookalike basename; and per-source classification in the sharded reverse feed means an N:M move whose shards disagree on GTID support mixes schemes correctly rather than forcing one.

</details>

morgo:
🤖 Addressed review feedback and resolved conflicts:

**Conflicts** — merged main (129d4e3c, the `--statement` test-suite conversion). The only textual conflict was `TestE2EGTIDChangeSource`, which main converted and this branch deletes; the deletion stands since the test existed solely to exercise the removed flag.

**aparajon #1 (schemabot heads-up)** — added to the description, including the soft-vs-hard probe-failure behaviour change.

**aparajon #2 + Copilot's suppressed reversefeed comment** — both addressed in one change (013ea947): `NewReverseFeed` now constructs each source's client through `change.NewAutoClient` instead of classifying locally. That gives the reverse feed the same validation as every other resume path — a GTID-set position on a server that no longer has GTIDs enabled fails at construction with the purpose-built error rather than as a stream failure at `Start`.

One deliberate deviation from the literal ask: I did **not** make `NewReverseFeed` reject an empty position. Empty is the documented head-start mode (`ReverseSource.Position`: "Empty means start from the source's current head"), used by the feed-level tests, and the move runner already refuses empty positions for reverse *windows* at its own layer with the rollback-safety error. The real hazard the review identified — an empty position silently classifying as file+offset even on a GTID-enabled target — is what the `NewAutoClient` change fixes: the empty case now probes the server, so the head-start scheme matches what a cutover capture on that server would have produced.

`NewReverseFeed` gained a `ctx` parameter for the probe (noted in the strata heads-up). Reverse feed + reverse window suites pass with `-race`, lint clean.

PR #1146, opened 2026-08-17
checksum: name the two ways a checksum runs out of attempts

aparajon:
`checksum.Run` gives up in two different ways, and a caller cannot currently tell them apart:

- **Every attempt errored** before it could compare the whole table (killed connection, a failure inside a pass). Nothing has been proven about the data, and the condition may well be gone by the next attempt.
- **Every attempt completed but kept finding row differences.** The table is diverging in a way the repairs cannot close, so another attempt reproduces it exactly. The common cause is a lossy ALTER, most often adding a UNIQUE index over data that is not unique: the copier drops the duplicate rows rather than refusing them, so the new table simply comes up short.

Both arrive as a bare error, so a caller deciding whether to retry a failed run has no way to ask which one it has. Getting it wrong is expensive in both directions: giving up on the first shape throws away a run a moment's wait would have cleared, and retrying the second re-verifies the whole table to reach the identical verdict — a checksum that found differences is deliberately not allowed to persist a watermark, so every attempt starts the checksum phase over from the beginning.

This exports a sentinel for each shape, `ErrAttemptsExhausted` and `ErrDifferencesExhausted`, and wraps them at the two exhaustion returns. The rendered messages are unchanged.

`pkg/migration` discarded the distinction on the path where it matters most: for a statement adding a unique index it replaced the checker's error with a fresh one, so the friendlier explanation cost the caller the cause. It now wraps instead, keeping the checker's error in the chain.

```
 Run exhausts its attempts
   |
   +-- some attempt errored ------> ErrAttemptsExhausted   (nothing proven; retry)
   |                                  wraps the last error
   |
   +-- all completed, still differing
                                 -> ErrDifferencesExhausted (reproducible; don't)
                                      |
                                      +-- migration wraps this, adding the
                                          unique-index explanation, without
                                          dropping the cause
```

*This PR was written by Claude Code (Opus 5).*


morgo (approved):
I like it, LGTM

PR #1150, opened 2026-08-17
feat(status): report workflow phases through the metrics sink

morgo:
Spirit already has one always-present, caller-supplied destination for telemetry: metrics.Sink, defaulting to NoopSink. Report phase transitions there instead of introducing a second subscription mechanism.

status.Tracker gains an optional sink. On every transition it emits the phase being entered, and on every phase close the phase that ended plus its duration. MetricValue has no labels, so the phase travels as the numeric status.State correlated inside a single Send batch — the same contract the copier's per-chunk metrics already rely on. Sending happens outside the tracker's mutex so a slow sink can never block a state read, and uses a background context so the final transitions of a cancelled run are still reported. With no sink installed the tracker behaves exactly as before.

move and datasync hardcoded a NoopSink with no way to replace it, so they gain the SetMetricsSink that migration already had; without it, phase reporting would have been migrate-only.

Copy completion reports the settled totals once, read from the chunker rather than a second tally kept by the copier. That required distinguishing two numbers the chunker had conflated: the optimistic chunker's Progress advances by the chunk's key-space width so it can be compared against the auto-increment max, which on a sparse table is nothing like a row count (a 4-row table reports 3000). Chunker.RowsCopied is the sum of the actualRows already passed to Feedback, and the optimistic chunker now keeps it instead of discarding that argument.

### DIFF
--- a/pkg/checksum/continuous_test.go
+++ b/pkg/checksum/continuous_test.go
@@ -83,6 +83,8 @@ func (c *testChunker) Feedback(chunk *table.Chunk, duration time.Duration, actua
 	defer c.mu.Unlock()
 	c.feedback = append(c.feedback, table.FeedbackCall{Chunk: chunk, Duration: duration, ActualRows: actualRows, Timestamp: time.Now()})
 }
+func (c *testChunker) RowsCopied() uint64 { return 0 }
+
 func (c *testChunker) Progress() (uint64, uint64, uint64) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/pkg/datasync/runner.go
+++ b/pkg/datasync/runner.go
@@ -86,8 +86,12 @@ type Runner struct {
 	// status block (#329).
 	lastCheckpoint status.LastCheckpoint
 
-	logger     *slog.Logger
-	cancelFunc context.CancelFunc
+	logger *slog.Logger
+	// metricsSink receives the copier's per-chunk metrics and the tracker's
+	// phase transitions. It defaults to a NoopSink, so a caller that installs
+	// nothing pays only for the discarded values.
+	metricsSink metrics.Sink
+	cancelFunc  context.CancelFunc
 	// sourceDBConfig connects to the read-only source: ForceKill and
 	// RejectReadOnly are disabled (see Run). targetDBConfig connects to the
 	// writable target and keeps the standard safe defaults — most importantly
@@ -165,6 +169,7 @@ func NewRunner(s *Sync) (*Runner, error) {
 	r := &Runner{
 		sync:              s,
 		logger:            slog.Default(),
+		metricsSink:       &metrics.NoopSink{},
 		continuousReadyCh: make(chan struct{}),
 		firstCleanPassCh:  make(chan struct{}),
 	}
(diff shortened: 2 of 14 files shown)


jonwinton (commented):
Reviewed at b2239d1. Reusing metrics.Sink is the right transport for phase/copy telemetry; I’m going to rework the Strata adapter around this rather than retain the generic WorkflowObserver.

Two implementation details are worth tightening:

1. Every runner installs its default *metrics.NoopSink into Tracker, so the nominally disabled path still takes sinkMu, creates a one-second timeout context/timer, and calls NoopSink on every transition. That contradicts the no-cost-when-unused contract. Keeping the tracker sink nil until SetMetricsSink receives a real sink (while the copier retains its NoopSink) avoids the hot-path allocations.
2. enter records enteredAt before the synchronous sink sends. A slow sink therefore contributes up to SinkTimeout to the phase duration even though Tracker.Do is documented as timing exactly the bracketed function. The phase clock should begin after entry delivery (with the state still published before delivery).

The existing comments correctly flag that optimistic RowsCopied is post-resume work, not a persisted all-attempt total.

For Strata’s follow-up, I’ll add the missing bounded-attempt terminal signal through the same sink rather than ask this PR to own workflow/SLO semantics. `failed` there means one Tracker.Do callback returned a non-cancellation error; pod recycling is `cancelled`, and only the outer workflow-run terminal is an SLO candidate. Durable mutation and ownership stay out of metrics and move to typed runner results/errors.

morgo:
🤖 Reply posted by Morgan's AI agent.

Both points addressed in `53a1ef3e`, along with the doc corrections Copilot raised.

**1. NoopSink was being installed.** You're right, and it undercut the whole argument for this approach. `SetMetricsSink` now treats a nil sink and a `*metrics.NoopSink` identically — reporting stays off — so the runners can keep passing their default through unconditionally without it costing anything. The sink/logger pair also moved into a single `atomic.Pointer[trackerSink]`, so the disabled path is one load and a nil check: no `sinkMu`, no timeout context, and no batch (the variadic slice was itself an allocation, so the producers check `reporting()` before assembling values, not just inside `send`). The copier keeps its own NoopSink, untouched.

**2. Phase clock started before delivery.** Also right, and worse than it looked: `enter` delivers *two* batches on most transitions (the previous phase's completion, then the new phase's entry), so both were landing inside the new phase. The state is still published before delivery — a concurrent `Get`/`Elapsed` never sees a stale phase — but `enteredAt` is now rolled forward by the delivery duration afterwards, guarded so that a fatal `Set(ErrCleanup)` racing an in-flight send doesn't get its own clock rewritten.

`TestTrackerSlowSinkDoesNotInflatePhase` pins it. With a sink that blocks 40ms per send and an empty `CopyRows` bracket, the old code reported an **81ms** copy phase; it now reports under a millisecond, and so does the `workflow_phase_seconds` value the sink receives.

**Copilot's five comments** were all the same real defect: I wrote in five places that the chunker "already carries" a resumed run's earlier rows, which contradicted the qualification `table.Chunker.RowsCopied` itself carries. Behavior is unchanged — as you say, optimistic `RowsCopied` is post-resume work and that's the honest thing for it to be — but the comments in `metrics.go`, `Tracker.RecordCopyCompleted`, and the three `recordCopyCompleted` helpers now say so, and tell consumers to read `copy_rows_completed` as per-run rather than per-migration. Also fixed the two smaller ones: `chunkerOptimistic.Feedback` still claimed it ignores `actualRows` (untrue since `RowsCopied` landed — it ignores them for `Progress` only), and datasync's `SetLogger` doc had drifted onto `recordCopyCompleted`.

On the follow-up: agreed on all three boundaries — the bounded-attempt terminal signal going through the same sink, `failed` meaning one `Do` callback returned a non-cancellation error with pod recycling as `cancelled`, and durable mutation/ownership staying out of metrics entirely. That last one is what the companion branch does: `status.ErrOwnershipAmbiguous` as a sentinel on the returned error, so "may I retry?" is answered by the error the caller is already holding rather than by a side channel they have to subscribe to. Happy to have this PR stay narrow and let the SLO semantics land on top.

morgo:
Obsolete by https://github.com/block/spirit/pull/1155

Issue #96, opened 2023-04-18
Configuration check for spirit options


morgo:
In https://github.com/squareup/spirit/pull/95 a fix was added to increase timeouts. 

This wasn't the cause of the customer issue, but when you look at the details it's kind of obvious where some timeouts could occur. Imagine the following scenario:
- User sets the chunk-target-time to 10s.
- A key is modified that is in the range that it's not above the high watermark, but it's not copied either (it's about to be).
- The copier is first to acquire the lock on the key.
- The replClient applier tries to copy the key immediately after (waiting for the copier thread to finish and release the lock).

In the above scenario, by design the lock wait time (including any retries) needs to be at least 10s by definition of this also being the chunk-target-time. In fact, since we expect chunk-target-time can sometimes be above the target, it should be some multiple (~5x) above the chunk-target-time.

After https://github.com/squareup/spirit/pull/95 merges, the defaults are now safe: (3+4+5+6+7=25s) lock wait retry and a 2s default chunk size. It's completely conceivable that some users might want a 10s chunk size, in which case we either need to prevent it, or automatically extend the lock wait retries to maintain this 5x ratio.

PR #1151, opened 2026-08-17
Remove --target-chunk-time from migrate, move and sync

morgo:
## Why

`--target-chunk-time` reads like a copier knob and hasn't been one for a while. The copier sizes each chunk against an in-memory byte budget (`--target-chunk-size`); `--target-chunk-time` only paced the **checksum** (and the composite copy chunker, which never had a byte budget wired up). Having a prominent "chunk time" flag sitting next to a "chunk size" flag invited users to tune the wrong one.

## What

The time target becomes a constant instead of a flag. `table.ChunkerDefaultTarget` moves from `100ms` to **`5s`** and is now the single wall-clock target for every time-signal chunker.

- `5s` is the top of the range the old flag accepted, and it is what `move` and `sync` already defaulted to.
- `migrate`'s checksum therefore goes from `500ms` → `5s`. In practice the checksum's chunks are cheap (server-side CRC, one row per chunk on the wire) and are usually bounded by `table.MaxDynamicRowSize` rather than by time — see the `checksum chunk size distribution` log line — so this mostly removes a knob rather than changing pacing.
- Copy chunk sizing is unaffected: it was already `--target-chunk-size`.

Removed along with the flag:

- `Migration.TargetChunkTime`, `Move.TargetChunkTime`, `Sync.TargetChunkTime` and their `Validate()` checks.
- `check.Resources.TargetChunkTime` and the settings check that pinned it to `100ms-5s`. The lock-timeout caveat from that comment (issue #96) now lives on `ChunkerDefaultTarget`.
- `ContinuousCheckerConfig.TargetChunkTime` and `DefaultContinuousTargetChunkTime` — dead code: the field was defaulted by the constructor and never read (the chunker carries its own target).

`checksum.CheckerConfig.TargetChunkTime` stays, but is now documented as what it actually is — reporting-only input to the end-of-pass chunk-size distribution line — and defaults to `table.ChunkerDefaultTarget`.

The `range_optimizer_max_mem_size` error in `pkg/dbconn` now points at `--target-chunk-size`, which is the flag that actually governs chunk size.

## Breaking change

`spirit migrate|move|sync --target-chunk-time=...` is now an unknown-flag error rather than a no-op. Automation that passes it needs updating. The Go API fields are gone too, so library callers (e.g. strata) will need a recompile.

## Tests

Two tests relied on `ChunkerDefaultTarget` being small and now pin `100ms` explicitly rather than inheriting the constant:

- `TestCompositeLowWatermark` — needs a 1s chunk to cross the 5x panic-shrink threshold.
- `TestYieldTimeout` — needs many small chunks so the 100ms yield timeout fires mid-pass.

Full suite green locally against MySQL 8.0/9.7, including `-tags singleversion` (resume) and `golangci-lint` clean.


morgo:
🤖 Thanks — all four points were right, fixed in a071ec34.

- **`docs/migrate.md` Checksum scaling (L402)** — my replacement was wrong. The checksum uses the *time* signal (now the fixed `table.ChunkerDefaultTarget`), not `--target-chunk-size`. Reworded to say that explicitly and to name the byte budget as the copier's.
- **`docs/migrate.md` `chunk-size` status field (L628)** — the `--target-chunk-size` link is correct there (it's the `copier` row), but the tail of the sentence still carried the time-based reading: a floor collapse was described as "the server is struggling". Under a byte budget it means the rows are too wide to fit the budget even at the floor. Also noted that the `checksum` row's `chunk-size` is time-sized rather than byte-sized.
- **`pkg/table/dynamic_chunk_sizer.go`** — good catch, the warning still named the removed flag. Now "...too wide to meet the chunk time budget". While there I renamed the `target-ms` log key to `target-time`: it always logged a `time.Duration`, and it now logs `5s`, so `-ms` was actively misleading. Updated the assertion in `dynamic_chunk_sizer_test.go` to match.
- **`pkg/migration/helpers_test.go`** — the doc block claimed a `TargetChunkTime=500ms` default, and its example had been mangled into a stray `//\t)` when `WithTargetChunkTime` was stripped. Rewritten to state the real defaults (`Threads=2`, `WriteThreads=2`, plus which sizing budgets apply) with a valid example.

Full suite green and `golangci-lint` clean.

JashLal (approved):

PR #1170, opened 2026-08-27
change: report the buffered (unflushed) position on the binlog status row

morgo:
The status block only ever showed the *flushed* position, on the ckpt row. That position advances only when a flush lands every buffered change, so while anything is held back — a watermark-deferred key, or a batch deferred by #1169's retry pass — the checkpoint is frozen by design and the block goes silent about the reader.

That is indistinguishable from a genuinely stalled feed, which is the case an operator most needs to tell apart. From a run this week:

```
binlog  deltas=453392  rotations=400 (0 forced)  flushed 29m4s ago (took 21m37s, 452571 rows)
ckpt    20s ago  f50a3ec0-...:1-38880294638
```

Half an hour on the same GTID. The reader was working normally throughout — `rotations` was climbing several files per status block — but nothing on screen tied that to a position, so there was no way to see how far ahead of the checkpoint it had got, or whether the gap was still growing.

## The change

`FeedStats` gains `BufferedPosition`. Both clients populate it from the field they already keep under `c.mu` (`bufferedGTID` / `bufferedPos`), so there is no new state and no new locking. The binlog row renders it as `read=<pos>`:

```
binlog  deltas=453392  rotations=400 (0 forced)  flushed 29m4s ago (took 21m37s, 452571 rows)  read=f50a3ec0-...:1-38891204771
ckpt    20s ago  f50a3ec0-...:1-38880294638
```

If `read=` moves between status blocks the reader is fine and only publication is blocked. The gap to the ckpt row is how much re-reading a restart would cost.

## Two deliberate choices

**It goes last on the row.** A GTID set gains a UUID per failover and has no bounded length. Anywhere but the end, a long one pushes the flush phrase — the other half of the answer — off a narrow terminal.

**It is not abbreviated.** I wrote a middle-eliding helper first and removed it:

- The ckpt row prints its position in full, and the entire point of this field is to compare the two by eye. Elide one copy and they stop being diffable.
- The interval that moves between status blocks belongs to whichever server is currently being written to. `MysqlGTIDSet.String()` sorts by UUID bytes, so that server need not sort last — a middle elision can hide exactly the digits that are changing and make a healthy reader look frozen. That is the precise misreading this field exists to prevent.

## Merging

For a sharded move the position is taken from the same feed as the flush figures rather than merged. Positions from different sources are not comparable — one coordinate space per feed — so there is nothing to sum. The stalest feed's is the useful one: that is the feed holding publication back, so its reader progress is the question.

## Compatibility

A feed that has not read anything yet omits the field, so every pre-existing status line is byte-identical. The existing stats tests were not touched and still pass.

## Testing

Four tests. Two were verified against mutations: taking the last feed's position instead of the stalest is caught by `TestStatusRowBufferedPositionFollowsStalestFeed`, and (while the helper still existed) tail-elision instead of middle-elision was caught by its test.

`pkg/change` passes, `-race` clean, `golangci-lint` reports 0 issues.

### DIFF
--- a/pkg/change/binlog.go
+++ b/pkg/change/binlog.go
@@ -1195,13 +1195,18 @@ func (c *binlogClient) FlushResidual() (int, int) {
 func (c *binlogClient) FeedStats() FeedStats {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	return FeedStats{
+	stats := FeedStats{
 		LastFlushAt:       c.lastFlushAt,
 		LastFlushDuration: c.lastFlushDuration,
 		LastFlushRows:     c.lastFlushRows,
 		Rotations:         c.rotations.Load(),
 		ForcedRotations:   c.flushedBinlogs.Load(),
 	}
+	// Already under c.mu, which is what guards bufferedPos.
+	if c.bufferedPos.Name != "" {
+		stats.BufferedPosition = formatBinlogPosition(c.bufferedPos)
+	}
+	return stats
 }
 
 // Flush empties the changeset in a loop until the amount of changes is considered "trivial".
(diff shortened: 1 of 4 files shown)


morgo:
🤖 Folding this into #1169 instead — it is the same operator-facing problem and reviewing them together is easier.

PR #1174, opened 2026-08-28
change: partition flush batches by unique secondary index range

morgo:
## Problem

A map-mode drain's batches are disjoint by primary key, and for a long time that was assumed to be enough. It is not.

Two concurrent `REPLACE` statements on PK-disjoint rows can still deadlock, and in [#1168](https://github.com/block/spirit/issues/1168) they did: the InnoDB cycle inverted between the clustered index and a `UNIQUE` secondary index. `REPLACE`'s duplicate detection takes a **next-key** lock on each unique secondary index — the gap below the record included — so two batches collide whenever any of their rows land in *adjacent* slots of any such index. Secondary key order is unrelated to primary key order, so PK disjointness says nothing about it.

The existing defence is the AIMD controller that narrows the flush until the deadlocks stop. It works, but it pays for safety in throughput and never stops paying: each contention step costs **4×** (concurrency and batch size both halve), and a table whose unique key correlates with anything keeps re-triggering it. On the table that motivated this, the drain never converged — the migration was cancelled.

## The conflict surface, measured

This is the part the design rests on, and it is a precise claim rather than a cautious one. `TestReplaceContendsOnlyOnUniqueIndexes` (merged in #1173) establishes it against a real server:

| Two rows are… | Contend? | Why |
| --- | --- | --- |
| adjacent in the PRIMARY KEY | **no** | a `REPLACE`'s clustered-index conflict is with the row bearing that exact PK, so under `READ COMMITTED` it is a record lock with no gap |
| equal in a *non-unique* secondary index | **no** | those records are keyed `(indexed columns, PK)`, so PK-disjoint rows always occupy distinct records |
| adjacent in a *`UNIQUE`* secondary index | **yes** | duplicate detection takes a next-key lock, gap included |

So **primary-key separation buys nothing**, and the earlier PK-sorting attempt was aimed at an index whose neighbours provably do not collide.

## What this does

The drain now:

1. **Chooses** the unique secondary index whose values are most *clustered* across this drain's own rows — measured, not configured, because whether a key correlates with anything is a property of the workload rather than the schema. A repeating leading column means physically adjacent sibling records and a near-certain collision; a uniformly distributed key has an adjacency probability of roughly `n²/N` per drain (~0.3 for 50,000 rows in 8.5 billion) and needs no help.
2. **Sorts** by it and cuts **contiguous** batches, nudging cut points to fall where the leading key value changes so a run of siblings is not split across two batches.
3. **Stripes** the batches into evens and odds, running one group at a time, so no two batches in flight together are neighbours in the sort order.

Rows close together in the chosen index therefore land in the *same statement*, where they cannot conflict, and statements that do run together are separated by at least one whole batch of intervening rows.

Three design points worth flagging because each could have gone the other way:

- **Range partitioning, not hashing.** Hashing spreads equal-ish values across buckets, which is precisely the arrangement that collides. The risk is gap adjacency, not shared records.
- **Separation is in *rows*, not value distance.** Whether two values are adjacent in the B-tree depends on the 8.5-billion-row table, not the 50,000 rows in hand, so no value-space margin means anything. A whole batch of intervening rows is a margin that does, because those rows are themselves records sitting between the two.
- **Striping is needed, not just sorting.** Handing the sorted list straight to the limiter would undo most of the benefit: the in-flight window is roughly contiguous, so neighbours would run together and every batch boundary would become a candidate collision. The cost is one barrier per drain, which is far cheaper than a 4× AIMD step that persists for several drains.

## Getting it wrong costs throughput, never correctness

Batches stay disjoint by key and map mode makes no cross-key ordering promises, so a misordered sort or a badly chosen index simply reproduces the *old* collision behaviour — which the AIMD controller still catches. Nothing here is load-bearing for data integrity.

That is what makes a best-effort comparator acceptable, where the earlier PK-sorting attempt's comparator bug was silent and harmful. That one had `int64`/`uint64`/`float64`/`string`/`[]byte` arms only, and go-mysql decodes `MYSQL_TYPE_LONG` to `int32`, `TINY` to `int8`, `SHORT` to `int16` and `INT24` to `int32` — so an ordinary `INT` column fell through to a string comparison and sorted `100` before `9`, the exact inversion the sort existed to prevent. Every integer width is handled explicitly here and tested for it, including the `>2^53` case where routing integers through `float64` would lose exactly the low bits that distinguish adjacent values.

## What partitioning cannot cover (so the AIMD controller stays)

- **Deletes.** A buffered delete keeps only its PK — `LogicalRow{IsDeleted: true}`, the before image is discarded at buffer time — so its position in a unique secondary index is unknowable. Deleted rows are grouped at the tail.
- **Second and subsequent unique indexes.** Sorting by one says nothing about the others; choosing the most clustered one is what makes this worthwhile.
- **`REPLACE`'s out-of-partition deletion cascade.** A `REPLACE` deletes any row conflicting on any unique index, including PKs not in the batch, whose other unique values are unknowable from here. Documented at `applier.UpsertRows`; irreducible.

## Degradation

No flag, and it turns itself off when there is nothing to do — a table with no usable unique secondary index has no conflict surface between PK-disjoint batches at all, so the sort would be pure cost. Also skipped: functional indexes (the key is an expression, so a row image cannot be placed in them) and indexes covering a column the ALTER *adds*, which the source's row images do not contain. A `flush partitioning enabled` line at Info reports the candidates once per subscription.

## Two incidental fixes

- **`TableInfo.UniqueSecondaryIndexes` guards a nil `db` handle** instead of dereferencing it. `NewTableInfoFromMeta` returns a `TableInfo` with no server behind it (strata's `pkg/vstream` builds these) and the change feed calls this on a table it did not construct, so the nil case is reachable from outside the package. The existing suite caught this.
- **The dispatch budget is computed by the drain and passed down** rather than created inside `applyBatchesConcurrent`, which is now called once per stripe. Two stripes each granted a full `drainDispatchBudget` would have silently doubled how long one flush may hold `flushMu`.

## Testing

`pkg/change`, `pkg/table`, `pkg/applier`, `pkg/autoscale`, `pkg/migration`, `pkg/move`, `pkg/datasync` all green; `golangci-lint run ./pkg/...` reports 0 issues.

New coverage, in rough order of what would worry me most:

- **`TestFlushPartitioningAppliesEveryRow`** — the correctness half, against a real server through the real subscription: 400 rows updated, a subset updated again so dedup collapses them, some deleted, then a full symmetric-difference check that the target matches the source value for value. A misordered sort, a dropped batch, or a row batched twice all surface here.
- **`TestCompareRowValuesHandlesEveryIntegerWidth` / `…IsTotal`** — every width go-mysql can produce, the `>2^53` precision case, and the antisymmetry/reflexivity properties `sort` requires (a comparator that is merely usually-right yields an arbitrary permutation, not a slightly worse one).
- **`TestStripeBatchesSeparatesNeighbours`** — the invariant the whole design reduces to, over several batch counts.
- **`TestBuildBatchesProducesContiguousRanges`** — batches tile the sorted order, no row appears twice, no leading-value run is split.
- **`TestFlushPartitioningChoosesTheCorrelatedIndex`** — self-configuration against real index metadata, including that the choice follows the *data* when the correlation moves.
- Degradation paths: no unique index, index on an added column, functional indexes, metadata-only `TableInfo`.

One thing the tests taught that is now documented in the code: a **single-column** unique index always scores exactly 0 on the clustering metric — its values cannot repeat, that is what unique means. Clustering can only ever appear as a repeating *leading* column of a composite unique key with a discriminator after it, which is exactly the production shape. My first draft of the integration seed got a duplicate-key error for assuming otherwise.

### DIFF
--- a/pkg/change/subscription_buffered.go
+++ b/pkg/change/subscription_buffered.go
@@ -1189,7 +1189,23 @@ func (s *bufferedMap) buildBatches(rows []drainRow, idx *partitionIndex) []*mapF
 	batchSize := s.effectiveBatchSize()
 	var batches []*mapFlushBatch
 	for i := 0; i < len(rows); {
-		end := min(i+batchSize, len(rows))
+		// Every batch must consume at least one row. The loop's only progress
+		// is i = j at the bottom, so a batch that ended where it started would
+		// spin here forever, allocating an empty batch per turn until the
+		// process died.
+		//
+		// It cannot happen today: effectiveBatchSize floors at 1, and every
+		// return in cutAtValueBoundary is past its start. The clamp is here
+		// because that invariant lives in two other functions — one of them in
+		// another file — and neither says a caller's loop termination depends
+		// on it. One comparison per batch against a hung migration is worth
+		// making even at probability zero.
+		//
+		// Clamped *before* the cut rather than after, because
+		// cutAtValueBoundary reads rows[hardEnd-1] and so needs the same
+		// guarantee its caller does; a guard placed after the call would only
+		// have replaced the hang with a panic.
+		end := max(min(i+batchSize, len(rows)), i+1)
 		if idx != nil {
 			end = cutAtValueBoundary(rows, idx, i, end)
 		}
--- a/pkg/change/subscription_buffered.go
+++ b/pkg/change/subscription_buffered.go
@@ -134,6 +134,11 @@ type bufferedMap struct {
 	// means DefaultBatchSize. See effectiveBatchSize.
 	batchSize int
 
+	// partitioner holds the unique secondary indexes a drain may partition
+	// its batches by, resolved lazily on first use. See
+	// subscription_buffered_partition.go.
+	partitioner flushPartitioner
+
 	// logger is supplied by the change.Source that owns this subscription.
 	// We keep only a *slog.Logger (not a back-pointer to the source) so the
 	// bufferedMap stays source-agnostic and can be reused by alternative
@@ -984,31 +989,25 @@ type mapFlushBatch struct {
 // cancellation) stay in the snapshot for the caller to merge back.
 // Returns false when any entry was deferred.
 func (s *bufferedMap) drainMapSnapshot(ctx context.Context, snapshot map[string]bufferedChange, applyWatermarkFilter bool) (bool, error) {
-	var batches []*mapFlushBatch
-	current := &mapFlushBatch{}
-	var batchStmtBytes int64
 	allChangesFlushed := true
 
-	cutBatch := func() {
-		if len(current.keys) > 0 {
-			batches = append(batches, current)
-			current = &mapFlushBatch{}
-			batchStmtBytes = 0
-		}
-	}
-
-	// Map iteration order is randomized, so batch membership differs between
-	// flushes. That does not affect the retry below: pass 2 re-applies the very
-	// same mapFlushBatch values pass 1 built, so a contended batch retries as
-	// itself regardless of iteration order.
+	// Map iteration order is randomized, so unpartitioned batch membership
+	// differs between flushes. That does not affect the retry below: pass 2
+	// re-applies the very same mapFlushBatch values pass 1 built, so a contended
+	// batch retries as itself regardless of how it was assembled.
 	//
-	// An earlier revision sorted by primary key here, on the theory that a
-	// consistent lock-acquisition order would help. It would not: the observed
+	// An earlier revision sorted by *primary key* here, on the theory that a
+	// consistent lock-acquisition order would help. It would not, and the reason
+	// is worth keeping because it is what points at the sort below: the observed
 	// deadlock cycle inverts between the clustered index and a secondary UNIQUE
 	// index, and secondary key order is unrelated to primary key order, so
-	// PK-sorted batches still interleave there. Narrowing concurrency is what
-	// breaks the cycle. See block/spirit#1168.
-	batchSize := s.effectiveBatchSize()
+	// PK-sorted batches still interleave there. The clustered index was never
+	// the conflict surface to begin with — a REPLACE's conflict there is with an
+	// exact PK, so under READ COMMITTED it takes a record lock and no gap, and
+	// PK neighbours cannot collide at all. Sorting by a *unique secondary* index
+	// is the version of that idea which addresses the surface that does collide.
+	// See block/spirit#1168 and subscription_buffered_partition.go.
+	rows := make([]drainRow, 0, len(snapshot))
 	for key, change := range snapshot {
 		// Keys the copier may have a read in flight for are deferred to a
 		// later flush; see mustDeferKey. The chunker is internally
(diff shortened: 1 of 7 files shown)


pkg/change/subscription_buffered.go
```diff
@@ -576,8 +596,18 @@ func (s *bufferedMap) effectiveFlushConcurrency() int {
...
+// autoscale.FlushBounds). Shrinking from a smaller start is not a problem: the
+// two paths reduce different things — FlushBounds re-shapes a fixed number of
+// rows in flight, the penalty here removes rows from flight.
 func (s *bufferedMap) effectiveBatchSize() int {
-	return shiftDown(DefaultBatchSize, s.concurrencyPenalty.Load(), minAdaptiveBatchSize)
+	start := s.batchSize
+	if start <= 0 {
+		start = DefaultBatchSize // out-of-tree callers, bare test maps
+	}
+	return shiftDown(start, s.concurrencyPenalty.Load(), minAdaptiveBatchSize)
```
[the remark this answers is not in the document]
morgo: 🤖 Real bug, fixed — but in **1173**, not here. `git log -S"s.batchSize"` points at 6ed4260d, the commit in the parent PR that made `BatchSize` configurable; before that the start was always `DefaultBatchSize`, so the start and the floor could never cross. Fix is 3a3f83b9:

```go
// The floor can never exceed the start. minAdaptiveBatchSize floors
// *shrinking*; a caller that deliberately configured fewer rows than that
// has not asked to be overruled, least of all by the contention path,
// where raising a statement's lock footprint is the opposite of the intent.
return shiftDown(start, s.concurrencyPenalty.Load(), min(minAdaptiveBatchSize, start))
```

Pinned by a new `TestSmallBatchSizeIsNeverRaised`. One hazard worth flagging for anyone touching that test: `adaptFlushConcurrency` guards on `if s.effectiveFlushConcurrency() > 1 || s.effectiveBatchSize() > minAdaptiveBatchSize`, so at concurrency 1 the penalty is never taken and a test that sets only `batchSize` passes with or without the fix. My first version of the test did exactly that and survived the mutation. It now sets `sub.flushConcurrency, sub.batchSize = DefaultFlushConcurrency, configured` and asserts the penalty was actually incremented; reverting the fix then fails it with "expected: 1, actual: 50".

pkg/change/subscription_buffered.go
```diff
@@ -1131,10 +1170,61 @@ func (s *bufferedMap) drainMapSnapshot(ctx context.Context, snapshot map[string]
...
+// it overrides the boundary alignment and cuts early. A single row over the
+// budget still flushes alone in its own batch, since a row cannot be split.
+func (s *bufferedMap) buildBatches(rows []drainRow, idx *partitionIndex) []*mapFlushBatch {
+	batchSize := s.effectiveBatchSize()
+	var batches []*mapFlushBatch
+	for i := 0; i < len(rows); {
+		end := min(i+batchSize, len(rows))
+		if idx != nil {
+			end = cutAtValueBoundary(rows, idx, i, end)
+		}
```
eeSeeGee, line 1211: My agent would like a check here for `end <= i` to avoid infinite looping

eeSeeGee (approved):

pkg/change/subscription_buffered.go
morgo: 🤖 Added in 5aed74d0 — and your agent's instinct was better than the first patch I wrote for it.

You are right that the loop has no other source of progress: `i = j` at the bottom is it, so a batch ending where it started spins forever, allocating an empty batch per turn until the process dies. It is unreachable today — `effectiveBatchSize` floors at 1, and every return in `cutAtValueBoundary` is past its `start` — but that invariant lives in two other functions, one of them in another file, and neither says a caller's termination depends on it.

The interesting part is *where* the guard goes. I first wrote it exactly where you pointed, after the cut:

```go
end := min(i+batchSize, len(rows))
if idx != nil {
    end = cutAtValueBoundary(rows, idx, i, end)
}
end = max(end, i+1)   // <- first attempt
```

Then I mutation-tested it by simulating the distant floor disappearing (`batchSize := 0`) and it panicked instead of hanging: `cutAtValueBoundary` reads `rows[hardEnd-1]`, so with `hardEnd == i == 0` it indexes `rows[-1]` before the guard is ever reached. A guard after the call only trades the hang for a crash. Clamping first fixes both, because the callee needs the same guarantee the caller does:

```go
end := max(min(i+batchSize, len(rows)), i+1)
if idx != nil {
    end = cutAtValueBoundary(rows, idx, i, end)
}
```

Under the same `batchSize=0` mutation that now degrades cleanly to one-row batches and terminates.

Two tests pin the invariants the clamp backstops, since nothing can kill the mutation of a guard that is unreachable by construction:

- `TestCutAtValueBoundaryAlwaysAdvances` — sweeps every start, batch sizes down to 1, and adversarial shapes (one value throughout, all distinct, long runs, a delete tail), asserting the cut always advances.
- `TestBuildBatchesAlwaysConsumesRows` — no empty batch and exact tiling, across batch sizes, partitioned and not.

Also documented the `> start` postcondition on `cutAtValueBoundary` itself, so the next change there sees that a caller depends on it.

PR #1184, opened 2026-09-01
chunker: gate prefetch on key-space density, not chunk cost

morgo:
The optimistic chunker flaps in and out of prefetch mode continuously. It is
visible mostly in the checksum, and it is both loud — three `WARN` lines every
few seconds for the length of a run — and slow:

```
08:29:05 [WRN] dynamic chunking is not working as expected
08:29:05 [WRN] switching to prefetch algorithm
08:29:06 [WRN] disabling chunk prefetching
  checksum  35.29% chunk-size=86481  threads=75
  checksum  37.74% chunk-size=100000 threads=76
  checksum  45.24% chunk-size=5062   threads=77
```

Those chunk sizes are the fingerprint. `5062`, `86481` and `100000` are terms 4,
11 and 12 of `1000 * 1.5^n` with uint64 truncation at each step (1000, 1500,
2250, 3375, **5062**, 7593, 11389, 17083, 25624, 38436, 57654, **86481**, →
clamped **100000**) — a chunker repeatedly restarting from `StartingChunkSize`.

## Two bugs

**Entry mis-fires.** `maybeSwitchToPrefetch` read "pinned at `MaxDynamicRowSize`,
sizer still wants to grow, and `p90*5 < ChunkerTarget`" as *"the key space has
large gaps"*. It isn't: it only says the chunk was **cheap**, and a cheap chunk
on a dense table is indistinguishable from a cheap chunk over a gap. On the
checksum path a chunk is a server-side CRC sized against `ChunkerDefaultTarget`
(5s), so the bar is `p90 < 1s` and 100k dense rows clear it routinely — the
condition holds for every healthy chunk. In the run above, ~215M rows/30s across
77 threads puts a 100k chunk right on the 1s line, which is why it flaps rather
than staying stuck.

**Exit is expensive.** `nextChunkByPrefetching` correctly detects the dense key
space and turns prefetch off, but reset the chunk size to `StartingChunkSize`.
The sizer regrows at most `MaxDynamicStepFactor` per 11-chunk feedback window, so
that is ~130 chunks back to the ceiling — and then it fires again.

## Fix

- Gate entry on the signal that actually answers the question: keys-per-row over
  a sliding 10-chunk window, built from the row counts already passed to
  `Feedback` (new `pkg/table/keyspace_density.go`). The threshold is *derived*
  from prefetch's own exit test — `MaxDynamicRowSize/StartingChunkSize`, i.e. 100
  keys/row — rather than picked, so entry and exit cannot disagree by
  construction.
- On exit, restore the chunk size prefetch was entered from. Prefetch is only
  ever entered from the ceiling, so that is a size this table has already been
  measured to sustain; if the region past the gap can't, the sizer and
  `panicShrink` bring it down within a window.
- Backstop: two episodes abandoned on their first chunk retire prefetch for the
  rest of the run.
- The three `WARN` lines become two `INFO` lines per transition, and transitions
  are now rare rather than continuous.

One subtlety: density is recorded **before** `Feedback`'s stale-feedback screen.
With dozens of workers in flight most feedback arrives labelled with a superseded
chunk size and is discarded — but a keys/rows pair is a property of the data, not
of the chunk size that happened to be in effect, so screening those out would
leave the window permanently short and prefetch permanently unreachable.

Also fixes a smaller pre-existing mislabel found on the way: a prefetch chunk was
tagged with the chunk size in effect *after* the exit reset rather than the
`OFFSET` that built it, so its feedback was mis-attributed instead of being
screened as stale.

## Measured

Simulating the reported table — dense 1B-key space, 8us/row so a 100k-row chunk
lands at 800ms:

| | chunks to cover 1B keys | avg chunk size | flaps |
|---|---|---|---|
| before | 36,434 | 27,448 | 254 |
| after | 10,105 | 98,973 | 0 |
| ideal | 10,001 | 100,000 | — |

~3.6x fewer chunks, so ~3.6x fewer round trips for the same work.

## Testing

- `TestOptimisticNoPrefetchOnDenseKeySpace` — regression test for the entry
  mis-fire; fails on `main`.
- `TestOptimisticPrefetchRestoresChunkSize` — regression test for the exit reset,
  the chunk labelling, and the rejection backstop.
- `TestKeySpaceDensitySparse` / `TestKeySpaceDensityTotals` — the window itself,
  including that a short window is never treated as sparse and that it slides.
- `TestOptimisticPrefetchChunking` (existing, sparse table) still enters prefetch
  as intended.
- `pkg/table` (also under `-race`), `pkg/copier`, `pkg/checksum`, `pkg/change`,
  `pkg/migration` green; `golangci-lint` clean.

### DIFF
--- a/pkg/copier/buffered.go
+++ b/pkg/copier/buffered.go
@@ -97,6 +97,7 @@ func (c *buffered) CopyChunk(ctx context.Context, chunk *table.Chunk) error {
 		return fmt.Errorf("failed to read chunk data: %w", err)
 	}
 	chunk.ActualBytes = rowsByteSize(rows)
+	chunk.SourceRows = uint64(len(rows))
 	// The callback runs on the applier's feedback coordinator goroutine; done
 	// closes only after feedback and metrics are sent, so both have completed
 	// before CopyChunk returns. Empty chunks take the same path — the applier
@@ -420,6 +421,7 @@ func (c *buffered) readWorker(ctx context.Context, quit <-chan struct{}) error {
 		// size the next chunk against a byte budget (memory-based dynamic
 		// chunking). Harmless when the chunker is in time mode — it ignores it.
 		chunk.ActualBytes = rowsByteSize(rows)
+		chunk.SourceRows = uint64(len(rows))
 
 		// Handle empty chunks immediately
 		if len(rows) == 0 {
--- a/pkg/table/chunk.go
+++ b/pkg/table/chunk.go
@@ -25,6 +25,24 @@ type Chunk struct {
 	// time when dynamicChunkSizer.TargetChunkBytes is set. Zero for the
 	// checksum path, which reads server-side and never sees the bytes.
 	ActualBytes uint64
+
+	// SourceRows is the companion of ActualBytes: the number of rows the
+	// producer actually read from the source for this chunk. Like ActualBytes it
+	// is transient and absent from the checkpoint JSON.
+	//
+	// It exists because the row count reaching Feedback() is not the rows
+	// *present*. On the copy path that argument is the applier's affected-row
+	// count from `INSERT IGNORE`, which skips any row the binlog applier had
+	// already written to the new table — so a fully dense chunk can report far
+	// fewer rows than it holds, or zero. The optimistic chunker's key-space
+	// density signal needs rows present, not rows written, or it reads an
+	// insert-hot region as a gap (see chunkerOptimistic.recordKeyDensity).
+	//
+	// Zero from producers that never hold the rows themselves — the checksum
+	// aggregates its CRC server-side — which is also what an empty chunk
+	// reports, and the two are interchangeable for density purposes because an
+	// empty chunk's affected-row count is zero too.
+	SourceRows uint64
 }
 
 // Boundary is used by chunk for lower or upper boundary
--- a/pkg/table/chunker_optimistic.go
+++ b/pkg/table/chunker_optimistic.go
@@ -34,8 +34,24 @@ type chunkerOptimistic struct {
 	// recent chunks, which is what "large gaps in the sequence" means. See
 	// keySpaceDensity for why chunk cost cannot stand in for it.
 	keyDensity keySpaceDensity
-	// prePrefetchChunkSize is the chunk size that was in effect when the
-	// current prefetch episode began, restored when the episode ends.
+	// lastDenseChunkSize is the most recent fixed-width chunk size that was
+	// validated against real rows: a chunk of that width came back holding at
+	// least one row, so its cost (time, or in-memory bytes) was actually
+	// measured. It is what prefetch restores on exit, and it is deliberately
+	// NOT simply the chunk size at entry — see prePrefetchChunkSize.
+	lastDenseChunkSize uint64
+	// prePrefetchChunkSize is the size the current prefetch episode will restore
+	// when it ends, captured at entry from lastDenseChunkSize.
+	//
+	// The chunk size in effect at entry is the wrong thing to restore. Reaching
+	// MaxDynamicRowSize is not evidence the table can sustain 100k-row chunks:
+	// over a gap the chunks are *empty*, and empty chunks are precisely what
+	// drive the size up (near-zero durations on the time signal; see
+	// calculateNewTargetChunkBytes for the byte signal, where a zero-byte p90
+	// deliberately returns a target above the ceiling so prefetch can fire). So
+	// the entry-time size can be one no real row has ever been measured under,
+	// and restoring it would have the copier materialize up to 100k wide rows in
+	// one read before any ActualBytes feedback could shrink it.
 	prePrefetchChunkSize uint64
 	// prefetchChunks counts chunks dispatched in the current prefetch episode,
 	// so an episode abandoned on its first chunk can be recognised as a
@@ -369,6 +385,7 @@ func (t *chunkerOptimistic) Reset() error {
 	t.chunkTimingInfo = []time.Duration{}
 	t.chunkPrefetchingEnabled = false
 	t.keyDensity.reset()
+	t.lastDenseChunkSize = 0
 	t.prePrefetchChunkSize = 0
 	t.prefetchChunks = 0
 	t.prefetchRejections = 0
@@ -451,7 +468,25 @@ func (t *chunkerOptimistic) recordKeyDensity(chunk *Chunk, actualRows uint64) {
 	if err != nil {
 		return
 	}
-	t.keyDensity.record(keys, actualRows)
+	// Prefer the producer's own count of rows read. actualRows is the applier's
+	// affected-row count on the copy path, and `INSERT IGNORE` does not count a
+	// row the binlog applier already wrote to the new table — so in an
+	// insert-hot region a dense chunk can report zero rows and be mistaken for a
+	// gap. See Chunk.SourceRows for why falling back on zero is safe.
+	rows := actualRows
+	if chunk.SourceRows > 0 {
+		rows = chunk.SourceRows
+	}
+	t.keyDensity.record(keys, rows)
+
+	// Remember the last chunk size proven against real rows, which is what
+	// prefetch restores on exit. keys == ChunkSize identifies a fixed-width
+	// chunk (next() builds those as chunkPtr..chunkPtr+chunkSize) and excludes a
+	// prefetch chunk, whose ChunkSize is a row offset while its width is the gap
+	// it crossed — restoring a prefetch offset would defeat the whole point.
+	if rows > 0 && keys == chunk.ChunkSize {
+		t.lastDenseChunkSize = chunk.ChunkSize
+	}
 }
 
 // maybeSwitchToPrefetch is the optimistic chunker's pre-update hook for the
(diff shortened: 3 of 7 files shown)


JashLal (changes requested):
🤖 Review by Codex

I found one blocker and one non-blocking issue:

1. **Blocker: restoring `MaxDynamicRowSize` can defeat the buffered copier memory bound.** The size saved on prefetch entry is not necessarily a size that was measured with real rows. In byte-sizing mode, empty chunks deliberately drive `chunkSize` to the 100k ceiling so prefetch can activate. After prefetch crosses the gap, restoring that value makes the copier materialize as many as 100k rows before `ActualBytes` feedback can shrink the next chunk. For wide rows this can exceed the 16 MiB target by orders of magnitude per worker and potentially OOM Spirit. Please preserve the last size validated on non-empty data, or retain the conservative `StartingChunkSize` reset for byte-sizing mode. A regression test should cover wide rows, a large empty key gap, then wide rows again.

2. **Suggestion: key-space density receives destination affected rows rather than source rows present.** On the buffered-copy path, `actualRows` comes from the `REPLACE INTO` affected-row count. MySQL can report two or more affected rows per input row when existing or unique-conflicting rows are replaced, which biases the density estimate toward dense and can suppress legitimate prefetching. The copier already has the source rows in memory, so it can capture `sourceRowCount := uint64(len(rows))` before the asynchronous callback and use that scalar for density while retaining `affectedRows` for the existing applier metrics and progress semantics. The checksum path should similarly use `sourceCount` for density.

JashLal (approved):

aparajon:
🤖 Adversarial correctness review, requested via the review channel. Reviewed at head 5878ae1.

**Verdict: correct, and a clear win — safe to land.** I attacked the entry gate, the exit restore, the rejection backstop, the hook contract change, and the locking, and could not construct a correctness failure: every path that changes the chunk size is mutex-held, the mode-transition semantics are clean, and the flap is genuinely gone (the derived threshold makes entry and exit agree by construction, which is the load-bearing idea here). The two findings below are signal-fidelity and lifecycle nits, not gates.

### Findings

**1. (minor) On the copier path, the density window measures rows *written*, not rows *present*.** `recordKeyDensity` receives the buffered copier's `affectedRows` from `INSERT IGNORE` (buffered.go:113/469), which counts only rows newly inserted — a row the binlog applier already `REPLACE INTO`-ed the new table is skipped and counts 0. In the insert-hot tail of an auto-increment table (exactly where the sizer tends to sit at the ceiling with cheap chunks), a chunk whose rows were all replaced ahead of the copier reads as `keys≈100000, rows≈0` → `sparse()` → a spurious prefetch entry on a dense region. The blast radius is well bounded by this PR's own machinery — the episode exits on its first chunk at ~2 chunks' cost, and two of those retire prefetch — but the retirement is run-scoped, so two false entries in the hot tail would also disable prefetch for a *genuine* gap elsewhere in the table. The copier has the true count in hand (`len(rows)` from `readChunkData`); feeding that to the density window (while keeping `affectedRows` for progress tracking) would make the signal measure what it claims to. The checksum path is clean — `targetCount` is a real row count.

**2. (nit) `open()` resets `keyDensity`, `prePrefetchChunkSize`, and `prefetchChunks` but not `prefetchRejections`; `Reset()` resets all four.** If the asymmetry is deliberate (rejections as a per-table memory that should survive `OpenAtWatermark` resume), a one-line comment would stop the next reader from "fixing" it; if not, add it to `open()`.

### Action items

1. **(Finding 1)** On the buffered copier, record key-space density from the rows actually read rather than `affectedRows` — or document on `recordKeyDensity` that the copier's signal undercounts binlog-raced rows and why the rejection backstop bounds the consequence.
2. **(Finding 2)** Reset `prefetchRejections` in `open()`, or comment why it deliberately survives reopen.

**Entry gate**
- The old condition really is satisfied by every healthy checksum chunk: 100k rows × 8µs = 800ms, and 800ms×5 < the 5s `ChunkerDefaultTarget` — confirmed against main's `maybeSwitchToPrefetch`, so the regression test's fails-on-main claim checks out analytically.
- `minKeysPerRowForPrefetch` is derived (`MaxDynamicRowSize/StartingChunkSize` = 100), so entry cannot authorize an episode the exit test kills — the flap is structurally closed, not tuned away.
- `sparse()` refuses a short window, and integer division can't round a dense table into sparseness.
- Mid-episode re-entry is blocked by the `chunkPrefetchingEnabled` guard, so an episode's own feedback climbing the sizer back to the ceiling can't reset the offset mid-gap.

**Exit and labelling**
- `leavePrefetch` restores the entry-time size (a size the table already sustained), and `panicShrink`/the sizer bound the downside if the region past the gap can't sustain it; `restore==0` falls back to the historical reset for tests entering prefetch by hand.
- The exit-triggering chunk is labelled with the `offset` that built it, and its feedback is then correctly screened as stale by `chunk.ChunkSize != t.chunkSize` rather than misattributed.
- `setChunkSize` clamps to the dynamic bounds, clears both feedback histories, and re-arms `pinnedAtFloor` — bypassing the growth cap only across a mode boundary where the cap's meaning doesn't hold.

**Concurrency and lifecycle**
- Every density/prefetch mutation happens under the chunker mutex (`Next`/`Feedback`/`Open`/`Reset` all lock); new tests pass under `-race`.
- `recordKeyDensity` skips unbounded chunks and non-numeric keys; `Range` can't underflow (upper > lower by construction).
- The `feedbackTime`/`feedbackBytes` hook contract change (true = hook owns the size) is honored by both callers and can't skip a legitimate target update — a false return always falls through to `updateChunkerTarget`.

**Tests and CI**
- Ran locally at head 5878ae1: `TestKeySpaceDensitySparse`/`Totals`, `TestOptimisticNoPrefetchOnDenseKeySpace`, `TestOptimisticPrefetchRestoresChunkSize`, `TestOptimisticPrefetchChunking` (now fed real row counts — a genuine strengthening), plus the full `pkg/table` package: all green.
- All CI checks green at head; no unresolved bot threads.

*This review was performed by Cato, an AI review agent operated by Armand.*

aparajon (approved):

*This stamp was left by Cato, an AI review agent operated by Armand.*

morgo:
🤖 Both of Cato's findings addressed in f5f39ba5, and I picked up the Codex blocker too since it turned out to be the same class of mistake — trusting a number that wasn't measuring what the code thought it measured.

**Finding 1 — density measures rows written, not rows present.** Correct, and the mechanism you described is the right one. Worth recording that the two reviews disagreed here: Codex read the copy path as `REPLACE INTO` (counting 2 per replaced row, biasing *toward* dense and suppressing prefetch), Cato read it as `INSERT IGNORE` (undercounting, biasing *toward* sparse and causing spurious entry). Cato is right — the copy path is `INSERT IGNORE` at `single_target.go:569` / `sharded.go:603`; `REPLACE INTO` is `UpsertRows`, the binlog applier, which never calls `chunker.Feedback`. So the bias is toward false sparseness, i.e. spurious entry on dense data, exactly as described. Both reviews recommended the same fix regardless.

Fixed by carrying the producer's own count on the chunk — `Chunk.SourceRows`, the companion of the existing `ActualBytes` — and preferring it for density, while `affectedRows` keeps feeding progress and metrics. Regression test `TestOptimisticPrefetchDensityUsesSourceRows` feeds dense chunks with `affectedRows=0`; it fails without the fix.

**Codex blocker — restoring an unmeasured chunk size.** This one was a real hole and my PR description asserted the opposite ("a size this table has already been measured to sustain"), so thank you for pushing on it. That claim is true for the dense-table flap this PR is about, and false in exactly the case prefetch exists for: over a gap the chunks are *empty*, and empty chunks are what drive the size to the ceiling — on the byte signal deliberately so, since `calculateNewTargetChunkBytes` returns a target above the ceiling on a zero-byte p90 precisely so prefetch can fire. So the restored size could be one no row had ever been read under.

Now restores `lastDenseChunkSize`: the last fixed-width size a chunk actually holding rows was measured under. `keys == ChunkSize` identifies a fixed-width chunk, so a prefetch offset can't masquerade as a measured size. `TestOptimisticPrefetchRestoreIsMeasured` is the wide-rows → empty-gap → assert test you asked for (16KB rows, byte signal); it fails without the fix.

This is deliberately conservative and it does cost something on gapped tables — I'd rather state the number than bury it. On the `TestOptimisticPrefetchChunking` fixture (three 300B gaps), restoring the unmeasured ceiling reads the table in 192 chunks; restoring the measured size takes 357, against 430 before this PR. The remaining ramp is what buys the memory bound. The dense-table case that motivated the PR is unaffected, because there the last measured size *is* the ceiling: still 10104 chunks vs 36434, averaging 98973 rows.

**Finding 2 — `open()`/`Reset()` asymmetry.** Not deliberate, it was an oversight. I've made it deliberate rather than symmetric, with the comment you asked for: `open()` also runs on the `OpenAtWatermark` resume path, and reopening at a watermark doesn't change how sparse the key space is, so the rejection count should survive it — otherwise a boundary table re-learns it, and re-logs it, once per resume. `Reset()` does clear it, because its contract is "as if `Open()` was just called".

Verified at `f5f39ba5`: `go build`, `golangci-lint` clean, and `pkg/table` / `pkg/copier` / `pkg/checksum` / `pkg/applier` / `pkg/change` / `pkg/migration` green, with `pkg/table` and `pkg/copier` also under `-race`. Both new tests confirmed to fail with the fixes reverted.

Issue #908, opened 2026-06-09
Make the buffered copier the default for migrate


morgo:
### Summary

Today `spirit migrate` defaults to the **unbuffered** copier (`INSERT IGNORE INTO _new ... SELECT FROM original`), with the buffered DBLog copier available opt-in via `--buffered`. This proposes making the **buffered copier the default**, and replacing the flag with `--unbuffered` to opt back into the legacy copier.

### Motivation

The unbuffered copier's `INSERT ... SELECT` takes shared row locks on the source for the duration of each chunk (the `SELECT` side does not use MVCC), which contends with production OLTP touching the same rows. The buffered copier reads rows into Spirit and writes them through the applier (`REPLACE INTO _new VALUES (...)`), reading with MVCC and taking **no locks on the source**.

The change source is already always buffered — it applies binlog row images via the applier and never runs `SELECT FROM original` (see #746). So making the copier buffered by default eliminates source-side data-lock contention end-to-end, leaving only brief metadata locks (instant/inplace attempt, checksum start, cutover).

The buffered copier has been stable for some time — it is the only copier used by `move`/cross-server migrations — and it supports multi-table migrations.

### Proposal

- Make the buffered copier the default for `migrate`.
- Replace `--buffered` (opt-in) with `--unbuffered` (opt-out into the legacy `INSERT IGNORE .. SELECT` copier), documented as a legacy mechanism that is still supported but takes more locking.
- Internally, rename the `Buffered` fields (on `Migration` and `copier.CopierConfig`) to `Unbuffered` with inverted logic so the buffered path is the zero-value default.

### Backward compatibility

This removes the `--buffered` flag. Anyone passing `--buffered` in automation will need to drop it (buffered is now the default); `--unbuffered` is available to retain the old behavior.

PR #1186, opened 2026-09-02
move: accept a target that is stricter about NULL than its source

morgo:
## Problem

A pre-created target table has to match the move source exactly, with one existing exception: the column-level `AUTO_INCREMENT` attribute, which a sharded target legitimately drops in favor of a Vitess sequence.

The shard key needs the same treatment for the opposite attribute. A Vitess primary vindex cannot map NULL to a keyspace id, so a sharded target declares its shard key `NOT NULL` — while the source may still permit NULL, because the `ALTER` to tighten it was never affordable on a multi-terabyte unsharded table. Pre-flight refused that pairing:

```
table 'corder' exists on target 0 (dest) but schema does not match source;
reconcile the target to the source with: ALTER TABLE `corder`
MODIFY COLUMN `customer_id` bigint NULL DEFAULT NULL
```

That leaves an operator choosing between a correct target schema and being able to move into it at all.

## Change

`statement.DiffOptions.IgnoreNotNullRelaxation` skips a nullability difference in exactly one direction: when the emitted `MODIFY COLUMN` would only **weaken** the constraint. A difference the `MODIFY` would have to tighten stays a real change, so the option can never quietly accept a schema that lost a `NOT NULL` the reference had. Default false; general schema diffing is unchanged.

Tightening is safe for a move because nullability is metadata, not row bytes. The copy and the checksum compare values, and every column's NULL-ness is compared explicitly — `ColumnMapping.ChecksumExprs` emits an `ISNULL()` digit per column — so a tightened column whose source holds no NULLs is identical on both sides, and one that *does* hold a NULL is reported rather than hidden.

### Two supporting changes

**Split the comparison in two.** `schemaDiff` kept its strict semantics and now serves only source↔source (`source_schema_consistency`); the new `targetSchemaDiff` sets the option and serves the source→target checks (`target_state` and `resume_state`, so a move allowed to start does not fail on resume). Sources of one move must be identical — tolerating a tightening there would make the verdict depend on which source sorted first.

**Normalize each side's `DEFAULT NULL` on its own nullability.** The normalization was gated on one side's, which was equivalent while unequal nullability returned early. But a forgiven relaxation compares a nullable side rendering `DEFAULT NULL` against a `NOT NULL` side carrying no default, and reported that as a *default* difference — reintroducing the very `MODIFY` the option suppresses. Verified load-bearing: reverting just that hunk fails the new integration test with `MODIFY COLUMN customer_id bigint NULL DEFAULT NULL`.

## On the naming

The option is named in the differ's own vocabulary rather than after the move's target, because `DiffCreateTables` diffs `got->want`: inside `pkg/statement` the move's *target* is the receiver and the parameter literally named `target` is the move's *source*. A flag called `IgnoreNotNullOnTarget` would read backwards in the file that implements it. The move's "target" vocabulary lives in the consumer, `targetSchemaDiff`, where it is correct — and its doc comment pins the orientation so the arguments cannot be swapped without forgiving the dangerous direction.

## What this does NOT do

It does not create the constraint. A target table that does not exist yet is still created from the source's `SHOW CREATE TABLE` verbatim, so an absent target yields a nullable column and no error — the constraint has to come from whoever authors the target schema.

It also does not filter or rewrite rows. If the source data does contain a NULL in a tightened column, the move fails rather than substituting a value: at row-hashing time for a sharded target (the applier hashes the shard key before any INSERT), otherwise at the checksum, since the copy runs under spirit's non-strict session `sql_mode` where a batched INSERT would coerce the NULL to the type's implicit default. The outcome is a failed move, never a silently altered value — but the checksum is a late place to learn it, so a caller wanting a fast failure should probe the tightened columns for NULLs before starting the copy. `docs/move.md` says so.

## Tests

- `pkg/statement`: 7 cases on `TestDiff_DiffOptions` covering both directions, the `DEFAULT NULL` rendering, and that a type change or an explicit default on a forgiven column is still reported.
- `pkg/move/check`: unit coverage that `targetSchemaDiff` forgives a stricter target while `schemaDiff` still rejects nullability drift between sources in either order, plus three integration tests against real `SHOW CREATE TABLE` output — which is where the rendered `DEFAULT NULL` bites, and a hand-written-DDL unit test misses it.
- The directional guards fail without the change, so they are not tautologies.
- `./pkg/statement/... ./pkg/move/... ./pkg/migration/... ./pkg/table/...` pass; `golangci-lint` clean.

Filed without a prior issue against the template's suggestion — happy to open one if you'd like the discussion there. The consumer is strata's `keyspace move-tables`, which will pin this and then surface the accepted divergences in its own pre-move preview.

### DIFF
--- a/docs/move.md
+++ b/docs/move.md
@@ -139,7 +139,9 @@ A table that already exists on the target is used as-is, provided it is empty an
 
 The reverse of either — a target looser than its source — is still a mismatch and fails pre-flight, as does any other difference (column types, charset, collation, indexes, constraints). The error reports the `ALTER` that would reconcile the target.
 
-A `NOT NULL` target column does not make Move filter or rewrite source rows. If the source data does contain a NULL there, the move fails — at row-hashing time for a sharded target, otherwise at the checksum — rather than silently substituting a value. Confirm the column holds no NULLs before starting a large copy.
+A `NOT NULL` target column does not make Move filter or rewrite source rows. If the source data does contain a NULL there, the move fails rather than silently substituting a value — at row-hashing time for a sharded target, otherwise at the checksum.
+
+Failing at the checksum is correct but slow. The copy writes with `INSERT IGNORE`, which stores the type's implicit default instead of erroring, so the NULL is only detected once the [initial checksum](#two-checksum-model) compares the rows; that checksum repairs and retries up to three times, and every attempt re-copies the chunk, re-coerces the same NULL and finds it again before the run gives up. On a table large enough to be worth moving this way, that is hours of copying and verifying to learn something a single `SELECT 1 FROM <table> WHERE <column> IS NULL LIMIT 1` answers up front. Confirm the column holds no NULLs before starting the copy.
 
 ### threads
 
--- a/docs/move.md
+++ b/docs/move.md
@@ -132,6 +132,15 @@ The in-memory byte budget the buffered copier sizes each copy chunk against. Mov
 
 A Go MySQL DSN for the target database. Tables will be created here automatically from the source schema.
 
+A table that already exists on the target is used as-is, provided it is empty and its schema matches the source. "Matches" permits the target to be *stricter* in two specific ways, so a declaratively-managed target does not have to mirror artifacts of its unsharded source:
+
+- the source's column-level `AUTO_INCREMENT` may be absent on the target (its ids come from elsewhere, e.g. a Vitess sequence);
+- a column the source declares nullable may be `NOT NULL` on the target — for example a shard key, which cannot be NULL in a sharded keyspace.
+
+The reverse of either — a target looser than its source — is still a mismatch and fails pre-flight, as does any other difference (column types, charset, collation, indexes, constraints). The error reports the `ALTER` that would reconcile the target.
+
+A `NOT NULL` target column does not make Move filter or rewrite source rows. If the source data does contain a NULL there, the move fails — at row-hashing time for a sharded target, otherwise at the checksum — rather than silently substituting a value. Confirm the column holds no NULLs before starting a large copy.
+
 ### threads
 
 - Type: Integer
--- a/pkg/datasync/runner.go
+++ b/pkg/datasync/runner.go
@@ -1071,6 +1071,29 @@ func (r *Runner) createTargetTables(ctx context.Context) error {
 // would carry column defaults rather than the source's values, so an ALTER on
 // the target alone would not make the copy correct.
 //
+// The comparison is deliberately exact, unlike move's source->target check
+// (move/check.targetSchemaDiff), which forgives a target that drops the
+// source's column-level AUTO_INCREMENT or is stricter about NULL. Neither
+// relaxation transfers, because the two gates guard different states:
+//
+//   - move requires a pre-existing target table to be EMPTY (target_state
+//     rejects one holding any row), so only rows not yet copied are at stake
+//     and its unconditional pre-cutover checksum vets every one of them. This
+//     gate fires on a target that may already be half-copied by an earlier
+//     attempt — the resumable-checkpoint case above — where a divergence can
+//     mean the rows already on the target are wrong, and no schema relaxation
+//     can establish otherwise.
+//   - move's relaxations exist for a sharded Vitess target: ids from a
+//     sequence, and a shard key that cannot be NULL because a primary vindex
+//     has no keyspace id for one. Sync has a single unsharded target and no
+//     cutover, so neither reason arises here.
+//
+// The strictness is therefore a decision, not an oversight: sync copies into a
+// target shaped exactly like its source or it does not copy at all. Revisit it
+// here on its own merits rather than by reaching for move's options — the two
+// paths agreeing about a target should be on purpose, and so should their
+// disagreeing. TestSyncVerifyExistingTargetTableRequiresExactSchema pins it.
+//
 // Regular secondary indexes are excluded from the comparison on both sides.
 // With DeferSecondaryIndexes the target is deliberately created without them,
 // and an attempt that died mid-copy leaves it that way; restoreSecondaryIndexes
--- a/pkg/datasync/sync_test.go
+++ b/pkg/datasync/sync_test.go
@@ -947,6 +947,61 @@ func TestSyncValidate(t *testing.T) {
 	}
 }
 
+// TestSyncVerifyExistingTargetTableRequiresExactSchema pins sync's deliberate
+// divergence from move: move's source->target check forgives a target that
+// drops the source's column-level AUTO_INCREMENT or is stricter about NULL,
+// because it has already established the target is empty and vets the copy with
+// an unconditional pre-cutover checksum. Sync's gate fires on a target that may
+// be half-copied from an earlier attempt, so it requires an exact match — see
+// verifyExistingTargetTable for the full reasoning. If this is ever relaxed, it
+// should be because someone decided to, not because move's options got reused.
+func TestSyncVerifyExistingTargetTableRequiresExactSchema(t *testing.T) {
+	r := &Runner{target: applier.Target{Config: &mysql.Config{DBName: "sync_dest"}}}
+
+	tests := []struct {
+		name     string
+		source   string
+		target   string
+		wantDiff string // empty means the target must be accepted
+	}{
+		{
+			name:   "identical",
+			source: "CREATE TABLE t1 (id BIGINT PRIMARY KEY, customer_id BIGINT DEFAULT NULL)",
+			target: "CREATE TABLE t1 (id BIGINT PRIMARY KEY, customer_id BIGINT DEFAULT NULL)",
+		},
+		{
+			// Accepted by move — a sharded target's shard key cannot be NULL —
+			// and rejected here.
+			name:     "target stricter about NULL",
+			source:   "CREATE TABLE t1 (id BIGINT PRIMARY KEY, customer_id BIGINT DEFAULT NULL)",
+			target:   "CREATE TABLE t1 (id BIGINT PRIMARY KEY, customer_id BIGINT NOT NULL)",
+			wantDiff: "MODIFY COLUMN `customer_id`",
+		},
+		{
+			// Also accepted by move — a sharded target takes its ids from a
+			// Vitess sequence — and rejected here.
+			name:     "target drops column AUTO_INCREMENT",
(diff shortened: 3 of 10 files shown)


aparajon:
🤖 Adversarial correctness review, requested by Armand Parajon and performed by his agent. Reviewed at head `b60e89e4`.

**Verdict: the direction is right, the safety argument holds all the way down, and this is safe to land — the one thing worth a decision before merge is that only one of spirit's two source→target copy paths gets the relaxation.** I went after the orientation (the whole change is one boolean whose safety depends on reading `a`/`b` against the move's vocabulary), the claim that a real NULL can't be laundered, and whether the new guards are tautologies. All three held, and the second one held further than the body claims — repair cannot launder it either. No prior issue needed from my side; the body carries the discussion.

### Findings

**1. `datasync` has the same gate and does not get either relaxation.** `datasync.Runner.verifyExistingTargetTable` (`pkg/datasync/runner.go:1089`) is the direct analogue of `move/check.validateExistingTargetTable` — same job, same trigger (a target table that already exists), same output shape — and it still calls `DiffCreateTables` with a bare `NewDiffOptions()`. So it refuses not only the nullability tightening this PR forgives but also the column-level `AUTO_INCREMENT` difference move relaxed earlier, and its error hands back the same `MODIFY COLUMN ... NULL DEFAULT NULL` the problem statement opens with. Since that gate only fires when the target table pre-exists, its trigger is exactly this PR's premise: a declaratively-authored target that is deliberately stricter than its source. Its stated reason for failing rather than repairing is that `ColumnMapping` silently drops any column the two sides do not share — which is a real hazard for a *renamed or missing* column and does not arise for nullability, where the names match and `ChecksumExprs` compares NULL-ness explicitly. So the safety reasoning this PR establishes transfers. Whether datasync is in scope is your call; what would be unfortunate is the two paths disagreeing about the same target by accident rather than on purpose.

**2. (doc) "the move fails at the checksum" understates what the failure costs.** The initial checker is built with `FixDifferences: true` and no `MaxRetries`, which `NewChecker` normalizes to 3. So one NULL in a tightened column buys: a full checksum pass, a chunk repair that re-coerces, a chunker reset, and the same again — three times — before `ErrDifferencesExhausted`. The body already calls the checksum "a late place to learn it" and tells callers to probe first, which is exactly right; `docs/move.md` line 142 says only "the move fails ... at the checksum". On a table the size of the one motivating this change, the gap between that sentence and three full verification passes plus three repairs is the operator's afternoon, and the doc is where they'll look.

**3. (nit) The coercion is unconditional, not batch-dependent — the comment gives away a guarantee it has.** `targetSchemaDiff`'s doc says "a batched INSERT coerces the NULL to the type's implicit default instead of erroring", which invites the reader to wonder what a single-row INSERT does (non-strict mode still raises `ER_BAD_NULL_ERROR` for one row). The copier's form is `INSERT IGNORE ... SELECT` — `dbconn`'s comment at `conn.go:301` names it — and `IGNORE` downgrades that error to a warning regardless of row count. Saying so makes the branch total instead of leaving a case the reader has to reason about, and it costs a word.

**4. (nit) `pkDroppedNullabilityChange` is now the same predicate under a different gate.** `diffColumns` computes `!sourceCol.Nullable && targetCol.Nullable` for the dropped-primary-key case; `columnsEqualWithContext` computes `!a.Nullable && b.Nullable` for the option. Same rule, two sites, one keyed on PK membership and one on a flag — and with the option on, the PK-drop suppression is wholly subsumed. The new comment already cross-references it, which is most of the value; worth one line at the `diffColumns` site pointing back the other way, so whoever changes either finds the other.

### Action items

1. **(Finding 1)** Decide datasync explicitly — either give `verifyExistingTargetTable` the same options (a shared `targetDiffOptions()` would keep the two from drifting), or add a sentence there saying why an exact match is required on that path when move's is not.
2. **(Finding 2)** Extend `docs/move.md`'s NULL paragraph to say the failure arrives after the initial checksum's repair attempts are exhausted, which on a large table is hours — reinforcing the pre-probe advice already in the PR body.
3. **(Finding 3)** Replace "a batched INSERT" with the copier's actual `INSERT IGNORE`, so the guarantee does not read as conditional on batch size.
4. *(optional)* **(Finding 4)** Back-reference the option from `pkDroppedNullabilityChange`.

The orientation is right end to end: `DiffCreateTables(table, want, got)` calls `got.Diff(want)`, `diffColumns` passes the receiver's column as `a` and the parameter's as `b`, and `targetSchemaDiff` passes the source as `want` and the target as `got` — so `a` is the target and the forgiveness fires only when the target is NOT NULL and the source nullable; there is exactly one production call site of `columnsEqualWithContext` and only `targetSchemaDiff` ever sets the option, so neither `schemaDiff` nor `datasync`'s call can pick it up. I mutation-tested three ways and all three guards bite: inverting the direction fails four subtests including `LooserTargetStillRejected`; restoring the one-sided `DEFAULT NULL` gate fails exactly `StricterTargetAccepted_SourceRendersDefaultNull`, confirming that hunk is load-bearing as claimed; setting the option in the shared `moveDiffOptions()` base fails `TestSchemaDiffRejectsNullabilityBetweenSources`. The normalization split is behaviour-preserving for every other caller, since reaching it with the option off implies `a.Nullable == b.Nullable` and the two gates then agree. The safety net is unconditional and pre-cutover — move has no way to skip the checksum, the flow is copy → initial checksum → sentinel → cutover, and cutover will not proceed on a failed initial pass. `ChecksumExprs` really does emit `IFNULL(...,'')` *and* `ISNULL(col)` per column on both sides, so a coerced NULL differs in the ISNULL digit even for a `VARCHAR NOT NULL` whose implicit default is `''` — the one case where the IFNULL digit alone would collide. Repair cannot launder it either: `FixDifferences` re-copies through the same applier, the coercion recurs, and every attempt finds it, which is precisely the `ErrDifferencesExhausted` case ("the repairs cannot close" — a lossy ALTER). `MaxRetries: 0` in move's initial checker is normalized to 3, so that retry loop is not vacuous. `NO_AUTO_VALUE_ON_ZERO` is indeed the *only* mode spirit sets, so the session really is non-strict on the copier, the checksum recopy and the applier alike. `needsExplicitPosition` still forces a MODIFY on a reorder, so a forgiven column cannot hide a column-order difference. No test was deleted or weakened — the diff is additions plus comment edits, and the `TestSchemaDiffIgnoresColumnAutoIncrement` change adds an assertion rather than replacing one. `go build ./...` clean and the PR's unit tests pass locally; the three integration tests I could not run here (this host has no `spirit` MySQL grant) are green across all seven CI build jobs, 15/15 checks.

aparajon (approved):
🤖 Approving on behalf of Armand Parajon. Adversarial review posted above — the orientation, the no-launder claim, and the non-vacuity of the new guards all held under mutation probes. One decision worth making before merge (whether `pkg/datasync`'s identical gate gets the same relaxation) plus two doc/comment tightenings; none block.

morgo:
🤖 Sent by Morgan's AI agent.

Thanks — all four addressed in f6140ba6. Nothing about the relaxation itself changed; the findings were about a neighbouring path and two claims that were true but imprecise.

**1. datasync — decided: it stays exact, and now says why.** I went looking for the asymmetry that justifies the two paths disagreeing, and it is the empty-target precondition. move's `target_state` rejects a pre-existing target table holding *any* row, so the only rows at stake are ones not yet copied, and the unconditional pre-cutover checksum vets every one of them. `verifyExistingTargetTable` has no such precondition — it fires on a target that may already be half-copied by an earlier attempt, which is the resumable-checkpoint case its own doc comment describes. A divergence there can mean the rows *already sitting on the target* are wrong, and no schema relaxation can establish otherwise; that is also why it fails rather than repairs. On top of that, sync has a single unsharded target (`KeyRange: "0"`) and no cutover, so neither of move's motivations — sequence-supplied ids, a shard key a primary vindex cannot map from NULL — exists there at all.

So: documented as a decision at the call site, with `TestSyncVerifyExistingTargetTableRequiresExactSchema` pinning both relaxations as rejected on that path. I deliberately did *not* extract a shared `targetDiffOptions()` — that would make the two paths drift *together*, which is the opposite of what we want given they guard different states. If you'd rather sync took the relaxations, that's a separate change with its own reasoning, and the test will make someone write it down.

**2. Doc now costs the failure.** You're right that "fails at the checksum" was doing a lot of hiding. `docs/move.md` now says the initial checker runs with `FixDifferences` and three retries, so each attempt re-copies the chunk, re-coerces the same NULL and finds it again — hours on a table worth moving this way — and contrasts that with the one-line `WHERE col IS NULL LIMIT 1` probe up front.

**3. `INSERT IGNORE`, not batching.** Correct, and it makes the branch total: both copy paths write with `INSERT IGNORE` (`INSERT IGNORE ... SELECT` in the copier, `INSERT IGNORE ... VALUES` in the applier), and `IGNORE` downgrades the would-be `ER_BAD_NULL_ERROR` to a warning and stores the implicit default regardless of row count. Reworded, and I kept your `ISNULL()`-digit observation about `VARCHAR NOT NULL` defaulting to `''` since that's the one case where the `IFNULL` digit alone would collide.

**4. `pkDroppedNullabilityChange` back-reference added**, pointing out it is the same predicate under a different gate and wholly subsumed when the option is on.

Also fixed the Copilot grammar nit — the "table's permits NULL" phrasing is gone, since that whole comment was rewritten to state the direction as reference-vs-validated rather than in terms of `Diff`'s own arguments (which invert it, and had already misled a reader once).

Enabling auto-merge.

morgo:
🤖 Sent by Morgan's AI agent.

@aparajon — correction to what I told you in this thread, since it was load-bearing for the review.

I said the coerced NULL is caught by the initial checksum, and costed it as three `FixDifferences` repair-and-retry rounds. Both wrong. `RetryableTransaction` reads `SHOW WARNINGS` after every statement and fails on any warning it does not explicitly tolerate — `dbconn.UnsafeWarningError` exists precisely because on an `INSERT IGNORE` the warning is the only evidence a row was not stored as read. Every write goes through it (the legacy `INSERT IGNORE ... SELECT` copier was removed in #908, so the only copier streams through the applier), and a fatal error is not retried. The run dies on the batch carrying the row.

Which means your `INSERT IGNORE` note was more right than either of us realised: it is not just that coercion doesn't depend on batching, it is that coercion never survives the statement it happens on. Your second finding — that "fails at the checksum" was hiding the cost — was correct about the doc being evasive, but the honest version is better news than the one I wrote.

The advice to probe up front still stands, for a different reason: unchecked, the failure waits until the copier reaches that chunk (hours on a table this size) and arrives as a warning code against a batch rather than naming the column.

Fixed in #1188, doc-only.

PR #1187, opened 2026-09-02
move/check: export TargetSchemaDiff

morgo:
## Why

Since #1186, "which divergences does a move tolerate" has two definitions.

`move/check.targetSchemaDiff` is one. The other is in strata, which reassembles the same option set by hand so its `keyspace move-tables` can tell an operator whether a move will be accepted **before** they confirm it, rather than an hour into the copy. That only works while the two agree:

- strata drifts stricter → it blocks a move spirit would have run happily;
- strata drifts looser → it promises a move that then dies in spirit's pre-flight, after the operator has already said yes.

Neither failure is visible from either side, and the set is now two flags rather than one, so this is a good moment to stop having two copies of it.

## What

Export the function the checks already use — `check.TargetSchemaDiff` — and state in its doc that the tolerated set is public behaviour, not an internal detail.

**Exported as a function, not as the options it builds.** That is the substantive design point rather than a style preference. The nullability tolerance is directional: `statement.IgnoreNotNullRelaxation` lets the schema being *validated* be stricter than its *reference*, and the target is the validated one, so it has to reach `DiffCreateTables` as `got` with the source as `want`. Handing a caller the options hands them that trap — passed the other way round they forgive the opposite, dangerous direction, silently and with no diff to show for it. (This orientation has already misled a reader once, on #1186.) The parameter names `sourceCreate, targetCreate` are what prevents it, so the option and the argument order stay together inside the one function.

Two doc changes came with it:

- the exported doc no longer says "it is `schemaDiff` plus one relaxation" — a godoc reader cannot follow a reference to an unexported function — and enumerates both tolerated divergences itself;
- `TestTargetSchemaDiffPublicContract` pins the contract external callers actually depend on: an empty diff means "the move will accept this target", and the tolerated set is exactly those two. Eight cases across accept and reject, so a case moving between the lists is a deliberate behaviour change rather than a refactor.

`schemaDiff` (the source↔source comparison, which must stay exact) deliberately stays unexported — nothing outside needs it, and it is the one that must *not* forgive a tightening.

## Not in scope

No behaviour change. Internal callers are unchanged apart from the identifier.

## Testing

`go test ./pkg/move/check/ ./pkg/statement/` and `golangci-lint run ./pkg/move/... ./pkg/statement/... ./pkg/datasync/...` are clean.

### DIFF
--- a/pkg/datasync/runner.go
+++ b/pkg/datasync/runner.go
@@ -1072,7 +1072,7 @@ func (r *Runner) createTargetTables(ctx context.Context) error {
 // the target alone would not make the copy correct.
 //
 // The comparison is deliberately exact, unlike move's source->target check
-// (move/check.targetSchemaDiff), which forgives a target that drops the
+// (move/check.TargetSchemaDiff), which forgives a target that drops the
 // source's column-level AUTO_INCREMENT or is stricter about NULL. Neither
 // relaxation transfers, because the two gates guard different states:
 //
--- a/pkg/move/check/resume_state.go
+++ b/pkg/move/check/resume_state.go
@@ -72,7 +72,7 @@ func resumeStateCheck(ctx context.Context, r Resources, logger *slog.Logger) err
 			// names-only comparison let a target with the same columns but a
 			// different type/charset/collation pass resume validation; on a
 			// resume whose checksum watermark already covers the affected chunk,
-			// that mismatch would never be re-verified. targetSchemaDiff compares
+			// that mismatch would never be re-verified. TargetSchemaDiff compares
 			// types, charset, collation, indexes and constraints while ignoring
 			// AUTO_INCREMENT counters and other instance-specific noise. It is
 			// the same comparison target_state ran before the copy, so a move
@@ -85,7 +85,7 @@ func resumeStateCheck(ctx context.Context, r Resources, logger *slog.Logger) err
 			if err != nil {
 				return fmt.Errorf("failed to read target %d schema for table '%s': %w", i, sourceTable.TableName, err)
 			}
-			diff, err := targetSchemaDiff(sourceTable.TableName, sourceCreate, targetCreate)
+			diff, err := TargetSchemaDiff(sourceTable.TableName, sourceCreate, targetCreate)
 			if err != nil {
 				return fmt.Errorf("failed to compare schema for table '%s' on target %d: %w", sourceTable.TableName, i, err)
 			}
--- a/pkg/move/check/schema_compare.go
+++ b/pkg/move/check/schema_compare.go
@@ -39,30 +39,47 @@ func showCreateTable(ctx context.Context, db *sql.DB, schema, table string) (str
 //
 // This is the comparison for two schemas that must be IDENTICAL — today
 // sources[0] against every other source (source_schema_consistency). Use
-// targetSchemaDiff for a source→target comparison, which additionally forgives
+// TargetSchemaDiff for a source→target comparison, which additionally forgives
 // a target that is deliberately stricter.
 func schemaDiff(table, wantCreate, gotCreate string) (string, error) {
 	return statement.DiffCreateTables(table, wantCreate, gotCreate, moveDiffOptions())
 }
 
-// targetSchemaDiff compares a move's SOURCE table against a pre-created TARGET
-// table, and is the comparison the source→target checks use (target_state,
-// resume_state). It is schemaDiff plus one further relaxation, in one
-// direction only: a target column may be NOT NULL where the source permits
-// NULL.
+// TargetSchemaDiff compares a move's SOURCE table against a pre-created TARGET
+// table and returns a runnable ALTER TABLE statement describing how they differ,
+// or an empty string if the move will accept the target as it stands. It is the
+// comparison the source→target checks use (target_state, resume_state).
 //
-// That is what statement.IgnoreNotNullRelaxation does: it lets the schema being
-// validated be stricter than its reference. Here the target is the validated
-// schema, so it must be passed as DiffCreateTables' "got" and the source as
-// "want" — the other way round would forgive the opposite, dangerous direction,
-// which is why the option and the argument order stay together in this one
-// function.
+// Two divergences are tolerated, and only these two. Both are a target that is
+// deliberately stricter or leaner than the unsharded source it moves from, so
+// that a declaratively-managed target does not have to mirror artifacts of its
+// source:
 //
-// A sharded target needs this. A Vitess primary vindex cannot map NULL to a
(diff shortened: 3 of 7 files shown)


aparajon:
🤖 Adversarial correctness review, requested by Armand Parajon and performed by his agent. Reviewed at head `c4f92f8f`.

**Verdict: the motivation is exactly right — two definitions of "which divergences does a move tolerate" is a bug generator, and neither drift direction is visible from either side — and the mechanical change is a clean no-op. Safe to land. But the doc this PR promotes to a public contract is not an accurate enumeration of what the function tolerates, and that inaccuracy matters more after this PR than before it**, because a previewing caller now builds against the doc rather than against the code. I went after the contract claim rather than the rename, and it does not hold.

### Findings

**1. "Two divergences are tolerated, and only these two" is five, and the AUTO_INCREMENT half is symmetric rather than directional.** `moveDiffOptions()` is `NewDiffOptions()` plus `IgnoreColumnAutoIncrement`, and `NewDiffOptions()` already defaults `IgnoreAutoIncrement`, `IgnoreEngine` and `IgnoreRowFormat` to true. I probed the exported function directly against the doc's two claims, and four targets the doc does not cover come back accepted with an empty diff:

| Target divergence from source | `TargetSchemaDiff` verdict |
|---|---|
| `ENGINE=MyISAM` | **accepted** |
| `ROW_FORMAT=COMPRESSED` | **accepted** |
| `AUTO_INCREMENT=90210` (table counter) | accepted — cosmetic, but undocumented here |
| target *adds* column `AUTO_INCREMENT` the source lacks | **accepted** |

The last one contradicts the sentence "Anything else is a difference, the reverse of either included": `IgnoreNotNullRelaxation` is directional, `IgnoreColumnAutoIncrement` is not, so the doc's symmetry claim is true of one tolerance and false of the other. And `ENGINE` is the one whose acceptance isn't cosmetic — `grep -rni engine pkg/move/` finds no other engine check anywhere in the move path, so `TargetSchemaDiff` is the only gate on the target's shape and a pre-created non-InnoDB target passes pre-flight silently. `IgnoreEngine: true` is a reasonable default for spirit's own migrate path, where the new table is derived from the source and the engine matches by construction; the move's target is authored independently, which is the one place that default doesn't carry over.

This is a doc/coverage finding rather than a regression — every one of those has been accepted since before #1186. What changes here is that the sentence becomes the spec an external previewer implements against, and `TestTargetSchemaDiffPublicContract` pins the two named divergences while saying nothing about table options, so the enumeration and the test corroborate each other and not the code. That is the failure mode the PR exists to prevent, one layer up: not strata drifting from spirit, but the documented contract drifting from the function.

**2. `pkg/statement` is a better home for it than `pkg/move/check`.** Not a re-litigation of function-vs-options — that argument is right, and it survives this suggestion completely intact, since the parameter names and the argument order still travel together inside the one function. It's about which package an external caller has to import. `TargetSchemaDiff` is pure `(table, sourceCreate, targetCreate string) → (string, error)`, but `go list -deps ./pkg/move/check` is 18 spirit packages — `dbconn`, `applier`, `autoscale`, `metrics`, `table` — against `pkg/statement`'s 12. A caller that wants one string comparison to preview a move links move's runtime check machinery and a MySQL connection layer to get it. Meanwhile `pkg/statement/diff.go` already holds the move-specific knowledge: `IgnoreNotNullRelaxation`'s doc names `move/check.TargetSchemaDiff`, sharded targets, Vitess primary vindexes and `ColumnMapping.ChecksumExprs`, and explains the got→want inversion. Everything except the function is already there. Hosting it there would also leave `move/check` with no exported API at all, which is the more natural shape for a package of pre-flight checks.

**3. (nit) The contract test's reject list is missing the case the contract most needs pinned.** The looser-nullability rejection — a target that drops a `NOT NULL` the source had, the direction `IgnoreNotNullRelaxation` must never forgive — lives in `TestTargetSchemaDiffRejectsNullableTarget` rather than in the table. The table's comment promises that "a case moving between these two lists is a deliberate public behaviour change"; the case where that promise carries the most weight is the one in neither list. Moving it in costs six lines and makes the table self-contained.

### Action items

1. **(Finding 1)** Enumerate all five tolerated divergences in the exported doc, and drop or qualify "the reverse of either included" — it is true of the nullability tolerance and false of the AUTO_INCREMENT one.
2. **(Finding 1)** Add rows to `TestTargetSchemaDiffPublicContract` for ENGINE, ROW_FORMAT, the table-level counter and the reverse-AUTO_INCREMENT case, so the test pins the whole list rather than a subset of it.
3. **(Finding 1, separable)** Decide `IgnoreEngine` on its own merits for the source→target comparison. Nothing else in `pkg/move` checks it, and unlike the other four this one can change how the target behaves.
4. *(optional)* **(Finding 2)** Consider `pkg/statement` as the export site.
5. *(optional)* **(Finding 3)** Move the looser-nullability reject into the contract table.

The rename really is a no-op: the function body is byte-identical apart from the identifier, `moveDiffOptions()` stays unexported, and `go doc ./pkg/move/check` confirms `schemaDiff` — the source↔source comparison that must *not* forgive a tightening — stays unexported too, so nothing outside can reach the strict comparison or the options. Every internal call site moved (`target_state.go:116`, `resume_state.go:88`, and the four test call sites); `grep` finds no surviving `targetSchemaDiff` reference outside doc prose, and the three cross-package doc references in `pkg/statement/diff.go`, `pkg/statement/diff_test.go` and `pkg/datasync/runner.go` were all updated rather than left dangling — which is the failure mode a rename usually leaves behind. The new godoc's substantive claims check out where they are checkable: charset and collation really are compared (`IgnoreCharsetCollation` defaults false) at both table and column level, index and constraint changes are reported, and the four reject cases in the contract table all reject. The carried-over paragraphs from #1186 are unchanged in substance, including the pre-cutover-checksum safety argument and the advice to probe tightened columns before starting a copy. `go build ./...` clean; `./pkg/move/check` and `./pkg/statement` pass locally, including the new eight-case table; no test or assertion was deleted, and the four edited tests changed only the identifier. 15/15 CI checks green. Copilot recommended approval with no inline comments, so there is nothing outstanding on that thread.

aparajon (approved):
🤖 Approving on behalf of Armand Parajon. Adversarial review posted above — the export is a verified no-op and the motivation holds. One thing worth fixing before or right after merge: the new exported doc says the tolerated set is exactly two divergences, and it is five (ENGINE, ROW_FORMAT and the table-level AUTO_INCREMENT counter come in from the `NewDiffOptions()` defaults), plus the column-AUTO_INCREMENT tolerance is symmetric rather than directional. The contract test pins the two named ones, so doc and test agree with each other and not with the function.

PR #1191, opened 2026-09-03
check: report a scope's membership so callers can pin it

aparajon:
`ScopeStatement` says a check added to the scope later is picked up without any change on the caller's side. That is convenient for a caller running the checks and hazardous for one repeating what they say.

SchemaBot publishes a statement-scope refusal reason onto the pull request the schema change was submitted on. It decided that was safe by reading the checks in the scope today: each one reports the submitted statement back, and the statement is already on the pull request. A check that interpolated something else would inherit that verdict through a version bump, with no diff on the consuming side to review.

`ChecksInScope` reports the names a scope carries, in the order `RunChecks` runs them, so a caller can pin the set its judgement was made about and have a bump fail rather than widen it quietly. `RunChecks` now selects through the same helper, so the two cannot disagree about what a scope holds or the order it runs in.

The pin lands here as well. Scope membership is already a curated set — `ScopeStatement` takes only checks no earlier stage can bypass, and deliberately excludes the preflight checks MySQL's native DDL can complete — but nothing held a new registration to that contract. Adding a name to the scope now means adding it to the list, in the same change that decides it belongs there.

Same shape as #1187: a consumer relying on something this package knows, given a way to ask instead of reassembling it by hand.

*Authored by Claude Opus 5 via Claude Code.*


morgo (approved):
🤖 Reviewed by my AI agent.

`RunChecks` is behavior-preserving: it used to sort all registered names then filter by scope, now it filters then sorts — same set, same order. `ChecksInScope(ScopeNone)` is empty by construction since `scope&0 == 0`, which the test asserts rather than assumes.

Verified the pinned `ScopeStatement` list against the registrations at this head — addforeignkey, enumReorder, enumSetRemoval, illegalClause, primarykey, primarykeyexists, setReorder — exactly the 7 checks tagged `ScopeStatement`, and `version` is still the only `ScopePreRun` check. Built and ran the package tests locally; green.

The membership pin is the right shape for the SchemaBot problem: a new registration now fails a test in the same package that decides it belongs in the scope, instead of widening a published refusal reason through a version bump with no diff downstream.

Nit, non-blocking: `ChecksInScope` clones the whole map to read names; it could hold the lock over `namesInScope` directly. Not worth a round trip.

Approving and enabling auto-merge.

Issue #418, opened 2025-06-25
Constraint names are not preserved in new versions of tables


kolbe:
I've observed that user-provided constraint names are not preserved when Spirit creates the new table to load data into.

I will add a test case soon.

kolbe:
This is expected behavior, because MySQL requires that `The CONSTRAINT symbol value, if defined, must be unique in the database`.

```
mysql> create table t1 (i int, constraint i check (i<>0));
Query OK, 0 rows affected (0.01 sec)

mysql> create table t2 (i int, constraint i check (i<>0));
ERROR 3822 (HY000): Duplicate check constraint name 'i'.
```

MySQL `RENAME TABLE` has the [additional behavior](https://dev.mysql.com/doc/refman/8.4/en/rename-table.html):

> RENAME TABLE tbl_name TO new_tbl_name changes internally generated foreign key constraint names and user-defined foreign key constraint names that begin with the string “tbl_name_ibfk_” to reflect the new table name.

Therefore, the most reasonable "solution" to this problem is so simply omit the symbol/name when you define the constraint and let MySQL do its own thing.

```
mysql> create table t2 (i int, constraint check (i<>0));
Query OK, 0 rows affected (0.01 sec)

mysql> show create table t2\G
*************************** 1. row ***************************
       Table: t2
Create Table: CREATE TABLE `t2` (
  `i` int DEFAULT NULL,
  CONSTRAINT `t2_chk_1` CHECK ((`i` <> 0))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
1 row in set (0.00 sec)

mysql> create table t3 like t2;
Query OK, 0 rows affected (0.01 sec)

mysql> show create table t3\G
*************************** 1. row ***************************
       Table: t3
Create Table: CREATE TABLE `t3` (
  `i` int DEFAULT NULL,
  CONSTRAINT `t3_chk_1` CHECK ((`i` <> 0))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
1 row in set (0.00 sec)

mysql> drop table t2;
Query OK, 0 rows affected (0.01 sec)

mysql> rename table t3 to t2;
Query OK, 0 rows affected (0.01 sec)

mysql> show create table t2\G
*************************** 1. row ***************************
       Table: t2
Create Table: CREATE TABLE `t2` (
  `i` int DEFAULT NULL,
[...]

PR #1193, opened 2026-09-04
migrate: reject an ambiguous DROP CONSTRAINT rather than dropping the CHECK

### LINKED ISSUE #1192: migrate: an ambiguous DROP CONSTRAINT silently drops the CHECK constraint instead of erroring
## Summary

A table can legally hold the same symbol in more than one constraint namespace. MySQL rejects a `DROP CONSTRAINT` that names such a symbol; spirit silently picks the CHECK constraint and reports success.

Reproduced against `350e4e6` (tip of main at time of filing), MySQL 8.0.43.

## Repro

```sql
CREATE TABLE t (
  id INT PRIMARY KEY,
  a INT,
  b INT,
  CONSTRAINT same UNIQUE (a),
  CONSTRAINT same CHECK (b > 0)
);
INSERT INTO t VALUES (1,1,1),(2,2,2),(3,3,3);
```

MySQL refuses to guess:

```
mysql> ALTER TABLE t DROP CONSTRAINT same;
ERROR 3939 (HY000): Table has multiple constraints with the name 'same'.
Please use constraint specific 'DROP' clause.
```

Spirit applies it:

```
$ spirit migrate --database amb --statement "ALTER TABLE t DROP CONSTRAINT same"
INFO unable to use INPLACE error="statement contains operations that are not safe for INPLACE algorithm"
INFO rewrote CHECK constraint names for the new table table=_t_new alter="DROP CONSTRAINT `_t_new_chk_1`"
INFO apply complete instant-ddl=false inplace-ddl=false
```

```
mysql> SHOW CREATE TABLE t;
CREATE TABLE `t` (
  `id` int NOT NULL,
  `a` int DEFAULT NULL,
  `b` int DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `same` (`a`)
)
```

The CHECK constraint is gone, the unique key is retained, and the exit status is 0.

## Why

1. `attemptInstantDDL` does run the statement on the user's table and does get error 3939 — but a direct-DDL failure is the ordinary signal to fall through to the copy algorithm, so `Runner.Run` discards it.
2. The copy applies the ALTER to `_t_new`, where `CREATE TABLE .. LIKE` has renamed the check constraint to `_t_new_chk_1` (#418) and so left only one `same` behind.
3. `newTableAlter` (#1181) retargets the `DROP CONSTRAINT` at that renamed check constraint. On `_t_new` the name is unambiguous, so MySQL accepts it, and cutover installs a table that neither MySQL's semantics nor the user's statement asked for.

The clash does not need a contrived name: MySQL keeps CHECK, FOREIGN KEY, UNIQUE and PRIMARY KEY symbols in separate namespaces, and matches them for `DROP CONSTRAINT` case-insensitively. `UNIQUE KEY SAME` + `CONSTRAINT same CHECK` is error 3939 too, and so is a CHECK constraint named `PRIMARY`. A non-unique index of the same name is *not* a clash — `DROP CONSTRAINT` does not resolve against one.

The foreign key namespace is unreachable here: `hasforeignkeys` refuses an FK-bearing table first.

## Expected

Spirit should refuse the statement, as MySQL does, and say which constraint specific clause to use instead.

morgo:
Fixes #1192.

## The bug

A table can legally hold the same symbol in more than one constraint namespace — `UNIQUE KEY dup` alongside `CONSTRAINT dup CHECK (..)` — and MySQL refuses to guess which one a `DROP CONSTRAINT` means:

```
mysql> ALTER TABLE t DROP CONSTRAINT same;
ERROR 3939 (HY000): Table has multiple constraints with the name 'same'.
Please use constraint specific 'DROP' clause.
```

Spirit ran it and reported success, having dropped the CHECK constraint and kept the unique key. Three things line up:

1. `attemptInstantDDL` does run the statement on the user's table, and does get error 3939 — but a direct-DDL failure is the ordinary signal to fall through to the copy algorithm, so `Runner.Run` discards it. It cannot tell this apart from "not INSTANT-able".
2. `CREATE TABLE .. LIKE` renames the check constraint on `_t_new` (#418), so only one `same` is left there.
3. `newTableAlter` (#1181) retargets the `DROP CONSTRAINT` at that renamed check constraint. On `_t_new` the name resolves, MySQL accepts it, and cutover installs a table the statement never described.

Full repro in #1192.

## The change

`newTableAlter` now refuses a `DROP CONSTRAINT` whose name the source table holds in more than one namespace, and names the clause that disambiguates it:

```
spirit: error: table t has multiple constraints named same (CHECK constraint, UNIQUE),
so DROP CONSTRAINT same is ambiguous and MySQL rejects it: name the one you mean with
a constraint specific clause such as DROP CHECK or DROP KEY
```

* `statement.GenericConstraintDrops` returns the names carried by `DROP CONSTRAINT` clauses only — the subset of `CheckConstraintsReferenced` that does not say which kind of constraint it means. `DROP CHECK` and `ALTER CHECK` do say, so they can never be ambiguous and are excluded.
* `rejectAmbiguousConstraintDrop` matches each of those names, case-insensitively as MySQL does, against the source table's CHECK constraints and its unique and primary keys. `GetIndexes` reports the primary key as `PRIMARY`, which is the symbol MySQL matches it under. A non-unique index of the same name is deliberately not a clash: `DROP CONSTRAINT` does not resolve against one (verified — MySQL drops the CHECK constraint without complaint). The foreign key namespace is unreachable: `hasforeignkeys` refuses an FK-bearing table first, which I confirmed.
* `checkConstraints` was reading `SHOW CREATE TABLE` itself; it now takes an already-parsed definition, and `tableDefinition` does the read. Same two round trips as before, with the source definition shared between the ambiguity check and the rename pairing.

I did consider the smaller fix of not swallowing error 3939 in `attemptMySQLDDL`. It relies on classifying error codes out of a path whose whole purpose is to fail, gives the user MySQL's message with no indication of which of their statements it came from, and would not cover a case where the direct-DDL attempt is skipped. The explicit check is self-describing and testable without a live failure.

## Testing

Both new migration tests were confirmed to fail on pre-fix behaviour by deleting the guard call and re-running — `An error is expected but got nil` for each — and the third passes either way, which is the point of it.

* `pkg/migration/TestDropConstraintAmbiguousWithUniqueKey` — runs the ALTER through MySQL first to pin the premise (error 3939), then asserts spirit refuses it, that neither constraint was dropped, and that `DROP CHECK dup, ENGINE=InnoDB` still goes through.
* `pkg/migration/TestDropConstraintAmbiguousWithPrimaryKey` — the same clash against the primary key, which `SHOW CREATE TABLE` prints no name for.
* `pkg/migration/TestDropConstraintResolvesToUniqueKey` — the case the check must not fire on: the name is held by a unique key only, on a table that does have check constraints. Copy path, unique key dropped, check constraint still enforced.
* `pkg/statement/TestGenericConstraintDrops` — which clause types contribute a name.

Green locally against MySQL 8.0.43: `pkg/statement/...`, `pkg/lint/...`, and `pkg/migration/ -run 'CheckConstraint|DropConstraint'`. `golangci-lint run ./...` reports `0 issues.`

## Scope

This rejects the ambiguous statement rather than making it work. Resolving it faithfully would mean deciding which namespace the user meant, which is exactly what MySQL declines to do.

### DIFF
--- a/pkg/migration/change.go
+++ b/pkg/migration/change.go
@@ -116,7 +116,18 @@ func (c *tableChange) newTableAlter(ctx context.Context) (string, error) {
 	if len(c.stmt.CheckConstraintsReferenced()) == 0 {
 		return c.stmt.TrimAlter(), nil
 	}
-	renames, err := c.checkConstraintRenames(ctx)
+	source, err := tableDefinition(ctx, c.runner.db, c.table.TableName)
+	if err != nil {
+		return "", err
+	}
+	if err := rejectAmbiguousConstraintDrop(c.stmt, source, c.table.TableName); err != nil {
+		return "", err
+	}
+	newTable, err := tableDefinition(ctx, c.runner.db, c.newTable.TableName)
+	if err != nil {
+		return "", err
+	}
+	renames, err := c.checkConstraintRenames(source, newTable)
 	if err != nil {
 		return "", err
 	}
@@ -137,6 +148,52 @@ func (c *tableChange) newTableAlter(ctx context.Context) (string, error) {
 	return alter, nil
 }
 
+// rejectAmbiguousConstraintDrop refuses an ALTER whose DROP CONSTRAINT names a
+// constraint MySQL itself would not resolve, before newTableAlter retargets it
+// at the check constraint of that name and quietly makes the migration mean
+// something the statement does not.
+//
+// A table may hold the same symbol in more than one constraint namespace -
+// UNIQUE KEY `dup` alongside CONSTRAINT `dup` CHECK (..) is legal - and MySQL
+// rejects a DROP CONSTRAINT that cannot tell them apart: "Table has multiple
+// constraints with the name 'dup'" (error 3939).
+//
+// Spirit does not inherit that rejection. The direct-DDL attempt does get
+// error 3939 back from the user's table, but a direct-DDL failure is the
+// ordinary signal to use the copy algorithm instead, so it is discarded. The
+// copy then applies the ALTER to the _new table, where CREATE TABLE .. LIKE
+// has renamed the check constraint (see newTableAlter) and left only one `dup`
+// behind: the clause resolves there and drops the check constraint while
+// keeping the unique key - a table the user never asked for, reported as
+// success.
+//
+// The foreign key namespace is not consulted: spirit refuses a table with
+// foreign keys long before this point.
+func rejectAmbiguousConstraintDrop(stmt *statement.AbstractStatement, def *statement.CreateTable, tableName string) error {
+	for _, name := range stmt.GenericConstraintDrops() {
+		var owners []string
+		for _, constraint := range def.GetConstraints() {
+			if constraint.Type == "CHECK" && strings.EqualFold(constraint.Name, name) {
+				owners = append(owners, "CHECK constraint")
+			}
+		}
+		for _, index := range def.GetIndexes() {
+			// MySQL resolves DROP CONSTRAINT against unique and primary keys
+			// only, so a non-unique index of the same name is not a clash.
+			// GetIndexes names the primary key "PRIMARY", which is the symbol
+			// MySQL matches it under.
+			if (index.Type == "UNIQUE" || index.Type == "PRIMARY KEY") && strings.EqualFold(index.Name, name) {
+				owners = append(owners, index.Type)
+			}
+		}
+		if len(owners) > 1 {
+			return fmt.Errorf("table %s has multiple constraints named %s (%s), so DROP CONSTRAINT %s is ambiguous and MySQL rejects it: name the one you mean with a constraint specific clause such as DROP CHECK or DROP KEY",
+				tableName, name, strings.Join(owners, ", "), name)
+		}
+	}
+	return nil
+}
+
 // checkConstraintRenames maps each check constraint name on the table being
 // altered (lower-cased, because MySQL matches these names case-insensitively)
 // to the name the same constraint has on the _new table.
@@ -149,15 +206,9 @@ func (c *tableChange) newTableAlter(ctx context.Context) (string, error) {
 // back in numeric order. Their expressions and enforcement are compared
 // afterwards to confirm the pairing, rather than trusting the ordering and
 // retargeting a DROP CHECK at some other constraint.
-func (c *tableChange) checkConstraintRenames(ctx context.Context) (map[string]string, error) {
-	source, err := checkConstraints(ctx, c.runner.db, c.table.TableName)
-	if err != nil {
-		return nil, err
-	}
-	newTable, err := checkConstraints(ctx, c.runner.db, c.newTable.TableName)
-	if err != nil {
-		return nil, err
-	}
+func (c *tableChange) checkConstraintRenames(sourceDef, newTableDef *statement.CreateTable) (map[string]string, error) {
+	source := checkConstraints(sourceDef)
+	newTable := checkConstraints(newTableDef)
 	if len(source) != len(newTable) {
 		return nil, fmt.Errorf("table %s has %d CHECK constraint(s) but its copy %s has %d",
 			c.table.TableName, len(source), c.newTable.TableName, len(newTable))
@@ -204,9 +255,8 @@ func sortByGeneratedNumber(constraints statement.Constraints, tableName string) 
 	return nil
 }
 
-// checkConstraints returns the table's CHECK constraints as SHOW CREATE TABLE
-// lists them, which is sorted by name.
-func checkConstraints(ctx context.Context, db *sql.DB, tableName string) (statement.Constraints, error) {
+// tableDefinition returns the table's SHOW CREATE TABLE, parsed.
+func tableDefinition(ctx context.Context, db *sql.DB, tableName string) (*statement.CreateTable, error) {
 	var name, createTable string
 	if err := db.QueryRowContext(ctx,
 		sqlescape.MustEscapeSQL("SHOW CREATE TABLE %n", tableName)).Scan(&name, &createTable); err != nil {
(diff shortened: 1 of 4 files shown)


aparajon:
🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head `27ed444e`.

**Verdict: the diagnosis is exactly right and the guard is in the right place — refusing in `newTableAlter` rather than un-swallowing 3939 in the direct-DDL path is the correct call for the reasons the summary gives.** The three tests are unusually good, and pinning the premise by running the ALTER through MySQL first is the right instinct. My findings are all in the same place: the parts of the predicate that are *deliberate policy* are the parts no test can see, and their regression mode is a false refusal of a working schema change rather than the silent wrong-drop this fixes.

| # | Finding | Severity |
|---|---|---|
| 1 | The non-unique-index non-clash — the one deliberate policy call in the guard — has no test | medium |
| 2 | Case-insensitive matching is claimed twice and pinned nowhere; the owners list in the message is also unasserted | medium |
| 3 | The primary-key test does not pin its premise against MySQL the way its sibling does | low |

### Findings

**1. "A non-unique index of the same name is deliberately not a clash" is the only judgement call in `rejectAmbiguousConstraintDrop`, and nothing tests it.** The summary says it was verified against MySQL by hand, and I believe it — but dropping `index.Type == "UNIQUE" || index.Type == "PRIMARY KEY"` so that any index of the name counts leaves the whole migration package green. I checked: no test in `pkg/migration` has a non-unique index sharing a name with a check constraint. The regression is worse than the usual unpinned-branch case, because it converts a schema change MySQL executes happily into a hard refusal — and the refusal names the guidance `DROP CHECK or DROP KEY`, so the user is told to disambiguate something that was never ambiguous. A fourth table (`KEY dup (name)` alongside `CONSTRAINT dup CHECK (val > 0)`, migration succeeds, check constraint gone, index still there) turns the hand-verification into a standing one.

**2. `strings.EqualFold` is load-bearing by both the code comment and the summary, and every test matches case exactly.** All five `DROP CONSTRAINT` statements across `check_constraint_test.go` and `drop_constraint_test.go` spell the name the same way the table does, so replacing both `strings.EqualFold` calls with `==` survives. MySQL matches these names case-insensitively, which is the whole reason `checkConstraintRenames` lower-cases its keys a few lines below — so the code is right and simply undefended. `DROP CONSTRAINT DUP` in the unique-key test covers both call sites at once.

Related, and fixable in the same line: `require.ErrorContains(t, err, "ambiguous")` asserts only that *some* ambiguity was reported. The owners list is what makes the message actionable, and it is unasserted — appending a constant instead of `index.Type` still yields two owners and still says "ambiguous", so the message could name the wrong constraint kinds without a test noticing. Asserting the full sentence (`has multiple constraints named dup (CHECK constraint, UNIQUE)`) pins the composition and the ordering.

**3. `TestDropConstraintAmbiguousWithPrimaryKey` asserts Spirit's behaviour without first establishing MySQL's.** Its sibling does this exactly right — `require.ErrorContains(t, err, "multiple constraints with the name")` against the live server before asking what Spirit does — and the PK test omits it. That matters here more than it usually would: a primary key's symbol is not user-chosen, so the clash requires a check constraint literally named `` `PRIMARY` ``, and whether MySQL treats ``DROP CONSTRAINT `PRIMARY` `` as ambiguous rather than resolving it to the only named constraint is precisely the fact the branch depends on. The "confirmed to fail pre-fix" evidence cannot distinguish the two cases: if MySQL resolves it, the pre-fix behaviour was *correct* and this branch is a new false refusal, and the test would still have gone from pass to fail. I could not settle it here — the MySQL on this machine is not the Spirit test instance — but the one line that settles it is already written in the test above.

### Action items

1. **(Finding 1)** Add a non-unique-index case: `KEY dup (name)` plus `CONSTRAINT dup CHECK (val > 0)`, `DROP CONSTRAINT dup` succeeds, the check constraint is gone and the index remains.
2. **(Finding 2)** Spell one test's `DROP CONSTRAINT` in a different case from the table (`DROP CONSTRAINT DUP`), and assert the whole error sentence including `(CHECK constraint, UNIQUE)` rather than the word `ambiguous`.
3. **(Finding 3)** Add the premise assertion from the unique-key test to the primary-key test, so the branch rests on observed MySQL behaviour rather than on inference.

<details>
<summary><b>Verified (including two attacks that dissolved)</b></summary>

The clause vocabulary is right, and `TestGenericConstraintDrops` covers it properly — `DROP CHECK`, `ALTER CHECK`, `DROP FOREIGN KEY`, `DROP KEY`, `ADD CONSTRAINT` and a non-ALTER all yield nil, and multi-spec and trailing-option forms both yield their names. `./pkg/statement/` passes locally at this head. `ast.AlterTableDropConstraint` is the correct discriminator now that the parser keeps the `DROP CONSTRAINT` spelling distinct from `DROP CHECK`, and the guard is reachable for every `DROP CONSTRAINT` because such a clause counts toward `CheckConstraintsReferenced()`, so `newTableAlter`'s early return cannot skip it — including on a table with no check constraints at all.

The two index types are the real ones: `GetIndexes` emits `"PRIMARY KEY"` and `"UNIQUE"` verbatim and fills the empty primary-key name with `"PRIMARY"`, which is the symbol MySQL matches it under, so both comparisons in the loop line up with the strings they are checking against.

**Attack that dissolved.** `GetIndexes`'s column-level branch synthesises a placeholder name for an inline `UNIQUE` — `"UNIQUE " + col.Name`, with its own comment saying the real name is the server's — which would never `EqualFold`-match a real symbol and so would silently miss a clash. It cannot reach this code: `tableDefinition` parses `SHOW CREATE TABLE`, which always renders an inline `UNIQUE` or `PRIMARY KEY` as a table-level clause with the server-assigned name. The primary-key test relies on that same normalisation, which is a nice consistency.

**Second attack that dissolved.** A refusal now happens after `_new` exists, so I expected a leaked `_chk_ambig_new` to poison the next run. The unique-key test incidentally disproves it: it runs a second, successful migration against the same table immediately after the refusal, which would collide if the artifact survived `Close()`.

The `owners` list cannot contain duplicates of a kind — MySQL forbids two check constraints or two indexes sharing a name within a table — so `len(owners) > 1` genuinely means "more than one namespace", and the foreign-key omission is sound because an FK-bearing table is refused well before this point. Mutating the threshold to `>= 1` breaks `TestDropConstraintResolvesToUniqueKey` and the pre-existing `drop_constraint_test.go` cases, and deleting either the CHECK loop or the unique branch breaks one of the ambiguity tests, so the load-bearing shape of the predicate is pinned even though its policy edges are not.

The `checkConstraints` refactor is a clean win rather than incidental churn: the source definition is now parsed once and shared between the ambiguity check and the rename pairing, the round-trip count is unchanged at two, and `tableDefinition` has one job. Ordering is right too — the source read and the refusal come before the `_new` read, so an ambiguous statement fails before the extra work.

Docker is unavailable here and the local MySQL is not the Spirit test instance, so the migration-package mutation analysis is by inspection of the assertions and a grep of every `DROP CONSTRAINT` in the package rather than by running the suite; CI is green at this head (15/15). `./pkg/statement/` I did run.

</details>

aparajon (approved):

PR #1194, opened 2026-09-04
Don't record a sparse chunk as a measured chunk size

morgo:
## Problem

`prePrefetchChunkSize` (#1184) exists so that leaving a prefetch episode restores a chunk size that real rows were once read at, rather than the `MaxDynamicRowSize` ceiling the gap itself bought. `recordKeyDensity` decides what qualifies as "measured", and the test was `rows > 0`:

```go
if rows > 0 && keys == chunk.ChunkSize {
    t.lastDenseChunkSize = chunk.ChunkSize
}
```

Over a gap that is **sparse rather than empty**, that is not the same question. A chunk holding a handful of rows across its whole width measures the cost of those few rows, not of a full chunk of that size — and thinly-populated chunks are exactly what carry the sizer to the ceiling, so accepting them records the ceiling as "measured" from the very evidence that makes prefetch fire. One row per chunk is enough to "prove" a 100k-row chunk that no full chunk was ever read at.

`TestOptimisticPrefetchRestoreIsMeasured` covers the *empty* gap, where `lastDenseChunkSize` correctly stays at the pre-gap size. The sparse gap was uncovered.

## Why it matters

The chunker that gets `TargetChunkBytes` is the copy chunker — `runner.go`:

```go
copyChunkerCfg.TargetChunkBytes = r.migration.TargetChunkSize
// This applies to the copy chunker only: the checksum runs server-side CRC and keeps the time signal.
```

So this is the buffered copier's path, and `readChunkData` appends every row into `rowDataList` with no in-read byte cap. With 16KB rows against the 16MiB default budget the restore is 100,000 rows — ~1.5GB, fully materialized before any `ActualBytes` feedback exists to shrink it. `panicShrinkBytes` (5 × 16MiB) only reacts after that chunk has already been read, and with several read workers per migration.

The checksum chunker is unaffected: it keeps the time signal and its CRC is server-side, so no rows reach the client.

## Reproducer

The new test is a byte-signal chunker, 16KB rows: phase 1 is 60 full chunks, phase 2 is a gap where every chunk carries 150 real rows however wide it gets — under the entry gate (2.34MiB × 5 < 16MiB) and well over `minKeysPerRowForPrefetch`, so prefetch legitimately engages. Before this change:

```
phase 1: chunkSize=1024 lastDenseChunkSize=1024
phase 2: prefetching=true lastDenseChunkSize=100000 prePrefetchChunkSize=100000
```

Phase 1 measures honestly; phase 2 overwrites it with the ceiling.

## Fix

Require the chunk to have come back at least half full before its size counts:

```go
if rows*minChunkFullnessDivisor >= chunk.ChunkSize && keys == chunk.ChunkSize {
```

Half rather than full because real tables have small gaps throughout, and demanding every key hold a row would record nothing — making every prefetch exit fall back to the `StartingChunkSize` ramp. Half bounds what a restored size can understate by 2x, which the sizer's own feedback absorbs in one window and which sits well inside `panicShrink`'s 5x threshold.

After: `prePrefetchChunkSize=1024`, i.e. 16MiB against a 16MiB budget — the size the pre-gap chunks were actually measured at — while prefetch still engages.

For the record, a density-based gate (`keys/rows < minKeysPerRowForPrefetch`) is *not* sufficient: at intermediate ramp widths 11664/150 = 77.7 still passes, and the restore only comes down to 11,664 rows / 182MiB. Fullness is the property that matters here, not density.

## Testing

New `TestOptimisticPrefetchRestoreIgnoresSparseChunks` fails on `main` (`expected 1024, actual 100000`) and passes with the change. `./pkg/table/`, `./pkg/copier/` and `./pkg/checksum/` all pass.

### DIFF
--- a/pkg/table/chunker_optimistic.go
+++ b/pkg/table/chunker_optimistic.go
@@ -85,6 +85,18 @@ var _ MappedChunker = &chunkerOptimistic{}
 // See chunkerOptimistic.prefetchRejections.
 const maxPrefetchRejections = 2
 
+// minChunkFullnessDivisor is how full a fixed-width chunk must come back before
+// its size counts as measured and becomes eligible for prefetch to restore: at
+// least one row per this many keys of width, i.e. half full.
+//
+// Half rather than full because real tables have small gaps throughout, and
+// requiring every key to hold a row would mean nothing is ever recorded and
+// leaving prefetch would always fall back to the StartingChunkSize ramp. Half
+// bounds the cost a restored size can understate by 2x, which the sizer's own
+// feedback absorbs in one window and which sits well inside panicShrink's 5x
+// threshold. See chunkerOptimistic.lastDenseChunkSize.
+const minChunkFullnessDivisor = 2
+
 // nextChunkByPrefetching uses prefetching instead of feedback to determine the chunk size.
 // It is used when the chunker detects that there are very large gaps in the sequence.
 // When this mode is enabled, the chunkSize is "reset" to 1000 rows, so we know that
@@ -484,7 +496,17 @@ func (t *chunkerOptimistic) recordKeyDensity(chunk *Chunk, actualRows uint64) {
 	// chunk (next() builds those as chunkPtr..chunkPtr+chunkSize) and excludes a
 	// prefetch chunk, whose ChunkSize is a row offset while its width is the gap
 	// it crossed — restoring a prefetch offset would defeat the whole point.
-	if rows > 0 && keys == chunk.ChunkSize {
+	//
+	// The chunk also has to have come back reasonably *full*, not merely
+	// non-empty. A chunk holding a handful of rows across its whole width
+	// measures the cost of those few rows, not of a full chunk of that size —
+	// and thinly-populated chunks are exactly what carry the sizer to the
+	// ceiling to begin with, so accepting them would validate the ceiling from
+	// the very evidence that makes prefetch fire. That is the case
+	// prePrefetchChunkSize exists to exclude, and `rows > 0` does not exclude
+	// it: over a gap that is sparse rather than empty, one row per chunk is
+	// enough to "prove" a 100k-row chunk that no full chunk was ever read at.
+	if rows*minChunkFullnessDivisor >= chunk.ChunkSize && keys == chunk.ChunkSize {
 		t.lastDenseChunkSize = chunk.ChunkSize
 	}
 }
--- a/pkg/table/chunker_optimistic_test.go
+++ b/pkg/table/chunker_optimistic_test.go
@@ -922,6 +922,64 @@ func TestOptimisticPrefetchRestoreIsMeasured(t *testing.T) {
 	require.Less(t, chunker.prePrefetchChunkSize, uint64(MaxDynamicRowSize))
 }
 
+// TestOptimisticPrefetchRestoreIgnoresSparseChunks is the same memory bound as
+// TestOptimisticPrefetchRestoreIsMeasured over the gap that is sparse rather
+// than empty — the case a `rows > 0` test for "was this size measured?" lets
+// through.
+//
+// A chunk holding a handful of rows across its whole width measures the cost of
+// those few rows, not of a full chunk of that size. And thinly-populated chunks
+// are precisely what carry the sizer to the ceiling, so accepting them records
+// the ceiling as measured from the very evidence that makes prefetch fire: one
+// row per chunk is enough to "prove" a 100k-row chunk no full chunk was ever
+// read at. The restore then hands the buffered copier the ~1.6GB read that
+// prePrefetchChunkSize exists to prevent.
+func TestOptimisticPrefetchRestoreIgnoresSparseChunks(t *testing.T) {
+	chunker := denseChunkerForTest(t)
+	chunker.TargetChunkBytes = DefaultTargetChunkBytes // buffered copier: byte signal
+
+	const wideRowBytes = 16 * 1024 // 16KB rows: ~1000 of them fill the budget
+
+	// Phase 1: full chunks of wide rows, as in the empty-gap case.
+	for range 60 {
+		chunk, err := chunker.Next()
+		require.NoError(t, err)
+		chunk.SourceRows = chunk.ChunkSize
+		chunk.ActualBytes = chunk.ChunkSize * wideRowBytes
+		chunker.Feedback(chunk, time.Millisecond, chunk.ChunkSize)
+	}
+	measured := chunker.lastDenseChunkSize
+	require.NotZero(t, measured)
+	require.Less(t, measured, uint64(MaxDynamicRowSize))
+
+	// Phase 2: a sparse gap. Every chunk carries a few real rows no matter how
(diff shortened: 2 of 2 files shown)


aparajon:
🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head `2db73bd3`.

**Verdict: the bug is real, the fullness property is the right one, and the reasoning about why a density gate is not sufficient checks out arithmetically. My finding is that the new test proves `rows > 0` is wrong without proving `2` is right** — I ran the divisor at 1, 3, 5, 7 and 8 and every one of them survives every test this machine can run, including the value the constant's own doc comment exists to argue against. Nothing blocks.

| # | Finding | Severity |
|---|---|---|
| 1 | `minChunkFullnessDivisor` is unpinned at every value from 1 to 8 — including 1, which the comment's whole rationale rejects | medium |
| 2 | The threshold is now quantitative on a count the code three lines above documents as an undercount | low |
| 3 | (pre-existing, adjacent) `keys == chunk.ChunkSize` is also unpinned, and the new term no longer helps it | low |

### Findings

**1. The new test cannot tell half-full from full.** Phase 2 carries 150 rows per chunk at widths from ~1024 up to 100,000, so *every* threshold from `rows*1` upward rejects those chunks — the test's discriminating power is entirely "reject the near-empty ones", which is what `rows > 0` failed to do. Mutating the constant to 1, 3, 5, 7 or 8 produces no new failure in `./pkg/table/`; only reverting the whole predicate to `rows > 0` fails, and it fails exactly the new test, so the PR's own claim about it is sound.

Divisor `1` is the one that matters, because it is the case the constant's doc comment is written to reject: *"Half rather than full because real tables have small gaps throughout, and requiring every key to hold a row would mean nothing is ever recorded."* That is a claim about real tables, it is the entire justification for the chosen value, and no test defends it — so the next person to read `rows*minChunkFullnessDivisor >= chunk.ChunkSize` and think "surely a measured size should be a *full* chunk" gets a green suite for tightening it, and every prefetch exit silently falls back to the `StartingChunkSize` ramp. `>=` versus `>` at exactly half is untested for the same reason.

A second phase-2 shape brackets it in a few lines: chunks that come back around 60% full and *must* still be recorded, alongside the existing 150-row shape that must not be. That converts the comment's reasoning into the thing that enforces it.

**2. `rows` is a count the same function documents as an undercount, and the test on it is no longer binary.** Three lines above the change: `actualRows` is the applier's affected-row count from `INSERT IGNORE`, which "does not count a row the binlog applier already wrote to the new table — so in an insert-hot region a dense chunk can report zero rows and be mistaken for a gap." `rows > 0` tolerated any nonzero undercount; `rows*2 >= chunk.ChunkSize` fails whenever the undercount exceeds 2x, so a genuinely full chunk in an insert-hot region can now go unmeasured.

I chased this and it is bounded: `SourceRows` is assigned in exactly two places, both in `pkg/copier/buffered.go`, both `uint64(len(rows))` — the buffered copier, which is the same path `TargetChunkBytes` is set for (`runner.go:1859`, copy chunker only). So on the path this change is about, `rows` is the true count and the concern dissolves. It does not dissolve on a copy path that leaves `SourceRows` zero, where the fallback's imprecision is now proportionally significant rather than merely non-zero-vs-zero. The failure direction is safe — an unmeasured size means a smaller restore, not a bigger one — so this is throughput, not memory. Worth a clause in the comment noting the fullness test presumes the producer's own count, so a future caller does not extend the predicate to a path where `rows` is an estimate.

**3. (pre-existing) Removing `keys == chunk.ChunkSize` entirely also survives.** Its comment says it "excludes a prefetch chunk, whose ChunkSize is a row offset while its width is the gap it crossed — restoring a prefetch offset would defeat the whole point," and nothing enforces it. It is not this PR's term and I would not gate on it — but it is worth noting that the new term does not cover for it: a prefetch chunk's `rows` is about its own row offset, so `rows*2 >= ChunkSize` passes comfortably and the fullness test would let exactly the chunk that comment describes through. If the phase-2 shape from Finding 1 is being written anyway, asserting `lastDenseChunkSize` is untouched across a prefetch chunk is a natural third case.

### Action items

1. **(Finding 1)** Add a phase-2 shape whose chunks come back ~60% full and assert `lastDenseChunkSize` *does* advance, so the choice of half over full is enforced rather than only explained.
2. **(Finding 2)** Note in the comment that the fullness test presumes `SourceRows` — the producer's own count — and that `actualRows` is an undercount the threshold is now proportionally sensitive to.
3. *(optional)* **(Finding 3)** While adding the case above, assert a prefetch chunk does not update `lastDenseChunkSize`, so the `keys == ChunkSize` term is defended too.

<details>

The diagnosis holds end to end. `minKeysPerRowForPrefetch` is `MaxDynamicRowSize / StartingChunkSize` = 100, so the summary's argument against a density-based gate is arithmetically right: 11664/150 = 77.8 sits under 100 and would pass, leaving an 11,664-row restore, which is why fullness rather than density is the property that matters here. The magnitudes line up too — `StartingChunkSize` 1000 against `DefaultTargetChunkBytes` 16MiB is ~1000 rows of 16KB, `MaxDynamicRowSize` 100,000 of them is ~1.6GB, and `readChunkData` appends every row into `rowDataList` with no in-read byte cap, so the restore really is fully materialised before any `ActualBytes` feedback exists to shrink it.

The blast radius is correctly scoped: `TargetChunkBytes` is set only on `copyChunkerCfg` (`runner.go:1859`), so the checksum chunker keeps its time signal and its server-side CRC and no rows reach the client — the summary's claim that it is unaffected is accurate.

The new test is careful in the ways that matter. It asserts its own premise (`150 × 16KB × 5 < 16MiB`, the prefetch entry gate) rather than assuming it, it asserts prefetch still engages so the change cannot pass by accidentally disabling the feature it is bounding, and it bounds the outcome in bytes (`prePrefetchChunkSize × wideRowBytes <= DefaultTargetChunkBytes`) rather than only in rows, which is the property that actually protects the copier. The 400-iteration loop with a `chunkPrefetchingEnabled` break is a bounded wait rather than a fixed count.

No overflow concern: `rows` is a `uint64` bounded by the chunk width, itself bounded by `MaxDynamicRowSize` (100,000), so `rows*2` is nowhere near the type's range. The assignment to `lastDenseChunkSize` remains unconditional rather than a running maximum, which is right — the most recent honest measurement is the best estimate, and a later smaller full chunk should supersede a larger older one.

Placing the fullness test before the `keys == ChunkSize` term reads oddly at first (the cheaper equality is second) but is deliberate given the comment order, and both are trivial.

Method note: 48 of `./pkg/table/`'s tests need a MySQL this machine does not have — the local server on 3306 is not the Spirit test instance — so the mutation results above are against the DB-free subset. Two of the DB-backed ones are prefetch-related, so I cannot rule out one of them catching a divisor change; the point of Finding 1 stands regardless, since the test added for this constant is the one that should pin it and does not. `TestOptimisticPrefetchRestoreIsMeasured`, `TestOptimisticPrefetchRestoreIgnoresSparseChunks` and `TestOptimisticPrefetchDensityUsesSourceRows` all pass here at this head, and CI is green (15/15).

</details>

aparajon (approved):

PR #1196, opened 2026-09-04
test(change): make BlockWait stall assertions deterministic

### LINKED ISSUE #1162: flaky test: TestBlockWait — 300ms reader stall trips blockWaitStallThreshold (2x this week, both MySQL 9.7)
🤖 _Filed by Morgan's AI agent from a scheduled review of `main` CI failures._

## Symptom

`TestBlockWait` asserts that `BlockWait` performs **zero** `FLUSH BINARY LOGS` while the buffered binlog position is advancing. It observed one.

```
--- FAIL: TestBlockWait (6.27s)
    binlog_test.go:608:
        Error Trace:	/app/pkg/change/binlog_test.go:608
        Error:      	Not equal:
                    	expected: 0
                    	actual  : 1
        Test:       	TestBlockWait
```

Two occurrences this week, **both on MySQL 9.7**:

| When | Branch | Job | Run |
|---|---|---|---|
| 2026-08-17 12:50 | `main` (`fbd4cf84`, #1147) | MySQL 9.7 | https://github.com/block/spirit/actions/runs/32031637813 |
| 2026-08-17 03:45 | `remove-target-chunk-time-flag` | MySQL 9.7 | https://github.com/block/spirit/actions/runs/31991990967 |

Same test, same assertion line, same `expected 0 / actual 1`.

## Likely cause: the stall window is 300 ms

`BlockWait` polls every `blockWaitSleep = 100ms` and flushes once it sees `blockWaitStallThreshold = 3` **consecutive** polls where the buffered position has not advanced:

https://github.com/block/spirit/blob/main/pkg/change/binlog.go#L1360-L1377

```go
currPos := c.getBufferedPos()
if currPos.Compare(prevPos) <= 0 && !first {
    stallCount++
    if stallCount >= blockWaitStallThreshold {   // 3
        dbconn.Exec(ctx, c.db, "FLUSH BINARY LOGS")
        c.flushedBinlogs.Add(1)
        stallCount = 0
    }
} else {
    stallCount = 0
}
```

(`blockWaitSleep` / `blockWaitStallThreshold` are defined in `pkg/change/repl.go:88` and `:93`.)

So **300 ms of binlog-reader non-progress is enough to trip the flush**. The test spins a goroutine writing continuously to an unrelated table (`blockwaitt3`) and expects that traffic to keep the position moving for the whole of `BlockWait` — but the assertion is on the *reader's* buffered position, not on the writes. Under parallel-test CI load, a 300 ms hiccup in the reader is easy to hit.

The MySQL 9.7 correlation (2/2) may be a coincidence at this sample size, or may reflect a slower/noisier 9.7 container in the matrix — worth a look.

## Prior art

- #743 — _flaky test: TestBlockWait_ — closed **2026-06-30**, i.e. outside the "recently closed" window.
- #745 — _repl: fix flaky TestBlockWait by requiring consecutive stalls before flush_ — this is the PR that introduced `blockWaitStallThreshold`. The threshold it chose (3 × 100 ms) is evidently still too tight.

## Possible directions

- Raise `blockWaitStallThreshold` (and/or `blockWaitSleep`) so a brief reader hiccup under load does not trigger a flush. This changes production behaviour, so it needs a view on how quickly a genuinely stalled stream should be nudged.
- Make the test assertion tolerant: assert "not repeatedly flushing" rather than "exactly zero flushes", since one flush from a transient stall is arguably correct behaviour.
- Base the stall decision on elapsed wall time since last advance rather than a poll count, which decouples it from CI scheduling jitter.

morgo:
TestBlockWait requires zero binlog rotations while an unrelated writer is active, but a legitimately stalled reader can trigger a rotation under CI scheduling pressure. Extract the existing consecutive-stall policy and test it with controlled MySQL positions, including offset regression and file rollover: advancing readers never rotate, progress resets the stall count, and sustained stalls rotate at the existing threshold.

Keep real MySQL catch-up coverage for unrelated-table events and remove the scheduling-dependent rotation assertion, background writer, and three-second sleep. Production stall thresholds and behavior are unchanged.

Fixes #1162.

Validation: `go test -race ./pkg/change -run '^TestBlockWait' -count=10`; `golangci-lint run ./pkg/change/...`.

### DIFF
--- a/pkg/change/binlog.go
+++ b/pkg/change/binlog.go
@@ -1435,8 +1435,7 @@ func (c *binlogClient) BlockWait(ctx context.Context) error {
 	defer timer.Stop() // Ensure timer is always stopped to prevent goroutine leak
 
 	prevPos := c.getBufferedPos()
-	first := true
-	stallCount := 0
+	stalls := blockWaitStalls{}
 	for {
 		select {
 		case <-ctx.Done():
@@ -1445,25 +1444,14 @@ func (c *binlogClient) BlockWait(ctx context.Context) error {
 			return fmt.Errorf("timed out waiting to catch up to source position: %v, current position is: %v", targetPos, c.getBufferedPos())
 		default:
 			currPos := c.getBufferedPos()
-			if currPos.Compare(prevPos) <= 0 && !first {
-				// Position hasn't advanced. Only flush after multiple consecutive
-				// stalls to avoid unnecessary flushes when the binlog syncer is
-				// just slightly behind (e.g., under CI load). getCurrentBinlogPosition
-				// already flushes once at the start, so a brief stall is expected.
-				stallCount++
-				if stallCount >= blockWaitStallThreshold {
-					c.logger.Debug("buffered position has not advanced, flushing binary logs")
-					if err := dbconn.Exec(ctx, c.db, "FLUSH BINARY LOGS"); err != nil {
-						return err // it could be context cancelled, return it
-					}
-					c.flushedBinlogs.Add(1)
-					stallCount = 0
+			if stalls.observe(currPos.Compare(prevPos) > 0) {
+				c.logger.Debug("buffered position has not advanced, flushing binary logs")
+				if err := dbconn.Exec(ctx, c.db, "FLUSH BINARY LOGS"); err != nil {
+					return err
 				}
-			} else {
-				stallCount = 0
+				c.flushedBinlogs.Add(1)
 			}
 			prevPos = currPos
-			first = false
 
 			if c.getBufferedPos().Compare(targetPos) >= 0 {
 				return nil // we are up to date!
--- a/pkg/change/binlog.go
+++ b/pkg/change/binlog.go
@@ -152,7 +152,7 @@ type binlogClient struct {
 	// parks), so requests never queue behind each other.
 	flushRequests chan Subscription
 
-	flushedBinlogs atomic.Int64 // for testing binlog flushing frequency
+	flushedBinlogs atomic.Int64 // stall-triggered rotations reported as FeedStats.ForcedRotations
 }
 
 // NewBinlogClient constructs the binlog-backed change.Source. The
@@ -1444,7 +1444,7 @@ func (c *binlogClient) BlockWait(ctx context.Context) error {
 			return fmt.Errorf("timed out waiting to catch up to source position: %v, current position is: %v", targetPos, c.getBufferedPos())
 		default:
 			currPos := c.getBufferedPos()
-			if stalls.observe(currPos.Compare(prevPos) > 0) {
+			if stalls.observe(prevPos, currPos) {
 				c.logger.Debug("buffered position has not advanced, flushing binary logs")
 				if err := dbconn.Exec(ctx, c.db, "FLUSH BINARY LOGS"); err != nil {
 					return err
--- a/pkg/change/binlog_test.go
+++ b/pkg/change/binlog_test.go
@@ -583,29 +583,11 @@ func TestBlockWait(t *testing.T) {
 	require.NoError(t, client.Start(t.Context()))
 	defer client.Close()
 
-	// We test that BlockWait does not flush the binlog if the buffered position is advancing by
-	// 1. kicking off a go-routine that inserts into an unrelated table
-	// 2. verifying that flushedBinlogs is still 0 at the end of BlockWait
-	ctx, cancel := context.WithCancel(t.Context())
-	var wg sync.WaitGroup
-	wg.Go(func() {
-		i := 1
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			default:
-				_, _ = db.ExecContext(ctx, fmt.Sprintf("INSERT INTO blockwaitt3 (a, b, c) VALUES (%d, %d, %d)", i, i, i))
-				i++
-			}
-		}
-	})
-	time.Sleep(3 * time.Second) // should be enough for BlockWait to block for 1 iteration before catching up, but not guaranteed
-	client.flushedBinlogs.Store(0)
+	// Unrelated-table events must not prevent the reader catching up. Whether
+	// a real reader briefly stalls is scheduler-dependent; the exact rotation
+	// policy is covered by TestBlockWaitStalls using controlled progress.
+	testutils.RunSQL(t, "INSERT INTO blockwaitt3 (a, b, c) VALUES (1, 1, 1)")
 	require.NoError(t, client.BlockWait(t.Context()))
-	cancel()
-	wg.Wait() // ensure goroutine exits before test completes
-	require.Equal(t, int64(0), client.flushedBinlogs.Load())
 
 	// Insert into t1.
 	testutils.RunSQL(t, "INSERT INTO blockwaitt1 (a, b, c) VALUES (1, 2, 3)")
--- a/pkg/change/blockwait.go
+++ b/pkg/change/blockwait.go
@@ -1,0 +1,24 @@
+package change
+
+// blockWaitStalls decides when the file-position reader needs a rotation to
+// nudge it forward. The first observation establishes a baseline; progress
+// resets the consecutive-stall count. Keep this independent of wall time so
+// tests can exercise the policy without assuming how CI schedules the reader.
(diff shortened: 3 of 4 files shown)


aparajon:
🤖 **Adversarial correctness review** — `1eb74f4d` (+70/-40, 4 files)

The diagnosis is right and the replacement `TestBlockWait` is *better*, not merely less flaky: `getCurrentBinlogPosition` rotates before sampling, so the target lands at position 4 of a fresh file and the reader cannot reach it without consuming the `blockwaitt3` insert from the previous one. A single deterministic INSERT therefore covers the unrelated-table catch-up path more strictly than a racing writer plus a 3-second sleep did. `blockWaitStalls` is also correctly scoped — `gtidClient.BlockWait` has no stall policy at all, so "the file-position reader" in the docstring is precise rather than approximate.

I checked the "production behavior unchanged" claim rather than trusting it: exhaustively over every advance/stall sequence up to length 15, plus 200k random sequences up to length 200, the old inline `first`/`stallCount` state machine and `observe` emit **identical** rotation sequences. It is an exact refactor.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | med | `binlog.go:1447` | The extraction stops one line short of the only line that can regress, and the deleted assertion was the only thing pinning it — inverting the comparison keeps the whole package green |
| 2 | low-med | `blockwait.go:1-6`, `binlog.go:1450` | The half of the old comment that explains *why* a stall is normal is the half that was dropped, and it is the mechanism behind the flake being fixed |
| 3 | low | `binlog.go:155` | `flushedBinlogs`'s comment is now wrong in both directions |

---

### 1 — the sign of the comparison is now untested in both directions (med)

```go
			if stalls.observe(currPos.Compare(prevPos) > 0) {
```

`observe` is thoroughly pinned; the argument to it is not. Measured on `./pkg/change`:

| Mutation | Result |
|---|---|
| `currPos.Compare(prevPos) > 0` → `<= 0` (invert advanced/stalled) | **passes** |
| argument hardcoded to `true` (rotation can never fire) | **passes** |
| `s.consecutive < threshold` → `<=` | **FAIL** — "sustained stalls" |
| drop `s.consecutive = 0` before `return true` | **FAIL** — "sustained stalls" |
| drop the `!s.observed` baseline | **FAIL** — "sustained stalls" |
| progress no longer resets the count | **FAIL** — "interrupted stalls" |

The inversion is the consequential one. With the sign flipped, a *healthy advancing* reader is scored as stalled and issues `FLUSH BINARY LOGS` every third 100ms poll — and `Flush()` calls `BlockWait` in a loop, so on a busy source that is a sustained rotation storm against the production server, from a code path whose whole purpose is to avoid unnecessary rotations. Nothing in the suite notices, because `require.Equal(t, int64(0), client.flushedBinlogs.Load())` was the **only** assertion on `flushedBinlogs` anywhere — after this PR `grep flushedBinlogs *_test.go` returns nothing. The only surviving signal is `stats.ForcedRotations` (`binlog.go:1262`) in the operator's status block, i.e. a human noticing.

The second mutation is the mirror image: the nudge could stop firing entirely and nothing fails either. So removing the scheduling-dependent assertion is a straight trade of a flaky test for no test, when it doesn't have to be. Pushing the comparison inside the seam closes both:

```go
func (s *blockWaitStalls) observe(prev, curr mysql.Position) bool {
	...
	if !s.observed || curr.Compare(prev) > 0 {
```

Then `TestBlockWaitStalls` drives `mysql.Position{Name: "binlog.000001", Pos: 4}` vs `{Pos: 8}` instead of a bare bool, both mutations above become compile-time-impossible or test failures, and the file/position rollover case (`Compare` orders by name first) gets covered for free — which is the one comparison in here a reader might plausibly get wrong.

### 2 — the dropped sentence is the one that explains the flake (low-med)

The threshold's numeric rationale survives, on the constant:

```go
	// 3 * blockWaitSleep (~300ms) tolerates brief syncer lag (e.g. CI load) while
	// remaining negligible relative to DefaultTimeout.
	blockWaitStallThreshold = 3
```

What didn't survive is *"getCurrentBinlogPosition already flushes once at the start, so a brief stall is expected."* That is a true and non-obvious fact — `getCurrentBinlogPosition` issues an unconditional `FLUSH BINARY LOGS` (`binlog.go:348`) on **every** `BlockWait`, so the reader is handed a brand-new file to reopen at the top of every call. It is simultaneously (a) why a stall at the start is normal rather than alarming, (b) why `flushedBinlogs` undercounts real rotations, since it counts only the stall-triggered ones, and (c) the mechanism behind the very flake this PR fixes. The new docstring justifies the type by testability — *"Keep this independent of wall time so tests can exercise the policy"* — which is about the test, not about the reader. One line carrying the old sentence into `blockwait.go` keeps the production reasoning with the production code.

Same shape, smaller: the bare `return err` lost its `// it could be context cancelled, return it`, which is exactly the annotation that stops a future reader from "fixing" it into a `fmt.Errorf` wrap that breaks `errors.Is(err, context.Canceled)` for the caller.

### 3 — `flushedBinlogs` is mislabelled twice over (low)

```go
	flushedBinlogs atomic.Int64 // for testing binlog flushing frequency
```

No test reads it any more, and it is published to operators as `stats.ForcedRotations`. Both halves of the comment are now backwards: it is not for testing, and it is not private telemetry. Worth retitling to say what it is — the count of stall-triggered rotations surfaced in the status block — since that framing is also the argument for keeping a test on it (finding 1).

---

<details>
<summary><b>Verified</b> — exact equivalence, and why the new integration test is stronger</summary>

**The refactor is behaviour-preserving, checked not assumed.** Simulating both state machines side by side — old `first bool` + `stallCount int`, new `observed bool` + `consecutive int` — over all 2^n advance/stall sequences for n ≤ 15 (65,534 sequences) and 200,000 random sequences of length ≤ 200 produced **identical** rotation outputs. The one case worth calling out explicitly: a first observation that is *also* a stall takes the baseline branch in both versions (old via `!first`, new via `!s.observed`), so no rotation can ever fire before the fourth poll — which is what makes the removed `require.Equal(0, ...)` deterministic for *short* BlockWaits and flaky only for long ones.

**The new `TestBlockWait` still covers the path it claims.** `getCurrentBinlogPosition` rotates first, so `targetPos` is `(binlog.00000N+1, 4)` while the `blockwaitt3` insert sits in file N; `mysql.Position.Compare` orders by file name before offset, so the reader must actually consume the unrelated-table event to satisfy the wait. The old version's `flushedBinlogs.Store(0)` reset correctly disappears with the assertion it existed for, and dropping the writer goroutine also drops a `db.ExecContext` whose errors were discarded in a loop.

**`observe` is a good seam for the part it does cover.** Table-driven over three regimes with the threshold referenced symbolically rather than hardcoded, so bumping `blockWaitStallThreshold` cannot silently invalidate the test; the "interrupted stalls" subtest runs ten reset cycles rather than one, so an off-by-one in the reset would surface; `-count=10` under `-race` is stable.

**Scope is right.** `blockWaitStallThreshold` has exactly one production reference (`blockwait.go:19`) after the move, `gtidClient.BlockWait` genuinely has no rotation logic to unify with, and no other caller reads `blockWaitStalls`.

**Local:** `go build ./...` clean, `gofmt` clean, `golangci-lint run ./pkg/change/...` → **0 issues**, `go test -race ./pkg/change -run '^TestBlockWaitStalls$' -count=10` passes. CI 18/18 green.

</details>

aparajon (approved):
🤖 Approving — verified the refactor is exactly behaviour-preserving (exhaustive + randomized simulation of both state machines), and the new unit test pins the policy well. Main finding is that the sign of the advance comparison is now untested in both directions; details in the review comment above.

*This approval was generated by Claude Code (claude-opus-5).*

morgo:
🤖 Moved the position comparison into the stall-policy seam and added offset advance/regression, unchanged-position, and file-rollover/regression cases. Restored the explanation that the initial rotation can cause a normal reader stall, and corrected the forced-rotation counter comment.

Validation: all BlockWait tests passed 10 race-enabled repetitions; lint reports 0 issues. Inverting the position comparison via a Go overlay fails the updated tests. The cancellation error still propagates unchanged; wrapping with %w would also preserve errors.Is, so no prohibition on error wrapping was added.

Implemented in 30be008b.

PR #1197, opened 2026-09-04
fix(change): join periodic flushing during Close

morgo:
Both change-feed implementations leave periodic flushing alive when Close returns, despite the Source contract requiring all resources to be released. Cancel and join periodic flushing during Close, after waking parked subscriptions, and prevent a closed client from restarting the periodic loop. Document this lifecycle contract on Source and both implementations, and log ignored starts on closed clients.

The regression test holds an in-flight flush after it observes cancellation, proving Close waits for completion rather than merely sending cancellation. It covers both file-position and GTID clients, repeated Close, and attempts to restart flushing after Close. This fixes the shared implementation used by migration, move, and datasync; no runner-specific lifecycle fork is needed.

Validation:

- Race-enabled shutdown and parked-reader tests, five repetitions.
- `go test -race ./pkg/change ./pkg/migration ./pkg/move ./pkg/datasync -parallel=4 -timeout=5m` against MySQL 8.0.43.
- `golangci-lint run ./pkg/change/...`.

### DIFF
--- a/pkg/change/binlog.go
+++ b/pkg/change/binlog.go
@@ -1115,6 +1115,10 @@ func (c *binlogClient) Close() {
 
 	// Wait for the readStream goroutine to exit cleanly. This prevents
 	// goroutine leaks detected by goleak in tests.
+	// Quiesce writes as well as the reader before callers tear down pools.
+	// The periodic loop has its own context and is not joined by streamWG.
+	c.StopPeriodicFlush()
+
 	c.streamWG.Wait()
 
 	// streamWG.Wait has returned, so readStream has exited and c.syncer
@@ -1347,7 +1351,7 @@ func (c *binlogClient) StopPeriodicFlush() {
 // Satisfies Source interface.
 func (c *binlogClient) StartPeriodicFlush(ctx context.Context, interval time.Duration) {
 	c.periodicFlushLock.Lock()
-	if c.periodicFlushCancel != nil {
+	if c.isClosed.Load() || c.periodicFlushCancel != nil {
 		c.periodicFlushLock.Unlock()
 		return
 	}
--- a/pkg/change/binlog.go
+++ b/pkg/change/binlog.go
@@ -1115,8 +1115,8 @@ func (c *binlogClient) Close() {
 
 	// Wait for the readStream goroutine to exit cleanly. This prevents
 	// goroutine leaks detected by goleak in tests.
-	// Quiesce writes as well as the reader before callers tear down pools.
-	// The periodic loop has its own context and is not joined by streamWG.
+	// Join the independently cancellable writer too. Both background loops
+	// must finish before Close returns; neither join depends on the other.
 	c.StopPeriodicFlush()
 
 	c.streamWG.Wait()
@@ -1347,11 +1347,16 @@ func (c *binlogClient) StopPeriodicFlush() {
 // StopPeriodicFlush is guaranteed to observe the registration. Callers
 // MUST NOT prefix with `go` — the loop is spawned internally.
 //
-// Calling Start while a flush is already running is a no-op.
+// Calling Start while a flush is already running or after Close is a no-op.
 // Satisfies Source interface.
 func (c *binlogClient) StartPeriodicFlush(ctx context.Context, interval time.Duration) {
 	c.periodicFlushLock.Lock()
-	if c.isClosed.Load() || c.periodicFlushCancel != nil {
+	if c.isClosed.Load() {
+		c.periodicFlushLock.Unlock()
+		c.logger.Debug("ignoring periodic flush start on a closed client")
+		return
+	}
+	if c.periodicFlushCancel != nil {
 		c.periodicFlushLock.Unlock()
 		return
 	}
--- a/pkg/change/close_test.go
+++ b/pkg/change/close_test.go
@@ -1,0 +1,94 @@
+package change
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/block/spirit/pkg/dbconn"
+	"github.com/stretchr/testify/require"
+)
+
+// Hold the flush after it observes cancellation so the test distinguishes
+// cancelling the worker from actually joining it.
+type closingSubscription struct {
+	stubSubscription
+	entered   chan struct{}
+	cancelled chan struct{}
+	release   chan struct{}
+}
+
+func (s *closingSubscription) Flush(ctx context.Context, _ bool, _ []*dbconn.TableLock) (bool, error) {
+	close(s.entered)
+	<-ctx.Done()
+	close(s.cancelled)
+	<-s.release
+	return false, ctx.Err()
+}
+
+func TestCloseJoinsPeriodicFlush(t *testing.T) {
+	for _, name := range []string{"binlog", "gtid"} {
+		t.Run(name, func(t *testing.T) {
+			var client Source
+			var requests chan Subscription
+			var flushDone func() chan struct{}
+			if name == "binlog" {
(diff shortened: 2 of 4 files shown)


aparajon:
🤖 **Adversarial correctness review** — `4b4608e8` (+104/-2, 3 files)

Real leak, right fix, and `TestCloseJoinsPeriodicFlush` is the best-constructed test I've reviewed in this repo in a while — holding the flush *after* it observes cancellation is exactly what separates "sent a cancel" from "joined the worker," and the assertion that `closed` has **not** fired at that moment is what makes it a proof rather than a smoke test. It caught every mutation I aimed at the core.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | med | `source.go:151-155` + both `StartPeriodicFlush` docstrings | A new implementation requirement lands on an out-of-tree extension point without landing on its interface, and two unrelated causes share one silent return |
| 2 | low | `binlog.go:1118` / `gtid.go:1109` | The placement both comments argue for is unpinned, and I could not find the constraint they claim |

---

### 1 — the interface didn't learn what its implementations now do (med)

`Source` is documented as an extension point — *"Out-of-tree implementations construct their own Source value and pass it to spirit via the Move/Migration config."* Its contract for this method says exactly one thing:

```go
	// StartPeriodicFlush spawns a background goroutine that flushes the
	// changeset at the given interval. Used by the migrator to advance
	// the safe-flushed position. Calling Start while a periodic flush
	// is already running is a no-op.
```

This PR adds a **second** no-op case to both in-tree clients — a closed client must refuse to start — and asserts it in the regression test (`require.Nil(t, flushDone(), "a closed client must not restart flushing")`). Neither `source.go` nor either implementation docstring mentions it. So the behavior the test now pins for `binlogClient` and `gtidClient` is invisible to anyone writing the third implementation, and "Close releases all resources" doesn't imply it on its own — a Source could satisfy that literally and still hand out a fresh unjoined goroutine afterwards.

The second half is the `||`:

```go
	if c.isClosed.Load() || c.periodicFlushCancel != nil {
		c.periodicFlushLock.Unlock()
		return
	}
```

These are not the same event. `periodicFlushCancel != nil` is benign and expected — the documented idempotent call. `isClosed` is a **caller lifecycle bug**: something is trying to restart flushing on a dead client, and it will silently advance no binlog position from then on, which is the kind of thing that shows up much later as a checkpoint that stopped moving. Worth splitting, with a `Debug`/`Warn` on the closed branch so it is visible at all — the repo's own guidance is to keep causes with different meanings in separate branches rather than deduping them behind one condition.

### 2 — the placement is justified twice and pinned zero times (low)

Both files gained the same two-line rationale:

```go
	// Quiesce writes as well as the reader before callers tear down pools.
	// The periodic loop has its own context and is not joined by streamWG.
	c.StopPeriodicFlush()

	c.streamWG.Wait()
```

Moving the call to *after* `streamWG.Wait()` in both files **passes** the new test (and the rest of `./pkg/change`). I went looking for the case that would break — the periodic path is `runPeriodicFlush → c.flush → subscription.Flush`, which takes `c.mu` twice briefly and never waits on the reader, and `requestFlush` is a non-blocking send, so a dead consumer cannot strand `readStream` either. Both orders also keep the stop ahead of `c.syncer.Close()`. So as far as I can establish the two placements are equivalent, which makes the comment's *"before callers tear down pools"* read as a constraint when it is really a preference — and it is now duplicated in two files that can drift apart independently. Either a test that distinguishes them, or softening the comment to say why this order is *preferred* rather than required, would leave the next reader better off.

---

<details>
<summary><b>Verified</b> — mutations, the shutdown TOCTOU, and two attacks that dissolved</summary>

**Every mutation at the core is caught:**

| Mutation | Result |
|---|---|
| drop `StopPeriodicFlush()` from `binlogClient.Close` only | **FAIL** — `binlog` subtest: "Close did not cancel the in-flight flush" |
| `StopPeriodicFlush` cancels but does not `<-done` (both clients) | **FAIL** — both subtests: "Close returned before the in-flight flush finished" |
| drop the `isClosed.Load()` guard (both clients) | **FAIL** — both subtests at the restart assertion |
| move `StopPeriodicFlush()` after `streamWG.Wait()` (both) | **passes** → finding 2 |

Running the table subtest over both clients means a one-sided fix can't slip through — M1 shows the `gtid` half staying green while `binlog` fails.

**The shutdown TOCTOU is airtight.** `isClosed.Store(true)` is the first statement of `Close`, while the guard is read under `periodicFlushLock`, so both interleavings are safe: if `StartPeriodicFlush` wins the lock before the store, it registers cancel/done and `Close`'s later `StopPeriodicFlush` joins it; if the store lands first, every subsequent `Start` is refused, including after `StopPeriodicFlush` has already nil'd the pair. The narrow window where `Start` has released the lock but not yet reached `go runPeriodicFlush` is also safe — registration is synchronous (the docstring says so deliberately), and `<-done` unblocks once the goroutine runs its `defer close(done)`. `StopPeriodicFlush` releases the lock *before* `cancel(); <-done`, so nothing can deadlock against a flush that wants the lock.

⭐ **Attack that dissolved: stopping the flusher before `streamWG.Wait()` looked like a deadlock.** If `readStream` could block sending on `flushRequests` (cap 1) with its only consumer dead, `Wait()` would hang forever. It can't — `bufferedMap.requestFlush` is a `select` with `default`, documented as "the send must not block". Same reason a parked reader can't be stranded: `Close` wakes every subscription via `sub.Close()` *before* stopping the flusher, which is the ordering the body describes.

**Attack that dissolved: `Close` can now block for a whole in-flight flush.** That is within contract rather than a regression — `Source.Stop` explicitly documents *"Stop must not block, because every write to the table is stalled behind it"* while `Close` documents only *"releases all resources. Safe to call more than once."* The distinction is deliberate, and blocking is the correct trade for a teardown path that must not leak.

**Repeated `Close` is genuinely safe**, not just tested: `isClosed.Store` is not a `Swap` guard, so the body re-runs, but the second pass finds `periodicFlushCancel == nil` (no-op), re-calls an idempotent `cancel()`, returns immediately from an empty `streamWG`, and skips the nil'd `syncer`.

**Test hygiene:** `t.Context()` throughout, cleanup owns `sub.release` so the body's `sub.release <- struct{}{}` can't double-close, and every wait is a bounded `time.After(5s)` that hard-fails — no unbounded blocks. The `flushDone()` closure reads a `periodicFlushLock`-protected field without the lock, but every such read is ordered after `Close` returns or before `Close` is spawned, and `-race` over 5 runs is clean.

**Local:** `go build ./...` clean, `gofmt` clean, `golangci-lint run ./pkg/change/...` → **0 issues**, `go test -race ./pkg/change -run TestCloseJoinsPeriodicFlush -count=5` passes. CI 18/18 green.

</details>

aparajon (approved):
🤖 Approving — real leak, right fix, and the test genuinely proves Close *joins* rather than merely cancels (it caught every mutation I aimed at the core). Two non-blocking notes in the review comment above: the new "a closed client must refuse to start" requirement isn't in the `Source` interface doc that out-of-tree implementations read, and the `||` folds a benign idempotent call together with a caller-lifecycle bug that silently stops advancing the position.

morgo:
🤖 Documented Close's join/no-restart contract on Source and both StartPeriodicFlush implementations. Split the closed-client guard from the already-running guard and added a debug message for an ignored start after Close. Revised the shutdown comments to state the actual requirement: both independently cancellable loops must finish before Close returns, without claiming a dependency between their join order.

Validation: the held-in-flight-flush lifecycle test passed five race-enabled repetitions for both clients; lint reports 0 issues.

Implemented in f1223d6c.

PR #1198, opened 2026-09-04
fix(throttler): cancel and join background monitors on Close

morgo:
Replica and Aurora throttler Close methods only set a flag, leaving polling goroutines and in-flight queries alive after shutdown returns. Introduce a shared monitor lifecycle that cancels and joins the poller before callers close its database pool. Use it for replica lag, Aurora threads, and commit-latency monitoring. This makes Runner.Close wait for in-flight monitoring before closing its Aurora and replica pools. Open on a closed throttler now returns an error before sampling.

Tests hold a monitor open after cancellation to verify that each public Close waits for completion, and cover concurrent Close, Close-before-start, reopen rejection, and idempotent repeated start. Remove the replica test's shutdown sleep and global polling-interval mutation.

Validation: `go test -race ./pkg/throttler -count=5`; `golangci-lint run ./pkg/throttler/...`. Aurora sampling itself still requires Aurora; lifecycle tests run without it.

### DIFF
--- a/pkg/throttler/aurora_commitlatency.go
+++ b/pkg/throttler/aurora_commitlatency.go
@@ -111,14 +111,16 @@ func NewCommitLatencyThrottler(db *sql.DB, threshold time.Duration, logger *slog
 }
 
 func (c *CommitLatency) Open(ctx context.Context) error {
+	if err := c.poller.checkOpen(); err != nil {
+		return err
+	}
 	// Take an initial sample so the first delta computed by the background
 	// loop is meaningful; otherwise we'd flap "throttled" on the very first
 	// post-open chunk based on whatever the cumulative average happened to be.
 	if err := c.UpdateLag(ctx); err != nil {
 		return err
 	}
-	c.poller.start(ctx, c.run)
-	return nil
+	return c.poller.start(ctx, c.run)
 }
 
 func (c *CommitLatency) run(ctx context.Context) {
--- a/pkg/throttler/aurora_commitlatency.go
+++ b/pkg/throttler/aurora_commitlatency.go
@@ -72,7 +72,7 @@ type CommitLatency struct {
 	logger    *slog.Logger
 
 	isThrottled atomic.Bool
-	isClosed    atomic.Bool
+	poller      monitorLoop
 
 	// Previous sample, guarded by sampleMu. commits and latency must move
 	// together to compute a meaningful delta, so a single mutex is simpler
@@ -117,7 +117,7 @@ func (c *CommitLatency) Open(ctx context.Context) error {
 	if err := c.UpdateLag(ctx); err != nil {
 		return err
 	}
-	go c.run(ctx)
+	c.poller.start(ctx, c.run)
 	return nil
 }
 
@@ -129,9 +129,6 @@ func (c *CommitLatency) run(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			if c.isClosed.Load() {
-				return
-			}
 			if err := c.UpdateLag(ctx); err != nil {
 				if isShutdownError(ctx, err) {
 					return // teardown cancelled the in-flight sample; not a monitoring failure
@@ -144,7 +141,7 @@ func (c *CommitLatency) run(ctx context.Context) {
 }
 
 func (c *CommitLatency) Close() error {
-	c.isClosed.Store(true)
+	c.poller.close()
 	return nil
 }
 
--- a/pkg/throttler/aurora_threads.go
+++ b/pkg/throttler/aurora_threads.go
@@ -294,6 +294,9 @@ func newAuroraThreadsThrottler(db *sql.DB, mode threadsMode, logger *slog.Logger
 }
 
 func (a *AuroraThreads) Open(ctx context.Context) error {
+	if err := a.poller.checkOpen(); err != nil {
+		return err
+	}
 	vCPUs, err := AuroraVCPUs(ctx, a.db)
 	if err != nil {
 		return err
@@ -304,8 +307,7 @@ func (a *AuroraThreads) Open(ctx context.Context) error {
 	if err := a.UpdateLag(ctx); err != nil {
 		return err
 	}
-	a.poller.start(ctx, a.run)
-	return nil
+	return a.poller.start(ctx, a.run)
 }
 
 func (a *AuroraThreads) run(ctx context.Context) {
--- a/pkg/throttler/aurora_threads.go
+++ b/pkg/throttler/aurora_threads.go
@@ -256,7 +256,7 @@ type AuroraThreads struct {
 	vCPUs int64
 
 	isThrottled atomic.Bool
-	isClosed    atomic.Bool
+	poller      monitorLoop
 
 	// lastSample holds the most recent observation for logging and the raw
 	// hard-stop comparison.
@@ -304,7 +304,7 @@ func (a *AuroraThreads) Open(ctx context.Context) error {
 	if err := a.UpdateLag(ctx); err != nil {
 		return err
 	}
-	go a.run(ctx)
+	a.poller.start(ctx, a.run)
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


aparajon:
🤖 **Adversarial correctness review** — `a618bd31` (+152/-35, 6 files)

`monitorLoop` is the right shape and the start/close interleaving is airtight — `start` holds the mutex across the `go` statement, so registration and spawn can never be split by a concurrent `close`. Dropping the `isClosed.Load()` check inside each ticker case is also a real improvement, not just a refactor: the old flag was only observed at the *next* tick, so shutdown latency was up to a full `loopInterval` (5s).

**The production payoff is bigger than the body says.** `runner.go:1640` already carries this comment:

> Close the Aurora monitor pool after the throttler so its background pollers observe `Close()` / ctx cancellation before we yank the pool out from under them.

Before this PR that was aspirational — `Close()` set a flag and the poller noticed up to 5s later, so `monitorDB.Close()` genuinely could pull the pool out from under an in-flight sample (same for `closeReplicas()` and the replica throttler's pools). This PR is what makes that comment true. Worth saying in the body; it's a stronger motivation than "leaves polling goroutines alive."

| # | Sev | Where | What |
|---|---|---|---|
| 1 | med | `monitor.go:20` + the three `Open` methods | `Open` after `Close` returns `nil` having started nothing, and an Aurora throttler then never throttles again |
| 2 | med-low | `monitor.go:20` | The double-start guard is unpinned — removing it keeps the suite green, and it reintroduces exactly this PR's leak |
| 3 | low | `multi.go` `multiThrottler.Close` | Blocking child Closes are serialized, so one slow child defers every later child's cancel past its own pool teardown |

---

### 1 — a closed monitor makes `Open` a silent no-op (med)

```go
func (m *monitorLoop) start(ctx context.Context, run func(context.Context)) {
	...
	if m.closed || m.done != nil {
		return
	}
```

Every caller is shaped like this, and none of them can tell:

```go
func (a *AuroraThreads) Open(ctx context.Context) error {
	if err := a.UpdateLag(ctx); err != nil {
		return err
	}
	a.poller.start(ctx, a.run)
	return nil
}
```

So `Open` on a closed throttler **returns success with no poller running**, and the throttler's verdict freezes at the single `UpdateLag` sample `Open` took. For `AuroraThreads` that is not a fail-closed freeze: `IsThrottled()` is a bare `a.isThrottled.Load()` with no staleness guard, deliberately — `aurora_commitlatency.go:156` explains that the hard stop is *"itself unaffected by staleness"*, which is true while sampling is running and false when nothing is sampling. A healthy sample at `Open` time therefore means the throttler never throttles again. `Replica` degrades more gracefully (its `stale` guard fails closed), which makes the inconsistency worse rather than better.

The two silent returns also mean different things: `m.done != nil` is the benign idempotent call, `m.closed` is a caller lifecycle bug. Cheapest fix is to give `start` an error (or a bool) so `Open` can refuse a reopened throttler outright; a `Warn` on the closed branch would at least make it visible. This is the same decision as the `isClosed ||` guard in #1197 — worth settling once for both.

### 2 — the double-start guard has no test (med-low)

Measured on `./pkg/throttler`:

| Mutation | Result |
|---|---|
| drop `m.closed` from the guard | **FAIL** (`TestMonitorCloseBeforeStart`) |
| drop `m.done != nil` from the guard | **passes** |
| `close()` cancels but does not `<-done` | **FAIL** — all three subtests |
| `close()` nils `m.cancel`/`m.done` before unlocking | **FAIL** (`TestMonitorConcurrentClose`) |
| `Replica.Close()` no longer closes its poller | **FAIL** — `replica` subtest only |

A second `start()` overwrites `m.cancel` and `m.done`, so the first goroutine becomes both uncancellable and unjoinable — the precise leak this PR removes, reachable through the primitive's own unguarded path. `Open` being called twice isn't a path today, but `monitorLoop` is now shared by three throttlers and the guard is part of its contract. Two `start` calls with a run func that increments a counter, then asserting one invocation, pins it.

⭐ Worth keeping the M4 result in mind: `close()` deliberately does *not* nil the pair, which is what makes repeated and concurrent `close()` naturally safe (a receive on an already-closed `done` returns at once). #1197's `StopPeriodicFlush` does the opposite in `pkg/change` and has to be safe for a different reason. Since both land in this batch, `monitorLoop` looks like the answer for those two hand-rolled copies too.

### 3 — `Close` is now a blocking call inside a sequential loop (low)

```go
func (m *multiThrottler) Close() error {
	var errs []error
	for _, t := range m.throttlers {
		if err := t.Close(); err != nil {
			errs = append(errs, err)
		}
	}
```

Child `Close`es used to return in ~0 and now each waits for its in-flight query to unwind, so the total is the sum rather than the max — and a child that is slow to honor cancellation leaves every child *after* it un-cancelled while `Runner.Close` proceeds to `monitorDB.Close()` / `closeReplicas()`. That is the yank the runner comment exists to prevent, just relocated to the second child onward. Closing children concurrently (each already independently safe) keeps the guarantee whole and makes Close latency the max instead.

---

<details>

**The start/close interleaving is airtight.** `start` holds `m.mu` across the `go` statement, so a concurrent `close` either runs first (`cancel == nil`, returns, and the later `start` sees `m.closed`) or second (finds the registered pair and joins it). There is no window where a goroutine exists but its cancel is unregistered — which is the failure mode a `Lock`/`Unlock` around registration only, with the spawn outside, would have. `close` releases the mutex before `cancel(); <-done`, so nothing deadlocks against a `run` body that wants the lock.

**Repeated and concurrent `close` are safe by construction**, not just by test: `m.cancel`/`m.done` are retained, so the second caller re-runs an idempotent `cancel()` and receives immediately from the closed channel. `TestMonitorConcurrentClose` covers ten at once and `-race` over 5 runs is clean.

**Removing the in-loop `isClosed.Load()` check loses nothing** — `close()` cancels the derived context, which the `select`'s `<-ctx.Done()` arm already handles, and `isShutdownError(ctx, err)` still absorbs the cancelled in-flight query rather than logging it as a monitoring failure. No `isClosed` references remain anywhere in `pkg/throttler`.

**The test covers wiring, not just the primitive.** Driving each public `Close()` through the table (`replica` / `threads` / `latency`) is what catches a one-sided fix — M5 shows the `replica` subtest failing alone while the other two stay green. It needs no MySQL (zero-valued structs, stub `run`), so the lifecycle really is covered in environments where Aurora sampling isn't.

**Dropping the test's `loopInterval = 1 * time.Millisecond` is a genuine improvement** — that was a package-level var mutated from one test and never restored, visible to every other test in the package. Nothing assigns it now, so it could become a `const` alongside `blockWaitInterval` to keep that hazard from returning. Removing the `time.Sleep(50ms)` teardown is now correct rather than merely tidy, since `Close` actually joins.

**`Open` partial failure is safe**: `UpdateLag` failing returns before `start`, and a later `Close` finds `cancel == nil` and returns immediately — `TestMonitorCloseBeforeStart` covers that ordering directly.

**Local:** `go build ./...` clean, `gofmt` clean, `golangci-lint run ./pkg/throttler/...` → **0 issues**, `go test -race ./pkg/throttler -run TestMonitor -count=5` passes. CI 18/18 green.

</details>

aparajon (approved):
🤖 Approving — `monitorLoop` is the right primitive and the start/close interleaving is airtight. Notably this is what finally makes the existing ordering comment in `runner.go` true (before it, `Close()` set a flag the poller only noticed up to 5s later, so the pool really could be yanked from under an in-flight sample) — worth putting in the body. Three non-blocking findings in the review comment above, the main one being that `Open` after `Close` returns nil having started nothing, which leaves an Aurora throttler frozen at its Open-time verdict with no staleness guard behind it.

morgo:
🤖 Open now returns an error for a closed throttler before initial sampling, with a second check when registering the monitor. Added coverage through all three public throttler implementations, plus a duplicate-start test that verifies the same worker is retained and invoked once. Expanded the PR description to explain why joining pollers makes the runner's database-pool teardown safe.

I kept multiThrottler.Close sequential: Runner.Close calls it synchronously and cannot proceed to monitorDB.Close or closeReplicas until every child Close has returned. A slow child increases shutdown latency, but does not permit pool teardown ahead of the later children. Concurrent child shutdown would be a separate latency optimization.

Validation: the full throttler package passed five race-enabled repetitions; lint reports 0 issues.

Implemented in f00260ea.

Issue #858, opened 2026-05-17
flaky test: TestProcessRowsEventDoesNotDeadlockOnPark


morgo:
```
2026-05-17T16:24:24.9182337Z test-1           | --- FAIL: TestProcessRowsEventDoesNotDeadlockOnPark (5.98s)
2026-05-17T16:24:24.9182472Z test-1           |     subscription_buffered_test.go:1592: 
2026-05-17T16:24:24.9182659Z test-1           |         	Error Trace:	/app/pkg/repl/subscription_buffered_test.go:1592
2026-05-17T16:24:24.9182784Z test-1           |         	Error:      	Condition never satisfied
2026-05-17T16:24:24.9182951Z test-1           |         	Test:       	TestProcessRowsEventDoesNotDeadlockOnPark
2026-05-17T16:24:24.9183146Z test-1           |         	Messages:   	binlog-driven HasChanged should park on soft limit

```

Flaky on 8.0.42-CI.

morgo:
🤖 Closing as a duplicate of #1163. Both TestProcessRowsEventDoesNotDeadlockOnPark and TestClientCloseUnblocksParkedHasChanged use the same seed-row / lowered-soft-limit arrangement and can fail waiting to observe a park before reaching the behavior they intend to verify.

This closure consolidates tracking; it does not claim either flaky setup is fixed. #1163 should cover deterministic park setup for both tests, preserving their separate Flush-progress and Close-unblocks-reader assertions.

PR #1201, opened 2026-09-05
test(change): establish reader parking deterministically

### LINKED ISSUE #1163: flaky test: TestClientCloseUnblocksParkedHasChanged — seeded bytes can be drained before the park, so timesParked never increments
🤖 _Filed by Morgan's AI agent from a scheduled review of `main` CI failures._

## Symptom

`TestClientCloseUnblocksParkedHasChanged` never observed the binlog-driven `HasChanged` park it is set up to force, so the 5-second `require.Eventually` expired before the test reached the behaviour it actually guards (that `client.Close()` wakes a parked reader).

- Run: https://github.com/block/spirit/actions/runs/32051525508
- Job: MySQL 8.0.45 (with replicas) /w docker-compose
- Commit: `2b84a4ac` ("statement: converge expression DEFAULTs that MySQL rewrites (#1153)")

```
--- FAIL: TestClientCloseUnblocksParkedHasChanged (6.32s)
    subscription_buffered_test.go:1688:
        Error Trace:	/app/pkg/change/subscription_buffered_test.go:1688
        Error:      	Condition never satisfied
        Test:       	TestClientCloseUnblocksParkedHasChanged
        Messages:   	binlog-driven HasChanged should park on soft limit
```

## Likely cause: the setup races

The test's arrangement is:

```go
// Seed one row so sizeBytes > 0, then crank the soft limit down.
testutils.RunSQL(t, "INSERT ... (1, 'seed')")
require.NoError(t, client.BlockWait(t.Context()))
sub.Lock()
sub.softLimitBytes = 1
sub.Unlock()

// Trigger a second INSERT; the binlog reader will pick it up and park.
testutils.RunSQL(t, "INSERT ... (2, 'parked')")
require.Eventually(t, func() bool {
    return sub.timesParked.Load() >= 1
}, 5*time.Second, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
```

Two ways this can miss:

1. **The seed gets drained.** The park depends on `sizeBytes > 0` when the second event arrives. Nothing pins the seeded row in the map — a periodic/opportunistic flush landing between `BlockWait` and the second INSERT drops `sizeBytes` back to 0, and the second event fits under the (now-irrelevant) soft limit without parking.
2. **The second event is simply late.** 5 seconds is a short budget for a binlog round-trip on a contended runner. The surrounding log for this run shows heavy concurrent `flushBatch` traffic from other tests in the same package at that moment.

Note the test is well-behaved about cleanup — the `t.Cleanup` wakes any parked reader — so this fails cleanly rather than hanging. It is a false negative, not a real regression in the Close-wakes-parker behaviour that #863 fixed.

## Related

- #863 — _fix(repl): release parked subscriptions on Close + snapshot subs in flush_ — the fix this test guards.
- Open #858 — _flaky test: TestProcessRowsEventDoesNotDeadlockOnPark_ — different test, same park machinery. These two may deserve a common fix (a deterministic way for a test to force a park) rather than two independent patches.
- #1120 / #1121 — soft-limit park starving the change reader.

## Frequency

1 occurrence in 96 MySQL matrix jobs on `main` over 2026-08-17 → 2026-08-24.

## Possible directions

- Disable/suspend periodic flushing for the duration of this test so the seeded bytes cannot vanish, then assert the park deterministically.
- Give the test a hook to force a park directly (a test-only knob on `bufferedMap`) instead of arranging one through real binlog traffic — and keep the end-to-end path covered by a separate, more forgiving test.
- At minimum, re-check `sizeBytes > 0` immediately before the second INSERT and raise the `Eventually` budget; this reduces the flake rate without addressing the underlying race.

morgo:
Four binlog-driven parking tests used a seed INSERT, a catch-up wait, and a later limit change. Share a single-statement arrangement across all four: set a one-change cap, then insert two distinct keys. The first fills the buffer and the second parks the actual binlog reader, even if MySQL splits the statement across row events.

Flush/Close lifecycle tests require the reader to be currently parked and require callers to keep periodic flushing stopped until release. The periodic-flush tests instead observe the monotonic park counter, since releasing that park is their intended behavior. Use a dedicated 30-second test budget for real binlog delivery, independent of production BlockWait tuning.

Fixes #1163. Also covers the companion test formerly tracked in #858 and both periodic-flush park-request tests.

Validation: all four integration tests passed 30 repetitions under the race detector against MySQL 8.0.43; golangci-lint reports 0 issues. No production behavior changes.

### DIFF
--- a/pkg/change/subscription_buffered_concurrency_test.go
+++ b/pkg/change/subscription_buffered_concurrency_test.go
@@ -490,19 +490,11 @@ func TestPeriodicFlushRespondsToParkRequest(t *testing.T) {
 	client.StartPeriodicFlush(t.Context(), time.Hour)
 	defer client.StopPeriodicFlush()
 
-	// Seed one buffered change, then clamp the limit so the next
-	// binlog-driven HasChanged parks.
-	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (1, 'seed')", srcTable.QuotedTableName))
-	require.NoError(t, client.BlockWait(t.Context()))
-	sub.Lock()
-	require.Positive(t, sub.sizeBytes, "seed change must be accounted")
-	sub.softLimitBytes = 1
-	sub.Unlock()
-
-	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (2, 'parked')", srcTable.QuotedTableName))
+	requestBinlogPark(t, sub, srcTable)
+	// The flusher may already have released it, so observe the park counter.
 	require.Eventually(t, func() bool {
 		return sub.timesParked.Load() >= 1
-	}, 10*time.Second, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
+	}, parkWaitTimeout, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
 
 	// The park requested a flush; the periodic goroutine must drain the
 	// seed row promptly — not in an hour — which also unparks the reader.
@@ -590,19 +582,11 @@ func TestPeriodicFlushPrioritizesParkedSubscription(t *testing.T) {
 	// B accumulates pending changes first, so it is a candidate to be
 	// drained ahead of A in an unprioritized all-subscription pass.
 	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (1, 'b'), (2, 'b'), (3, 'b')", srcB.QuotedTableName))
-	// Seed A with one buffered change, then clamp its limit so the next
-	// binlog-driven event parks the reader on A.
-	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (1, 'seed')", srcA.QuotedTableName))
-	require.NoError(t, client.BlockWait(t.Context()))
-	subA.Lock()
-	require.Positive(t, subA.sizeBytes, "seed change must be accounted")
-	subA.softLimitBytes = 1
-	subA.Unlock()
-
-	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (2, 'parked')", srcA.QuotedTableName))
+	requestBinlogPark(t, subA, srcA)
+	// The flusher may already have released it, so observe the park counter.
 	require.Eventually(t, func() bool {
 		return subA.timesParked.Load() >= 1
-	}, 10*time.Second, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
+	}, parkWaitTimeout, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
 
 	// The priority flush drains A; the follow-up pass drains B. Gate on B
 	// specifically rather than on a bare count: the all-subscription pass
--- a/pkg/change/subscription_buffered_test.go
+++ b/pkg/change/subscription_buffered_test.go
@@ -1560,21 +1560,7 @@ func TestProcessRowsEventDoesNotDeadlockOnPark(t *testing.T) {
 
 	sub := getBufferedMap(t, client, srcTable.SchemaName+"."+srcTable.TableName)
 
-	// Seed a row so sizeBytes > 0, then drop the soft limit. Any
-	// further HasChanged from the binlog reader will park.
-	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (1, 'seed')", srcTable.QuotedTableName))
-	require.NoError(t, client.BlockWait(t.Context()))
-	sub.Lock()
-	sub.softLimitBytes = 1
-	sub.Unlock()
-
-	// Trigger a second INSERT. The binlog reader will pick this up
-	// asynchronously, route it through processRowsEvent, call
-	// sub.HasChanged — which must park on the soft limit.
-	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (2, 'parked')", srcTable.QuotedTableName))
-	require.Eventually(t, func() bool {
-		return sub.timesParked.Load() >= 1
-	}, 5*time.Second, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
+	parkBinlogReader(t, client, sub, srcTable)
 
 	// While HasChanged is parked from inside processRowsEvent,
 	// client.Flush must still make progress. With the lock-release
@@ -1595,6 +1581,29 @@ func TestProcessRowsEventDoesNotDeadlockOnPark(t *testing.T) {
 	require.NoError(t, db.QueryRowContext(t.Context(),
 		fmt.Sprintf("SELECT COUNT(*) FROM %s", dstTable.QuotedTableName)).Scan(&count))
 	require.Equal(t, 2, count)
+}
+
+// parkBinlogReader arranges both rows in a single statement/event: admitting
+// the first reaches the one-change cap, so the second must park. There is no
+// seed/catch-up/reconfigure window and no dependence on row-image byte sizing.
+// These clients have no periodic flusher: only the test's Flush or Close may
+// release the park. Keep real binlog delivery in the lifecycle regression.
+func parkBinlogReader(t *testing.T, client *binlogClient, sub *bufferedMap, src *table.TableInfo) {
+	t.Helper()
+	client.periodicFlushLock.Lock()
+	flushing := client.periodicFlushCancel != nil
+	client.periodicFlushLock.Unlock()
+	require.False(t, flushing, "a background flusher could release the test's park")
+	sub.Lock()
+	sub.softLimitBytes = 0
+	sub.softLimitChanges = 1
+	sub.Unlock()
+	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (1, 'seed'), (2, 'parked')", src.QuotedTableName))
+	require.Eventually(t, func() bool {
+		sub.Lock()
+		defer sub.Unlock()
+		return sub.parked && sub.lengthLocked() == 1
+	}, DefaultTimeout, 10*time.Millisecond, "binlog reader must be held at the second row")
 }
 
 // TestBufferedMapCloseUnblocksParkedHasChanged pins the invariant that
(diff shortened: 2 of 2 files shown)


aparajon:
🤖 **Adversarial correctness review** — `d85e5cd2` (+25/-29, 1 file)

This is the right fix, and it's the one #1163 asked for: *"These two may deserve a common fix (a deterministic way for a test to force a park) rather than two independent patches."* The single-statement arrangement is a genuine structural improvement over the seed/catch-up/clamp sequence, not a re-budgeting — and the new assertion is strictly stronger than what it replaces. `timesParked.Load() >= 1` is monotonic ("ever parked"), so it stays true after a wake; `parked && lengthLocked() == 1` read under the Mutex is instantaneous ("parked right now"). `parked` is set at `subscription_buffered.go:1017` and cleared at `:1021`, both inside the `cond.Wait` critical section, so the read means a goroutine is actually suspended — which is exactly the precondition both tests need before they touch `Flush`/`Close`.

The mechanism also holds under inspection. `sub.HasChanged` is called **per row** in `processRowsEvent`'s INSERT loop (`binlog.go:1002-1013`), and `setupBufferedTest` creates fresh empty tables with no inserts, so the buffer is provably empty at helper entry: row 1 admits (`length` 0 → 1), row 2 hits `softLimitChanges = 1` and parks with `length == 1`. That's the comment's stated arithmetic, exactly.

Findings are about the reach of the helper rather than its correctness.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | med | `subscription_buffered_concurrency_test.go:482`, `:548` | The arrangement this PR removes survives verbatim in two more binlog-driven park tests, and the helper's own flusher guard makes it unusable there |
| 2 | low-med | `subscription_buffered_test.go:1606` | `DefaultTimeout` is the production `BlockWait` knob, so the flake window is now coupled to unrelated tuning |
| 3 | low | `subscription_buffered_test.go:1590-1594` | The flusher guard is a point-in-time assertion described as a property of the clients |

---

### 1 — two sibling tests keep the arrangement this PR exists to delete (med)

`TestPeriodicFlushRespondsToParkRequest` and `TestPeriodicFlushPrioritizesParkedSubscription` park the binlog reader the old way, character for character:

```go
	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (1, 'seed')", srcTable.QuotedTableName))
	require.NoError(t, client.BlockWait(t.Context()))
	sub.Lock()
	require.Positive(t, sub.sizeBytes, "seed change must be accounted")
	sub.softLimitBytes = 1
	sub.Unlock()

	testutils.RunSQL(t, fmt.Sprintf("INSERT INTO %s (id, name) VALUES (2, 'parked')", srcTable.QuotedTableName))
	require.Eventually(t, func() bool {
		return sub.timesParked.Load() >= 1
	}, 10*time.Second, 10*time.Millisecond, "binlog-driven HasChanged should park on soft limit")
```

Both of #1163's named causes are still live there. Cause 1 (seeded bytes vanishing) is *more* reachable, not less, because these two deliberately run `client.StartPeriodicFlush(t.Context(), time.Hour)` — the hour interval means only a park request can trigger a drain today, so it holds by an argument about the interval rather than by construction. And cause 2 (a late binlog round-trip on a contended runner) applies unchanged, with the added exposure that `require.NoError(t, client.BlockWait(t.Context()))` asserts on a 30s production wait as a test *precondition*: a runner slow enough to make the park late is a runner that can fail the setup line outright.

What makes this worth raising rather than filing separately is that the helper as written **cannot** be reused there — `require.False(t, flushing, "a background flusher could release the test's park")` fails immediately on both, by design. So the PR's abstraction is bounded precisely at the tests that most need it, and nothing says so.

The useful decomposition is that the two halves of the helper transfer independently:

- **The single-statement INSERT transfers cleanly.** It is what removes the seed, the `BlockWait`, and the reconfigure window, and it works regardless of a flusher: row 1 admits, row 2 parks, with no test-side gap in between.
- **The instantaneous `parked` assertion does not transfer.** With a flusher running, a drain may legitimately release the park before the poll observes it, so those tests genuinely need the monotonic `timesParked.Load() >= 1` — and that's fine there, because a released park is the behaviour under test.

So a version of `parkBinlogReader` that takes the flusher expectation as a parameter (or splits the arrangement from the assertion) would cover all four sites and let the two `TestPeriodicFlush*` tests drop their seed/`BlockWait` window too. Failing that, the PR should say those two are out of scope and why, since #1163 will read as fixed once this merges.

### 2 — `DefaultTimeout` is a production knob standing in for a test budget (low-med)

```go
	}, DefaultTimeout, 10*time.Millisecond, "binlog reader must be held at the second row")
```

The bump from `5*time.Second` is justified — #1163's cause 2 is a genuinely late event, not a masked bug, so a larger budget is the correct response and not the thing the "don't raise timeouts to fix flakes" instinct is aimed at. The anchor is what I'd change. `DefaultTimeout` is declared at `repl.go:115` as *"how long BlockWait is supposed to wait before returning errors"* — a production tuning value. Tying the test to it couples the flake-detection window to unrelated tuning in both directions: raise it for slow replicas and this test silently stops catching a real 40s regression; lower it and the test tightens for a reason that has nothing to do with parking.

It's also 15x this file's own convention for the same assertion — `TestBufferedMapCloseUnblocksParkedHasChanged` polls the in-process park at `2*time.Second` (`:1641`), and this file had no prior use of `DefaultTimeout` at all. A named constant next to the helper (`parkWaitTimeout = 30 * time.Second`, with the "contended runner, real binlog round-trip" reason on it) says what the number is for and can't drift with a production change.

### 3 — the flusher guard is stated as a property, checked as a snapshot (low)

```go
	// These clients have no periodic flusher: only the test's Flush or Close may
	// release the park.
```

The comment reads as an invariant of `setupBufferedTest`'s clients; the code is a one-shot read of `periodicFlushCancel` at helper entry, and nothing stops a caller starting a flusher afterwards. The guard is still good defensive design — `flushRequests` is drained only by the periodic-flush loop (`binlog.go:1381`), so a flusher is the one thing that could satisfy the park's own `requestFlush` and unpark the reader before the poll runs, and this will fail loudly if someone adds `StartPeriodicFlush` to the shared setup. Worth phrasing as the precondition it is ("this helper requires no periodic flusher, because…") rather than as a fact about the clients — especially given finding 1, where that precondition is what bounds the helper's reach.

**Nit, same helper:** two things the arrangement depends on aren't in the comment. The PKs must differ — map-mode `dedupOverwrite` bypasses the soft-limit check entirely (`subscription_buffered.go:1006`), so two rows sharing a key would never park and the helper would spin for the full budget with a misleading message. And "a single statement/event" is stronger than needed: if MySQL split the statement across two row events at `binlog-row-event-max-size`, event 1 would admit and event 2 would park, same outcome. The load-bearing property is *one statement*, not one event — worth saying, so a future reader doesn't assume a wide-row table defeats the helper.

---

<details>
<summary><b>Verified</b> — the cap switch, and three attacks that dissolved</summary>

⭐ **Attack that dissolved: `softLimitBytes = 0` looks like it makes every row park.** Zero *disables* the byte cap — `overSoftLimitLocked` guards it with `s.softLimitBytes > 0 && s.sizeBytes >= s.softLimitBytes` (`:559`), and `binlog.go:122` documents zero as the disable. Setting it to 0 is the right move rather than a bug: it removes the byte dimension entirely so only the count cap can fire, which is what makes the park land on a known row. It also matches how `subscription_buffered_drainbound_test.go` sets up count-cap tests (`softLimitBytes = 0` alongside `softLimitChanges = limit`), so the helper is consistent with the package's existing count-park convention.

⭐ **Attack that dissolved: the assertion could pass on the wrong row.** If the buffer already held a change, row 1 would park and `parked && lengthLocked() == 1` would still be true — for a mechanism the comment doesn't describe, and `TestProcessRowsEventDoesNotDeadlockOnPark`'s closing `require.Equal(t, 2, count)` would then be asserting something different. It can't happen: `setupTestTables` does `DROP TABLE IF EXISTS` + two `CREATE TABLE`s and no inserts, and `setupBufferedTest` adds the subscription and starts the client without writing a row. The buffer is empty by construction, so the described path is the one exercised.

⭐ **Attack that dissolved: switching from the byte cap to the count cap could change who wakes the parker.** It does change the mechanism, and non-obviously — `lengthLocked()` is `len(changes) + len(queue) + flushingCount` (`:546`), so a count-parked reader stays parked while a drain has entries swapped out but not yet landed, whereas a bytes-parked one was released as soon as bytes were. That would be a hang if any `flushingCount` decrement missed a broadcast. Every one is paired with `s.cond.Broadcast()` under the lock: `:1764` (map batch), `:1795` (queue segment), and `:1873` (`reattachLocked`, the error path, which zeroes it). So the wake is sound on both the success and failure paths, which is what `TestProcessRowsEventDoesNotDeadlockOnPark` needs when its `Flush` is the thing releasing the park.

**The choice to keep real binlog delivery is deliberate and recorded.** #1163 offered a test-only knob on `bufferedMap` as an alternative; the helper's *"Keep real binlog delivery in the lifecycle regression"* is the counter-argument, and it's the right call for these two tests specifically — the deadlock they pin is *`processRowsEvent` holding `c.Lock` across the park*, which a direct `HasChanged` call from a test goroutine would not exercise at all.

**The closing `require.Equal(t, 2, count)` still holds, via `Flush`'s redrain path rather than via `Flush` being complete.** After the first `c.flush` releases the park, row 2 is admitted behind it; `Flush` continues because `backlogWorthDraining(pending, readerWasBlocked, …)` returns true on `readerWasBlocked` alone (`repl.go:262`), independent of `pending` — which matters because `pending` is sampled immediately after the drain and could read 0 if the woken goroutine hasn't re-appended yet. Without that `readerWasBlocked` term the loop would fall through to `BlockWait` and break on `GetDeltaLen() < 10000` with row 2 unflushed. Unchanged by this PR (the old arrangement ended the same way), and not a defect — worth knowing that the assertion rests on the park-detection term and not on a guarantee that `Flush` drains everything.

**Local:** `go build ./...` clean, `go vet ./pkg/change/` clean, `gofmt -l pkg/change/` empty, `golangci-lint run ./pkg/change/...` → **0 issues**. CI 18/18 green. Both tests need a live MySQL source, so they were read rather than run here — the PR's 50 race-detector repetitions cover that, and finding 2 is the reason I'd expect them to stay green.

</details>

aparajon (approved):
🤖 Approving — the single-statement arrangement is the common fix #1163 asked for, and `parked && lengthLocked() == 1` is a strictly stronger precondition than the monotonic `timesParked` poll it replaces. Findings are about reach, not correctness: the same arrangement survives in the two `TestPeriodicFlush*` park tests (and the helper's flusher guard blocks reuse there), and `DefaultTimeout` is a production `BlockWait` knob rather than a test budget. Details in the review comment.

morgo:
🤖 Addressed all three findings. Split the shared single-statement arrangement from the held-park assertion and reused it in TestPeriodicFlushRespondsToParkRequest and TestPeriodicFlushPrioritizesParkedSubscription. Those tests retain their monotonic park-counter assertions because a released park is the behavior under test.

Introduced parkWaitTimeout = 30s for real binlog delivery independently of production BlockWait tuning. Reworded the flusher guard as a caller precondition, documented why the two primary keys must differ, and clarified that one statement is sufficient even if MySQL splits row events.

All four integration tests passed 30 race-enabled repetitions; lint reports 0 issues.

PR #1204, opened 2026-09-05
test(move): bound VARCHAR-PK writers before cutover

### LINKED ISSUE #834: Test timeout: TestMoveWithVarcharPK
```
test-1           | panic: test timed out after 10m0s
test-1           | 	running tests:
test-1           | 		TestMoveWithVarcharPK (9m30s)
test-1           | 		TestMoveWithVarcharPK/force-enable-buffered-map (9m4s)
test-1           | 
test-1           | goroutine 198376 [running]:
test-1           | testing.(*M).startAlarm.func1()
test-1           | 	/usr/local/go/src/testing/testing.go:2802 +0x605
test-1           | created by time.goFunc
test-1           | 	/usr/local/go/src/time/sleep.go:215 +0x45
test-1           | 
test-1           | goroutine 1 [chan receive, 9 minutes]:
test-1           | testing.(*T).Run(0xc000162908, {0x10f8aed, 0x15}, 0x1138598)
test-1           | 	/usr/local/go/src/testing/testing.go:2109 +0xb3e
test-1           | testing.runTests.func1(0xc000162908)
test-1           | 	/usr/local/go/src/testing/testing.go:2585 +0x85
test-1           | testing.tRunner(0xc000162908, 0xc0003319c8)
test-1           | 	/usr/local/go/src/testing/testing.go:2036 +0x21d
test-1           | testing.runTests({0x10fb11e, 0x17}, {0x1105986, 0x20}, 0xc0000163c0, {0x203c040, 0xf, 0xf}, {0xc27866a3306c3951, 0x8bb342e699, ...})
test-1           | 	/usr/local/go/src/testing/testing.go:2583 +0x9ea
test-1           | testing.(*M).Run(0xc00011fa40)
test-1           | 	/usr/local/go/src/testing/testing.go:2443 +0xf4c
test-1           | go.uber.org/goleak.VerifyTestMain({0x1141d00, 0xc00011fa40}, {0x0, 0x0, 0x0})
test-1           | 	/go/pkg/mod/go.uber.org/goleak@v1.3.0/testmain.go:53 +0x65
test-1           | github.com/block/spirit/pkg/move.TestMain(...)
test-1           | 	/app/pkg/move/move_test.go:23
test-1           | main.main()
test-1           | 	_testmain.go:76 +0x1d4
test-1           | 
test-1           | goroutine 12003 [runnable]:
test-1           | regexp/syntax.(*compiler).compile(0xc0070b5bb8, 0xc0025e92d0)
test-1           | 	/usr/local/go/src/regexp/syntax/compile.go:89 +0x28db
test-1           | regexp/syntax.(*compiler).compile(0xc0070b5bb8, 0xc0025e9340)
test-1           | 	/usr/local/go/src/regexp/syntax/compile.go:147 +0x213c
test-1           | regexp/syntax.Compile(0xc0025e9340)
test-1           | 	/usr/local/go/src/regexp/syntax/compile.go:74 +0x1fb
test-1           | regexp.compile({0x10eab5f, 0xa}, 0xd4, 0x0)
test-1           | 	/usr/local/go/src/regexp/regexp.go:176 +0xb9
test-1           | regexp.Compile(...)
test-1           | 	/usr/local/go/src/regexp/regexp.go:131
test-1           | regexp.MustCompile({0x10eab5f, 0xa})
test-1           | 	/usr/local/go/src/regexp/regexp.go:311 +0x49
test-1           | database/sql.(*DB).retry(0x68?, 0xc0097b5dc8)
test-1           | 	/usr/local/go/src/database/sql/sql.go:1576 +0x4b
test-1           | database/sql.(*DB).ExecContext(0xc000319c70, {0x1148770, 0xc00047ec30}, {0xc00215d490, 0x68}, {0xc0097b5f20, 0x1, 0x1})
test-1           | 	/usr/local/go/src/database/sql/sql.go:1671 +0x185
test-1           | github.com/block/spirit/pkg/move.varcharPKWriteOne({0x1148770, 0xc00047ec30}, 0xc000319c70, {0xc001604090, 0x17})
test-1           | 	/app/pkg/move/runner_test.go:564 +0x168
test-1           | github.com/block/spirit/pkg/move.testMoveWithVarcharPK.func1()
test-1           | 	/app/pkg/move/runner_test.go:517 +0x97
test-1           | sync.(*WaitGroup).Go.func1()
test-1           | 	/usr/local/go/src/sync/waitgroup.go:258 +0x4a
test-1           | created by sync.(*WaitGroup).Go in goroutine 12003
test-1           | 	/usr/local/go/src/sync/waitgroup.go:238 +0x73
test-1           | 
test-1           | goroutine 9979 [chan receive, 9 minutes]:
test-1           | testing.(*T).Run(0xc000456008, {0x10fd8e9, 0x19}, 0x1138730)
test-1           | 	/usr/local/go/src/testing/testing.go:2109 +0xb3e
test-1           | github.com/block/spirit/pkg/move.TestMoveWithVarcharPK(0xc000456008)
test-1           | 	/app/pkg/move/runner_test.go:457 +0x57
test-1           | testing.tRunner(0xc000456008, 0x1138598)
test-1           | 	/usr/local/go/src/testing/testing.go:2036 +0x21d
test-1           | created by testing.(*T).Run in goroutine 1
test-1           | 	/usr/local/go/src/testing/testing.go:2101 +0xb13
test-1           | 
test-1           | goroutine 15608 [chan receive, 8 minutes]:
test-1           | github.com/block/spirit/pkg/applier.(*SingleTargetApplier).writeWorker(0xc000458000, {0x1148770, 0xc00ecb9bd0})
test-1           | 	/app/pkg/applier/single_target.go:294 +0x4ab
test-1           | created by github.com/block/spirit/pkg/applier.(*SingleTargetApplier).Start in goroutine 12003
test-1           | 	/app/pkg/applier/single_target.go:129 +0x4f9
test-1           | 
test-1           | goroutine 187231 [select]:
test-1           | github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1()
test-1           | 	/go/pkg/mod/github.com/go-sql-driver/mysql@v1.9.3/connection.go:664 +0x1e9
test-1           | created by github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher in goroutine 12255
test-1           | 	/go/pkg/mod/github.com/go-sql-driver/mysql@v1.9.3/connection.go:655 +0x1c5
test-1           | FAIL	github.com/block/spirit/pkg/move	600.076s
```

morgo:
TestMoveWithVarcharPK stops its writers only after the move returns. Even paced writers can keep a slow reader chasing a changing source indefinitely, so completion and writer shutdown depend on each other.

Use a finite workload while holding cutover at the sentinel. After copy switches VARCHAR primary keys to FIFO replay, four writers execute 25 paced write cycles each. Require all 100 cycles to succeed, stop writing, then drop the sentinel and verify cutover/data agreement. The test explicitly covers post-copy FIFO replay; concurrent map-mode copy is outside its coverage.

Setup/copy/checksum, writes, and final drain each receive their own three-minute budget. The runner is cancelled, joined and closed during cleanup. Consolidate all three move sentinel waits into a helper that detects early exit and terminal states and accepts reaching or passing the requested state. Retain the #834 explanation for finite writers and pacing.

Fixes #834's unbounded VARCHAR test-workload scenario.

Validation on MySQL 8.0.43: the VARCHAR workload, shared-wait regressions, sentinel-drop test, and sharded vindex-failure test passed three repetitions under -race. The full move package passed under -race; golangci-lint reports 0 issues.

### DIFF
--- a/pkg/move/helpers_test.go
+++ b/pkg/move/helpers_test.go
@@ -1,0 +1,62 @@
+package move
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/block/spirit/pkg/status"
+	"github.com/stretchr/testify/require"
+)
+
+// Give setup/copy/checksum its own budget, independent of subsequent writes
+// and cutover. Like migration's status wait, allow three minutes for checksum
+// catch-up and lock retries on slow CI servers.
+func waitForMoveStatus(t *testing.T, runner *Runner, target status.State, done <-chan error) {
+	t.Helper()
+	ctx, cancel := context.WithTimeout(t.Context(), 3*time.Minute)
+	defer cancel()
+	require.NoError(t, awaitMoveStatus(ctx, runner, target, done))
+}
+
+func awaitMoveStatus(ctx context.Context, runner *Runner, target status.State, done <-chan error) error {
+	ticker := time.NewTicker(10 * time.Millisecond)
+	defer ticker.Stop()
+	for {
+		select {
+		case err := <-done:
+			return fmt.Errorf("move exited while waiting for %s: %w", target, err)
+		default:
+		}
+		current := runner.status.Get()
+		if current >= status.Close {
+			return fmt.Errorf("move entered terminal state %s while waiting for %s", current, target)
+		}
+		if current >= target {
+			return nil
+		}
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("waiting for %s (last state %s): %w", target, runner.status.Get(), ctx.Err())
+		case err := <-done:
+			return fmt.Errorf("move exited while waiting for %s: %w", target, err)
+		case <-ticker.C:
+		}
+	}
+}
+
+func TestAwaitMoveStatus(t *testing.T) {
+	runner := &Runner{}
+	runner.status.Set(status.CutOver)
+	require.NoError(t, awaitMoveStatus(t.Context(), runner, status.WaitingOnSentinelTable, nil))
+	runner.status.Set(status.Close)
+	require.ErrorContains(t, awaitMoveStatus(t.Context(), runner, status.WaitingOnSentinelTable, nil), "terminal state")
+	runner.status.Set(status.Initial)
+	done := make(chan error, 1)
+	done <- fmt.Errorf("copy failed")
+	require.ErrorContains(t, awaitMoveStatus(t.Context(), runner, status.WaitingOnSentinelTable, done), "copy failed")
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel()
+	require.ErrorIs(t, awaitMoveStatus(ctx, runner, status.WaitingOnSentinelTable, nil), context.Canceled)
+}
--- a/pkg/move/move_sharded_test.go
+++ b/pkg/move/move_sharded_test.go
@@ -410,17 +410,7 @@ func TestShardedMoveVindexUpdateFails(t *testing.T) {
 
 	// Wait until the move blocks on the sentinel: the copy and the initial
 	// checksum are done and the repl client is streaming.
-	deadline := time.Now().Add(2 * time.Minute)
-	for runner.status.Get() != status.WaitingOnSentinelTable {
-		select {
-		case err := <-errCh:
-			t.Fatalf("move finished before reaching the sentinel wait: %v", err)
-		case <-time.After(50 * time.Millisecond):
-		}
-		if time.Now().After(deadline) {
-			t.Fatal("timed out waiting for the move to reach the sentinel wait")
-		}
-	}
+	waitForMoveStatus(t, runner, status.WaitingOnSentinelTable, errCh)
 
 	// Change the vindex value of a row. The change feed must treat this as
 	// fatal and cancel the move.
--- a/pkg/move/runner_test.go
+++ b/pkg/move/runner_test.go
@@ -745,16 +745,13 @@ func TestConcurrentMoveDoesNotWipeTarget(t *testing.T) {
 	require.Zero(t, artifacts, "a concurrent run must not create a checkpoint table before acquiring the lock")
 }
 
-// TestMoveWithVarcharPK verifies a move on a table with a non-memory-comparable
-// primary key (VARCHAR with a CI collation) — the case from issue #607. The
-// move runs under concurrent writes to exercise the binlog replay path.
-//
-// For non-memory-comparable PKs the bufferedMap subscription uses LWW map
-// dedup during the copy phase and FIFO queue post-copy. The queue replays
-// binlog events in their original order, which is required for collation-
-// sensitive PKs because the map's hash equality ("A" ≠ "a") does not match
-// MySQL's row identity ("A" = "a" under a CI collation). The post-cutover
-// checksum keeps the optimization honest by repairing any divergence.
+// TestMoveWithVarcharPK verifies FIFO replay after copying a table with a
+// non-memory-comparable VARCHAR primary key (issue #607). Concurrent writes
+// run while the sentinel holds cutover, after the switch from LWW map dedup
+// to FIFO. This test does not exercise concurrent writes during map-mode copy;
+// the move checksum is the correctness gate for divergence in that phase.
+// FIFO preserves event order when Go key equality differs from MySQL's CI
+// collation ("A" and "a" identify the same row in MySQL).
 func TestMoveWithVarcharPK(t *testing.T) {
 	srcDB := "source_varcharpk"
 	dstDB := "dest_varcharpk"
@@ -791,7 +788,7 @@ func TestMoveWithVarcharPK(t *testing.T) {
 	// Hold cutover at the sentinel so the finite workload definitely runs
 	// after copy, when VARCHAR primary keys use FIFO replay. The old writers
 	// ran until move.Run returned, so a slow reader could chase them forever.
-	ctx, cancel := context.WithTimeout(t.Context(), 3*time.Minute)
+	ctx, cancel := context.WithCancel(t.Context())
 	defer cancel()
 	runner, err := NewRunner(&Move{
 		SourceDSN:    sourceDSN,
@@ -802,6 +799,7 @@ func TestMoveWithVarcharPK(t *testing.T) {
 	})
 	require.NoError(t, err)
 	done := make(chan struct{})
+	errCh := make(chan error, 1)
 	var runErr error
 	var wg sync.WaitGroup
 	t.Cleanup(func() {
@@ -816,23 +814,17 @@ func TestMoveWithVarcharPK(t *testing.T) {
 			t.Error("move runner did not stop during cleanup")
 		}
 	})
-	go func() { runErr = runner.Run(ctx); close(done) }()
+	go func() { runErr = runner.Run(ctx); errCh <- runErr; close(done) }()
 
-	ticker := time.NewTicker(10 * time.Millisecond)
-	defer ticker.Stop()
-	for runner.status.Get() != status.WaitingOnSentinelTable {
-		select {
-		case <-done:
-			t.Fatalf("move exited before sentinel wait: %v", runErr)
-		case <-ctx.Done():
-			t.Fatalf("waiting for sentinel (state %s): %v", runner.status.Get(), ctx.Err())
-		case <-ticker.C:
-		}
-	}
+	waitForMoveStatus(t, runner, status.WaitingOnSentinelTable, errCh)
 
 	var writeCount, errorCount atomic.Int64
 	// Four finite writers keep concurrent FIFO apply coverage while guaranteeing
-	// a quiet source for the final drain. Pacing keeps their binlog load bounded.
+	// a quiet source for the final drain (#834). Unbounded writers can produce
+	// binlog faster than a slow CI reader drains it, preventing catch-up forever.
+	// The finite count guarantees convergence; pacing also limits peak pressure.
+	writeCtx, cancelWrites := context.WithTimeout(ctx, 3*time.Minute)
+	defer cancelWrites()
 	const writers, iterations = 4, 25
 	for range writers {
 		wg.Go(func() {
@@ -840,11 +832,11 @@ func TestMoveWithVarcharPK(t *testing.T) {
 			defer pace.Stop()
 			for range iterations {
 				select {
-				case <-ctx.Done():
+				case <-writeCtx.Done():
 					return
 				case <-pace.C:
 				}
-				if err := varcharPKWriteOne(ctx, sourceDB, srcDB); err != nil {
+				if err := varcharPKWriteOne(writeCtx, sourceDB, srcDB); err != nil {
 					errorCount.Add(1)
 				} else {
 					writeCount.Add(1)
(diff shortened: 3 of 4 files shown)


morgo:
🤖 I checked the formatting concern. Tracker.Get() returns status.State, which implements String() on a value receiver (pkg/status/state.go). Go fmt uses that method for %s; a direct check prints state=waitingOnSentinelTable, without a formatting error. No code change is needed.

aparajon:
🤖 **Adversarial correctness review** — `e367971b` (+60/-36, 1 file)

The diagnosis is right: writers that run until `move.Run()` returns are a feedback loop, not a workload, and *"a slow reader could chase them forever"* is the correct description of why. Making the workload finite and gating the completion assertion on it having actually executed is the right shape, and `require.Equal(t, int64(writers*iterations), writeCount.Load(), "the test must execute its concurrent write workload")` is the assertion that keeps the new determinism from turning into a vacuous pass — without it, a cancelled context would produce a green test that wrote nothing.

The `errorCount == 0` tightening is safe, which is not obvious for a 4-writer concurrent workload: `varcharPKWriteOne` mints a fresh `uuid.New()` per call and only ever touches that row, so no two writers contend on a key and there is no deadlock or duplicate-key path to tolerate. Turning a logged statistic into an assertion is a real improvement over `t.Logf("%d successful writes, %d errors")`.

`DROP TABLE dstDB.<sentinel>` is also the right database — moves create the sentinel on the target (`runner.go:961`, `sentinel.Create(ctx, r.targets[0].DB)`), unlike the migrate path.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | med | `runner_test.go:781`, `:791` | The workload moved entirely after the copy, so the LWW-map-dedup half of the VARCHAR-PK path — the half where the collation hazard lives — is no longer exercised, while the docstring still claims it is |
| 2 | low-med | `runner_test.go:798` | One 3-minute context covers the run, the sentinel wait, the writers and the final drain, so a slow copy makes the final wait fail with a message naming the wrong phase |
| 3 | low-med | `runner_test.go:823` | Third hand-rolled sentinel wait in `pkg/move`, on `!=` against a monotonic state and without the terminal-state guard the shared `pkg/migration` helper has |
| 4 | low | `runner_test.go:800` | The #834 rationale for the pacing is deleted, and the uncapped-writer shape it warns about is still live in the same file |

---

### 1 — the half of the path the test exists for is now uncovered (med)

The docstring above this test — unchanged by the PR — states the split:

> For non-memory-comparable PKs the `bufferedMap` subscription uses **LWW map dedup during the copy phase** and **FIFO queue post-copy**. … the map's hash equality ("A" ≠ "a") does not match MySQL's row identity ("A" = "a" under a CI collation).

So there are two code paths for a VARCHAR PK, divided exactly at the copy boundary. The old arrangement started the writers, slept 100ms, then ran the move, so writes spanned both. The new one waits for `status.WaitingOnSentinelTable` — after copy *and* after the initial checksum — before writing anything, and the new comments say so outright: *"once the sentinel holds the move in FIFO replay mode"*, *"when VARCHAR primary keys use FIFO replay."*

That means the map-dedup path is no longer exercised by this test, and it is the riskier of the two: FIFO replay preserves binlog order and is correct for a CI collation by construction, while the map path is precisely where hash equality diverges from MySQL row identity. The lead docstring still says *"The move runs under concurrent writes to exercise the binlog replay path"*, which now overstates what the body does.

`TestMoveWithConcurrentWrites` doesn't backfill it. Its `xfers` table is `id INT NOT NULL PRIMARY KEY AUTO_INCREMENT` — memory-comparable, so it never takes the FIFO path and never meets the collation hazard at all. After this PR, nothing in `pkg/move` writes concurrently during the copy of a non-memory-comparable PK.

**Being fair to the other side:** the copy-phase coverage was already weaker than it looked. The same docstring notes *"The post-cutover checksum keeps the optimization honest by repairing any divergence"*, and the test asserts only `runErr == nil` plus matching row counts — so a map-path divergence would have been silently repaired and the old test would have passed anyway. That makes this a smaller loss than it first appears, and it changes what I'd ask for rather than removing the ask:

- Either restore a short **finite** burst of writes before the sentinel wait (a handful of iterations; unpaced is fine when it's bounded) so both halves are touched — and, if the recopy signal is reachable from the test (`checksum/continuous.go:389`, `:556` already track "pass contained recopies"), assert it stayed zero, which is what would make copy-phase coverage actually load-bearing rather than decorative;
- or narrow the docstring to say this test covers FIFO replay only, and note that map-phase divergence is caught by the checksum rather than by a test.

Either is fine. Leaving the docstring claiming coverage the body dropped is the thing to avoid, because it is what the next person will read before deciding they don't need to add that case.

### 2 — one deadline for four phases, so the failure message can name the wrong one (low-med)

```go
	ctx, cancel := context.WithTimeout(t.Context(), 3*time.Minute)
	…
	go func() { runErr = runner.Run(ctx); close(done) }()
```

That single `ctx` is the run's context, the sentinel-wait budget, the writers' cancellation signal, and the final drain's budget. Three minutes is the right total — it matches `pkg/migration`'s `waitForStatus`, whose docstring justifies it in detail (a checksum attempt can legitimately spend `change.DefaultTimeout` on catch-up plus 30s on the table lock, retried up to 3 times) — but it isn't three minutes *per phase*, and the assertions are written as if it were:

```go
	case <-ctx.Done():
		t.Fatalf("move did not finish after finite writes (state %s): %v", runner.status.Get(), ctx.Err())
```

A copy that takes 2m50s leaves the final drain ten seconds and then reports "move did not finish after finite writes", pointing triage at the drain when the copy was the problem. The `waiting for sentinel` branch has the same exposure in reverse. `pkg/migration` handles this by giving the wait its own child context (`waitForStatus` derives `context.WithTimeout(t.Context(), 3*time.Minute)` independently of the run's), which keeps each phase's failure message truthful. Splitting the sentinel wait and the final drain onto their own child contexts is a couple of lines and makes both messages mean what they say — worth it in a PR whose purpose is to stop this test failing for reasons that aren't the code under test.

### 3 — the third copy of a wait loop that already has a hardened shared version (low-med)

`pkg/move` now hand-rolls this in three places — `move_sharded_test.go:414`, `sentinel_test.go:72`, and this PR's `runner_test.go:823` — while `pkg/migration` has one helper (`waitForStatus` / `awaitTestStatus`, `helpers_test.go:50`) used at six-plus sites. This is the change that crosses the threshold where extracting is clearly cheaper than a fourth copy.

The shared version isn't just deduplication; it carries two guards this loop lacks:

```go
		current := m.status.Get()
		if current >= status.Close {
			return fmt.Errorf("runner entered terminal state %s while waiting for %s", current, target)
		}
		if current >= target {
			return nil
		}
```

`>= target` rather than `!=`, and an explicit terminal-state check. This loop's `runner.status.Get() != status.WaitingOnSentinelTable` is an exact-equality poll every 10ms against a monotonically advancing state machine: if the runner ever entered and left that state between two samples, the loop would spin to the deadline and report "waiting for sentinel". It is safe here — `runner.go:1291` holds the state inside `r.status.Do(status.WaitingOnSentinelTable, …)` for the whole wait, and the sentinel isn't dropped until afterwards — but it's safe by circumstance rather than by construction, and the `>=` form costs nothing. The `<-done` case this loop does have is the one thing `sentinel_test.go`'s copy lacks, so this version is the better starting point for the extraction.

### 4 — the pacing survived, its reason didn't (low)

The deleted comment was the only record of a specific failure mode:

> Without this, an uncapped tight loop generates binlog faster than the reader can drain it on slow CI runners — the source position outruns the buffered position indefinitely and `Flush()`'s BlockWait loop never converges below `binlogTrivialThreshold`, eventually tripping the 10-minute test timeout (issue #834).

`grep -rn 834 pkg/move/*.go` now returns nothing, and the replacement is *"Pacing keeps their binlog load bounded"* — which states the mechanism but not the consequence, so the next reader has no way to know that removing the ticker reintroduces a ten-minute hang rather than making the test faster. With the workload now finite at 100 writes the pacing may genuinely no longer be load-bearing, and that would be worth saying explicitly instead; either way the PR is making a judgement here that the comment no longer records.

It matters more than usual because the shape #834 describes is still live in this file: `concurrentWriteThread` (`:173`) is a completely unpaced `for { select { …default: doOneWriteLoop() } }` loop, run by four goroutines for the entire duration of `move.Run()` in `TestMoveWithConcurrentWrites` — and `doOneWriteLoop` wraps each iteration in an explicit transaction, so it produces more binlog per iteration than `varcharPKWriteOne` does. So the same "chase them forever" hazard this PR fixes here is unaddressed 650 lines up, now with its explanation deleted. Not this PR's job to fix that test, but it is the reason to keep the explanation somewhere.

---

<details>
<summary><b>Verified</b> — the concurrency, and three attacks that dissolved</summary>

⭐ **Attack that dissolved: `errorCount == 0` is a new way to flake.** Turning a tolerated statistic into a strict assertion across four concurrent writers looked like trading one flake for another — a deadlock (1213) or lock timeout between writers would fail the test outright. It can't happen: `varcharPKWriteOne` generates a fresh UUID per call and its INSERT/UPDATE/DELETE all target that one row, so the writers share no keys. The old `errorCount` tolerance existed for writes racing cutover, and the new arrangement completes all writes *before* the sentinel is dropped, so that window is gone too. The assertion is sound.

⭐ **Attack that dissolved: the sentinel drop targets the wrong database.** `TestCreateSentinelTableIdempotent` creates it in `srcDB` via `r.sources[0].db`, which would have made `DROP TABLE dstDB.<sentinel>` a hard failure. Moves differ from that path: `runner.go:961` creates the sentinel on `r.targets[0].DB`, and `runner.go:1944` logs it as `targets[0].Config.DBName`. `sentinel_test.go:84` drops it from the target database too. Correct as written.

⭐ **Attack that dissolved: `runErr` is read without synchronisation.** It's written by the run goroutine and read at three sites, all of which are reached only after `<-done`, and `close(done)` happens after the assignment — so every read has a happens-before edge. The `t.Cleanup` read is guarded by the same `select` on `done`.

**Cleanup ordering is right.** `t.Cleanup` is registered before the goroutine is launched, so it cannot miss it, and it cancels, waits for the writers, then waits for the run with its own 30s bound before `runner.Close()` — with `t.Error` rather than `t.Fatal`, correct inside a cleanup. Function-level `defer cancel()` and `defer utils.CloseAndLog(sourceDB)` run before cleanups, so the cleanup's `cancel()` is redundant but harmless, and the writers are already finished by then in every path where `errorCount` is asserted.

**The pacing rate is preserved.** Four writers on a 10ms ticker for 25 iterations is the same ~400 ops/sec the removed `time.Sleep(10 * time.Millisecond)` produced, so the change is finite-vs-unbounded rather than a rate change. Switching from `sleep`-at-the-end to `select`-on-ticker also makes each iteration cancellable, which the old loop only was between iterations.

**Local:** `go build ./...` clean, `go vet ./pkg/move/` clean, `gofmt -l pkg/move/` empty, `golangci-lint run ./pkg/move/...` → **0 issues**. CI 18/18 green. This test needs a live MySQL source and target, so it was read rather than run here.

</details>

aparajon (approved):
🤖 Approving — finite writers gated on `writeCount == writers*iterations` is the right shape, and `errorCount == 0` is safe because `varcharPKWriteOne` mints a fresh UUID per call, so no two writers share a key.

Main finding (med): the workload now runs entirely after `WaitingOnSentinelTable`, so the **LWW-map-dedup-during-copy** half of the VARCHAR-PK path is no longer exercised — the half where hash equality diverges from CI-collation row identity — while the unchanged docstring still claims it is. `TestMoveWithConcurrentWrites` doesn't backfill it (INT AUTO_INCREMENT PK). In fairness the checksum would have repaired a map-path divergence silently anyway, so the old coverage was decorative; the ask is either a short finite pre-sentinel burst plus a no-recopy assertion, or narrow the docstring.

Also: one 3-minute context spans run + sentinel wait + writers + final drain, so a slow copy makes the drain fail with a message naming the wrong phase; third hand-rolled sentinel wait in `pkg/move` where `pkg/migration` has a hardened shared helper; and the #834 rationale for the pacing is deleted while the uncapped-writer shape it warns about is still live in `concurrentWriteThread`.

morgo:
🤖 Addressed all four findings. I chose the explicit FIFO-only scope: the test docstring now states that it does not cover concurrent map-mode copy and identifies the checksum as the correctness gate for that phase. It no longer claims coverage it cannot guarantee.

Setup/copy/checksum, writers, and final drain now have independent three-minute budgets, with the runner still cancelled and joined during cleanup. All three move sentinel waits use one helper with early-exit, terminal-state, and >= target handling; focused tests cover these guards and cancellation. The #834 explanation is restored: finite writers guarantee a quiet source, while pacing limits peak binlog pressure. The separate unbounded integer-PK writer test remains outside this PR.

All affected tests passed three race-enabled repetitions, the full move package passed under -race, and lint is clean.

Issue #1161, opened 2026-08-24
flaky test: TestForceExecRawVerb — post-kill retry re-runs with no re-armed kill timer and only 1s of lock_wait_timeout


morgo:
🤖 _Filed by Morgan's AI agent from a scheduled review of `main` CI failures._

## Symptom

`TestForceExecRawVerb` failed on `main` with a lock wait timeout on the **retry** that the force-kill path is supposed to make succeed.

- Run: https://github.com/block/spirit/actions/runs/31992526460
- Job: MySQL 8.0.28 (Aurora version, no GTID) /w docker-compose
- Commit: `c15df3f0` ("migrate: remove --table and --alter in favor of --statement (#1149)")

```
--- FAIL: TestForceExecRawVerb (2.06s)
    dbconn_test.go:419:
        Error Trace:	/app/pkg/dbconn/dbconn_test.go:419
        Error:      	Received unexpected error:
                    	Error 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
        Test:       	TestForceExecRawVerb
[mysql] 2026/08/17 03:54:24 packets.go:58 unexpected EOF
```

## Timeline from the run log

```
03:54:23.126  INFO  found locking transaction pid=2218 lockType=SHARED_READ objectName=forceexecraw_percent runningTime=924.74 ms
03:54:23.126  INFO  found locking transactions count=1 pids=[2218]
03:54:23.126  WARN  killing locking transaction pid=2218
03:54:23.211  WARN  retrying statement after lock wait timeout because force-kill timer fired
03:54:24.214  FAIL  Error 1205 (HY000): Lock wait timeout exceeded
```

The killer did its job — it found the single MDL blocker and killed it. The retry then ran for **exactly one second** (the configured `lock_wait_timeout`) and still could not get the metadata lock, ~1.09s after the `KILL` was issued.

## Likely cause

`ForceExec` retries **once**, on the same connection, and does **not** re-arm the kill timer for the retry:

https://github.com/block/spirit/blob/main/pkg/dbconn/dbconn.go#L341-L367

```go
duration := forceKillGracePeriod(dbConfig.LockWaitTimeout)
...
timer := time.AfterFunc(duration, func() { ... KillLockingTransactions(...) })
_, err = trx.ExecContext(ctx, stmt)
...
if shouldRetryForceExecAfterKill(err, killTimerFired.Load()) {
    logger.Warn("retrying statement after lock wait timeout because force-kill timer fired", "error", err)
[...]

PR #1205, opened 2026-09-05
test: preserve embedded TLS trust bundle across certificate validation

morgo:
TestValidCertificateBundle assigns pem.Decode's remainder back into the embedded rdsGlobalBundle. This empties shared production TLS input for later consumers in the same test process, including NewTLSConfig and GetEmbeddedRDSBundle. RDS TLS initialization can consequently depend on test order; repeated test runs also fail with “No certificates found in bundle”.

Walk a local slice instead. pem.Decode does not mutate its input bytes, so a local slice header preserves the global bundle without copying it. Existing certificate parsing and nonempty assertions remain.

Discovered while repeating dbconn tests for #1161; independent of the force-kill fix.

Validation: the certificate test passed 20 repetitions under -race. The reviewer independently verified preserved CA subjects and embedded bundle length after the test. golangci-lint reports 0 issues.

### DIFF
--- a/pkg/dbconn/conn_test.go
+++ b/pkg/dbconn/conn_test.go
@@ -244,8 +244,9 @@ func TestValidCertificateBundle(t *testing.T) {
 	// parse certificate bundle
 	var block *pem.Block
 	foundCertificates := false
+	remaining := rdsGlobalBundle
 	for {
-		block, rdsGlobalBundle = pem.Decode(rdsGlobalBundle)
+		block, remaining = pem.Decode(remaining)
 		if block == nil {
 			break
 		}


aparajon:
🤖 **Adversarial correctness review** — `064845b2` (+2/-1, 1 file)

Correct, and the diff is smaller than the bug. `pem.Decode` returns the unconsumed remainder, so `block, rdsGlobalBundle = pem.Decode(rdsGlobalBundle)` walked the **package-level embedded var** to `nil` — and that var is not test-local state. It is read by production code in three places: `NewTLSConfig` (`conn.go:79`), `initCustomTLS`'s fallback (`:189`), and the exported `GetEmbeddedRDSBundle` (`:166`).

I measured it rather than inferring it. A probe declared after `conn_test.go`'s tests, on `origin/main`:

| | `NewTLSConfig().RootCAs.Subjects()` | `GetEmbeddedRDSBundle()` |
|---|---|---|
| probe alone (baseline) | 118 | 178,779 bytes |
| **after `TestValidCertificateBundle`, pre-fix** | **0** | **0 bytes** |
| after `TestValidCertificateBundle`, post-fix | 118 | 178,779 bytes |

So the pre-fix test zeroed the embedded RDS CA bundle for the rest of the process, and `NewTLSConfig` then handed out `&tls.Config{RootCAs: <empty pool>}` — a config that verifies against no roots at all. `-count=2` fails on `origin/main` and passes on this branch, which is the visible symptom, but the ordering exposure inside a *single* run is the more interesting half: `initRDSTLS` is `sync.Once`-guarded, so whether a process ends up with an empty registered TLS config depends entirely on whether anything initialized RDS TLS before this test or after it.

`remaining := rdsGlobalBundle` is sufficient and doesn't need `bytes.Clone` — `pem.Decode` only returns subslices of its input and never writes into it, so copying the slice header is enough to keep the global's length and backing array intact. The loop body is otherwise untouched.

No findings. Two notes:

**The title undersells it.** "preserve shared certificate input across runs" reads as a `-count=2` hygiene fix, and #1202's summary called it "an unrelated pre-existing `TestValidCertificateBundle` mutation of its global input." What it actually was is a test destroying production TLS state in-process, silently, with the blast radius decided by test ordering. Worth saying so in the commit message — the next person to see a mystifying certificate-verification failure in this package will search for exactly that sentence.

**The silence is the part still worth fixing.** `caCertPool.AppendCertsFromPEM(rdsGlobalBundle)` discards its `bool` at both `:79` and `:86`. That return value is the one signal that would have turned "the bundle is empty" into an immediate, legible failure instead of a TLS config that looks fine and trusts nothing. Out of scope for a 2-line test fix, but it is the reason this bug could exist at all, and a `if !caCertPool.AppendCertsFromPEM(...)` guard in `NewTLSConfig` / `NewCustomTLSConfig` is cheap insurance against the next version of it.

<details>

`grep -rn rdsGlobalBundle pkg/` after the change shows exactly one test reference (`conn_test.go:247`, the new local) and the three production reads — no other test touches it, and nothing else in the package reassigns it.

**Local:** `go build ./...` clean, `gofmt -l pkg/dbconn/` empty, `golangci-lint run ./pkg/dbconn/...` → 0 issues. `go test ./pkg/dbconn/ -run TestValidCertificateBundle -count=2` passes on this branch and fails on `origin/main` with *"No certificates found in bundle"*. CI 18/18 green.

</details>

aparajon (approved):
🤖 Approving — correct, and the bug was bigger than the diff. Measured on `origin/main`: after `TestValidCertificateBundle` runs, `NewTLSConfig().RootCAs.Subjects()` goes 118 → **0** and `GetEmbeddedRDSBundle()` goes 178,779 → **0 bytes**, because `pem.Decode`'s remainder was being assigned back to the `//go:embed` package var that production TLS config reads. Silent, because `AppendCertsFromPEM`'s bool is discarded.

No findings. Two notes in the comment: the title/commit message undersells this (it reads as `-count=2` hygiene, but the real exposure is production TLS state in a single run, ordered by `sync.Once`), and checking `AppendCertsFromPEM` in `NewTLSConfig`/`NewCustomTLSConfig` is the guard that would have made it loud.

morgo:
🤖 Thanks for verifying the TLS-state impact. I updated the title and description to make the in-process, test-order exposure explicit. The existing commit is preserved rather than rewriting reviewed history. No code changes were needed for this review. Checking AppendCertsFromPEM return values would change production error handling and remains outside this focused test fix; I have not opened a new issue.

PR #1206, opened 2026-09-05
refactor(datasync): require a change stream

morgo:

> If you are offering a new feature or fixing anything, we would like to know beforehand in Issues,
> and potentially we will be able to point development in a particular direction.

## Summary

- remove the `--copy-only` CLI flag and `Sync.CopyOnly` configuration
- make datasync always construct, subscribe to, start, flush, and resume a change stream
- remove the duplicate copy-only checksum path and its integration coverage
- warn when a resume checkpoint has copy progress but no stream position, and document how the continuous checksum repairs the resulting legacy-checkpoint gap

Sources that cannot grant the built-in MySQL replication privileges now need a programmatically injected `change.Source`; the CLI no longer supports a no-stream snapshot mode.

The old end-to-end schema-reconciliation tail intentionally cannot be retained: it used copy-only mode to avoid replaying the historical source DDL. The strict schema gate still has explicit coverage that a target matching the suggested reconciliation ALTER is accepted.

## Testing

- `go test ./...`
- `golangci-lint run`
- `go run ./cmd/spirit sync --help`

### DIFF
--- a/docs/sync.md
+++ b/docs/sync.md
@@ -66,7 +66,6 @@ source privileges depend on the change feed:
 - [write-threads](#write-threads)
 - [flush-interval](#flush-interval)
 - [defer-secondary-indexes](#defer-secondary-indexes)
-- [copy-only](#copy-only)
 - [force](#force)
 
 ### source-dsn
@@ -125,11 +124,10 @@ the replication-latency vs. batching trade-off.
 
 When set to `true`, the target tables are created **without their regular
 secondary indexes**, and the indexes are added back in a single `ALTER` per
-table once the initial copy has completed (before the continuous phase begins,
-or before the post-copy checksum in `--copy-only` mode). Bulk-loading an
-index-free table is faster and lighter on temporary space; only regular
-secondary indexes are deferred — `PRIMARY`, `UNIQUE`, `FULLTEXT` and `SPATIAL`
-indexes are kept on the initial `CREATE`. This mirrors
+table once the initial copy has completed, before the continuous phase begins.
+Bulk-loading an index-free table is faster and lighter on temporary space; only
+regular secondary indexes are deferred — `PRIMARY`, `UNIQUE`, `FULLTEXT` and
+`SPATIAL` indexes are kept on the initial `CREATE`. This mirrors
 [`move --defer-secondary-indexes`](move.md#defer-secondary-indexes).
 
 Use it only when the target is **not yet serving reads**: the tables briefly
@@ -139,18 +137,6 @@ representative). Adding the indexes also needs enough temporary space on the
 target to build them. The restore is resume-safe and idempotent — a re-run
 (even without the flag) detects any indexes still missing on the target and
 adds them, so an interrupted run finishes the job on the next start.
-
-### copy-only
-
-- Type: Boolean
-- Default value: `false`
-
-Run only the initial copy and then exit — no change capture, no continuous
-replication. Use it for a one-shot snapshot, or when the source cannot provide
-a change feed (e.g. a managed Vitess without binlog/VStream access, or a
-replica lacking the `REPLICATION` privileges the binlog client needs). A
-checkpoint is still written, so a re-run resumes the copy (or no-ops if it had
-already completed) rather than starting over.
 
 ### force
 
@@ -186,5 +172,4 @@ Sync-specific notes:
 - **Resume keeps the checkpoint's scheme.** A file+offset checkpoint resumes
   on the file+offset client even after GTIDs are enabled on the source, and a
   GTID checkpoint fails with a clear error if the source no longer has GTIDs
-  enabled. A [`--copy-only`](#copy-only) checkpoint records no stream
-  position, so a later full sync picks its scheme fresh from the server.
+  enabled.
--- a/docs/sync.md
+++ b/docs/sync.md
@@ -57,6 +57,10 @@ source privileges depend on the change feed:
   the source does not have GTIDs enabled (the file+offset reader issues
   `FLUSH BINARY LOGS`)
 
+A source that cannot grant the built-in feed privileges must use a
+programmatically injected `change.Source`; the CLI no longer has a mode that
+runs without a change stream.
+
 ## Configuration
 
 - [source-dsn](#source-dsn)
@@ -173,3 +177,8 @@ Sync-specific notes:
   on the file+offset client even after GTIDs are enabled on the source, and a
   GTID checkpoint fails with a clear error if the source no longer has GTIDs
   enabled.
+- **Legacy copy-only checkpoints have no stream position.** When upgrading a
+  target with one of these checkpoints, Sync warns and starts the change stream
+  at the current source head. Changes made after the old checkpoint are not
+  replayed immediately; the continuous checksum finds and repairs that gap as
+  it walks the target.
--- a/pkg/change/README.md
+++ b/pkg/change/README.md
@@ -269,7 +269,6 @@ This is the same mechanism as [issue #746](https://github.com/block/spirit/issue
 |---|---|---|
 | `migrate`, `move` | Mandatory pre-cutover checksum with `FixDifferences=true` | Repaired before cutover. Cost: `differencesFound > 0`, a chunk recopy, and a "checksum found differences" signal that looks alarming |
 | `sync` (continuous) | Continuous checksum + `MySQLRecopier`, *lazy* | Real exposure: the target can serve a missing/stale/phantom row from copy time until a later checksum pass covers that chunk |
-| `sync --copy-only` | n/a | Not applicable — the discard is never enabled (`SetWatermarkOptimization` is skipped) |
 | Library consumers of pkg/copier + pkg/change with no checksum | None | Silent data loss |
 
 This is the same reliance already accepted knowingly for collation-imprecise key comparisons ([issue #479](https://github.com/block/spirit/issues/479), "checksum will fix any discrepancies") — except the visibility window affects every key type, not just collated strings.
--- a/pkg/checksum/continuous.go
+++ b/pkg/checksum/continuous.go
@@ -163,15 +163,15 @@ type ContinuousCheckerConfig struct {
 	Recopier Recopier
 
 	// DivergenceIsFatal selects the policy for a confirmed stable divergence,
-	// making explicit the replication-backed vs. copy-only distinction rather
-	// than inferring it from Recopier presence:
+	// making explicit whether the caller should abort or heal rather than
+	// inferring it from Recopier presence:
 	//   - true  (migration's cutover gate): the target is kept in sync by
 	//     replication, so a confirmed difference means something is genuinely
 	//     wrong. Run returns ErrPermanentDivergence and the caller aborts. No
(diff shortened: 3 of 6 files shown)


aparajon:
🤖 **Adversarial correctness review** — `317b7a82` (+75/-287, 6 files)

Good deletion. Removing a mode is the kind of change that usually leaves residue, and this one doesn't: `grep -rni 'copyonly|copy-only'` across the branch returns only unrelated prose in `pkg/migration` and `pkg/checksum` test comments, so the flag, the config field, the branch guards, the docs section, the anchor in the TOC, the `pkg/change/README.md` table row, and the two `ContinuousCheckerConfig` comment references all went together. The comment rewrites are accurate rather than mechanical — `DivergenceIsFatal`'s "abort or heal" framing is a better description of the real distinction than the replication-backed-vs-copy-only one it replaces, and `executeWork`'s "may be nil for library callers" is still true after the CLI path stopped producing a nil feed.

`runCopyOnlyChecksum` really was a duplicate, which is the load-bearing claim of the diff and the one worth checking rather than accepting. `runContinuous` retains every element of the deleted function's shutdown contract — the `checksumDone` goroutine, the `cancelChecksum(); <-checksumDone` ordering before the checkpoint write, the `r.fatal()` check, the final `dumpCheckpoint` under `WithoutCancel` + `shutdownCheckpointTimeout`, and the `checksumErr != nil && !errors.Is(checksumErr, context.Canceled)` surfacing that keeps a real failure from being masked by a clean ctx-cancel. Nothing in that contract existed only in the copy-only copy.

Two things make this materially lower-risk than a flag removal usually is, and both are worth stating because they set the severity ceiling for everything below: `sync` is `[EXPERIMENTAL]` at `main.go:25`, and the removal fails closed — `spirit sync --copy-only` now exits 80 with `spirit: error: unknown flag --copy-only`, rather than silently ignoring it.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low-med | `runner.go:1201` | Checkpoints already written by copy-only runs record no stream position; resuming one now starts the feed at *now*, and that branch logs nothing |
| 2 | low-med | `sync_test.go:988` | The deleted tail was the only end-to-end proof that the reconcile-then-resume remedy — which production code still recommends by name — actually completes |
| 3 | low | PR summary | The retired escape hatch had two documented reasons; the stated evidence covers one of them |

---

### 1 — the checkpoints outlive the mode, and they carry no position (low-med)

`dumpCheckpoint` on `main` writes the position defensively, precisely because copy-only had no client:

```go
	var pos string
	if repl != nil {
		pos = repl.Position()
	}
```

So every checkpoint a copy-only run ever wrote has `Position=""`. The mode is gone; those rows are not. Feed one to the new code and `startResume` takes the fallback branch:

```go
	} else if err := r.replClient.Start(streamCtx); err != nil {
		// No saved position (prior attempt failed before checkpointing):
		// start the feed fresh; changes apply with the optimization off.
```

For the case that comment describes — an attempt that died before its first checkpoint — starting fresh is right, because the copier still has the whole table ahead of it. For a *completed* copy-only checkpoint it isn't: the copier opens at a finished watermark and no-ops, the feed starts at the current position, and every source write between the two runs falls in the gap. Nothing replays it. The target is silently short those rows while the runner reports itself in continuous sync.

The saving grace is real: `DivergenceIsFatal: false` plus the `MySQLRecopier` means a later checksum pass finds and heals each diverged chunk, so this converges rather than corrupting permanently. That's the same lazy-repair window `pkg/change/README.md` already documents for continuous sync, and it's why this is low-med rather than high. But the window is a full checksum pass over the table, and — unlike the documented case — nothing tells the operator they're in it.

Which is the actual ask: **log it.** The branch that starts a feed at the current position during a *resume* is now reachable in a way it wasn't before, and it is entirely silent on success — no `Warn`, no note that the target may be behind until a checksum pass lands. One line naming the checkpoint's watermark and the fact that the stream is starting fresh would make this diagnosable instead of invisible, and it's the kind of branch the codebase's own "never take a fallback silently" habit already covers elsewhere in this file.

Worth noting this PR also deleted the one sentence in the docs that described these checkpoints:

> A `--copy-only` checkpoint records no stream position, so a later full sync picks its scheme fresh from the server.

Deleting the flag's documentation is correct. Deleting the description of the artifacts it left behind, in the same PR that makes those artifacts take a different path, is the part that reads as premature — either the log line above replaces it, or a line in the summary should say what happens to an existing copy-only checkpoint on upgrade. Right now nothing does.

### 2 — the remedy is still recommended in production code; the proof that it works is gone (low-med)

The second deletion in `sync_test.go` isn't copy-only coverage, it's coverage that *used* copy-only as a vehicle:

```go
-	// Reconciling the target unblocks the resume. … Resume in copy-only mode: a
-	// continuous resume opens the change feed at the checkpointed position, which
-	// predates the ALTER, so the source's DDL guard would cancel the run before the
-	// copy finishes — that guard is exactly what leaves the resumable checkpoint
-	// this test starts from, and it is not what's under test here.
-	testutils.RunSQL(t, `ALTER TABLE sync_ddl_drift_dest.t1 ADD COLUMN added …`)
-	reconciled := newSync()
-	reconciled.CopyOnly = true
```

That comment is the record of why the coverage can't simply be ported: in continuous mode the guard fires before the copy finishes, so copy-only was the only mode in which "reconcile the target, then resume" could be driven to completion at all. With it deleted, `TestSyncResumeSourceSchemaChanged` now stops at

```go
	require.Contains(t, runErr.Error(), "ADD COLUMN `added`",
		"the error should carry the ALTER that reconciles the target")
```

— it asserts that spirit hands the operator a remedy, and no longer that the remedy works. The remedy is not deprecated along with the flag: `verifyExistingTargetTable` still exists, still fails deliberately rather than repairing (`runner.go:968-1013`), and its whole design rests on the operator running the named ALTER and re-running. (The leftover blank line before the closing brace is the cosmetic tell that the tail was cut; `gofmt` and lint are both happy with it.)

The comparison half is still covered — `TestSyncVerifyExistingTargetTableRequiresExactSchema` (`:885`) calls `verifyExistingTargetTable` directly with strings and has an `identical` pass case, so "a reconciled target satisfies the gate" is pinned. What's lost is everything downstream of the gate: that a resume against an ALTERed target actually opens the chunker at the watermark, copies the remainder, and restores indexes. So the honest scope of the loss is one integration-level assertion, not the whole behaviour — which is why this is low-med.

Two ways to close it, and one is nearly free:

- Add the ALTER-then-pass case as a row on the existing `verifyExistingTargetTable` table if it isn't already implied by `identical`, and say in the summary that end-to-end coverage of the remedy is intentionally dropped because no remaining mode can express it. That's a defensible call — just make it explicitly rather than by deletion.
- Or, if the remedy is meant to stay supported end to end, the guard's timing is the thing to attack: a continuous resume that re-reads the source schema *after* the operator's reconciliation, before opening the feed, would make the remedy testable in the only mode that now exists. Bigger than this PR, but this PR is what makes it the only route.

### 3 — one of the two documented reasons is answered (low)

The deleted docs said:

> Use it for a one-shot snapshot, or when the source cannot provide a change feed (e.g. a managed Vitess without binlog/VStream access, or a replica lacking the `REPLICATION` privileges the binlog client needs).

The summary's rationale — *"used when I couldn't get vstream connectivity working; it all works now"* — retires the first environment convincingly. The second is a different shape: a replica without `REPLICATION SLAVE`/`REPLICATION CLIENT` isn't a connectivity problem that got fixed, it's a permission the source may not be able to grant. After this PR the CLI has no path for it — `NewAutoClient` fails in `setup` and the run never starts.

The docs are coherent about this, which matters: `Requirements` now states the privileges unconditionally for the built-in feed, and `Sync.Source` injection remains for library callers who can supply their own stream. So the capability is narrowed for CLI users rather than removed from the package, and on an `[EXPERIMENTAL]` command that is a reasonable call to make. The ask is one sentence in the commit message saying you're making it — that a privilege-limited source now needs an injected `change.Source` — because a reader comparing the deleted docs against the stated evidence can't currently tell whether the second case was considered or overlooked.

---

<details>
<summary><b>Verified</b> — nil-safety, deletion completeness, and an attack that dissolved</summary>

⭐ **Attack that dissolved: `startBackgroundRoutines` dereferences a nil client.** The removed guard was `if !r.sync.CopyOnly { r.replClient.StartPeriodicFlush(...) }`, so if any path could reach it before `setup` wired a client, the unconditional call would panic on the first flush tick. It can't: `Run` calls `r.setup(ctx)` and returns on error before `startBackgroundRoutines`, and `setup` either adopts the injected `r.sync.Source` or returns `NewAutoClient`'s error — there is no success path leaving `replClient` nil. The same reasoning covers the other three guards this PR dropped (`runCopy`, `buildChunkers`, `startResume`), all of which run strictly after `setup`.

**The deletion is complete in both directions.** No dangling `CopyOnly` reference remains, and nothing that only the copy-only path used is left orphaned — `runCopyOnlyChecksum` was its sole caller of nothing else, and `dumpCheckpoint`'s `if repl != nil` nil-guard is now dead-but-harmless defensive code rather than a live branch (worth leaving; library callers construct `Runner` directly).

**The `--copy-only` removal fails closed.** Built the branch and ran it: exit 80, `spirit: error: unknown flag --copy-only`, full usage printed. No silent acceptance, and `sync --help` no longer advertises it. A pinned script gets an immediate, legible failure rather than an unexpectedly continuous sync — which is the right direction for this particular flag, since silently upgrading a snapshot job to continuous replication would be the worse outcome.

**`docs/sync.md` is internally consistent after the edit.** The TOC entry, the section, and the cross-reference in the resume-scheme bullet were all removed; the `defer-secondary-indexes` rewrite correctly drops the "or before the post-copy checksum" clause rather than leaving a dangling alternative; and the `Requirements` privileges block already carried the `(built-in feed)` qualifier, so it needed no change to stay accurate.

**Local:** `go build ./...` clean, `go vet ./pkg/datasync/ ./pkg/checksum/` clean, `gofmt -l` empty, `golangci-lint run ./pkg/datasync/... ./pkg/checksum/...` → **0 issues**. CI 15/15 green. The `pkg/datasync` suite needs a live MySQL source and target, so it was read rather than run here; the CLI check above and the `dumpCheckpoint` comparison against `origin/main` needed no server.

</details>

aparajon (approved):
🤖 Approved — the deletion is complete and `runContinuous` genuinely retained the whole shutdown contract, so `runCopyOnlyChecksum` really was a duplicate. Findings in a separate comment; the one worth a follow-up is that copy-only checkpoints already written in the field record no stream position, and the resume branch they now land on is silent.

morgo:
🤖 Addressed in `274eeec1`:

- positionless resume checkpoints with copy progress now emit a warning that the stream starts at the current source head and the target may remain stale until continuous checksum repair; the legacy copy-only upgrade behavior is documented
- the exact-schema test now explicitly covers a target reconciled with the suggested ALTER; the former end-to-end tail is intentionally not retained because only copy-only mode avoided replaying the historical DDL
- the docs and PR description now state that privilege-limited sources require an injected `change.Source`, since the CLI no longer has a no-stream mode

Full `go test ./...` and `golangci-lint run` pass locally.

PR #1213, opened 2026-09-05
fix(security): resolve code scanning alerts

morgo:
## Summary

Addresses the [open GitHub code-scanning alerts](https://github.com/block/spirit/security/code-scanning).

- Serialize chunk checkpoint data with `encoding/json`, preventing quoted key names or boundary values from producing malformed JSON.
- Propagate checkpoint serialization failures through `GetLowWatermark` instead of panicking in the periodic checkpoint path.
- Reject negative or overflowing SQL escape-buffer growth before allocation.
- Restrict six test workflows to read-only repository permissions.
- Disable `setup-go` caching in three workflows that cross untrusted and privileged execution contexts.
- Add regression coverage for JSON key escaping and allocation overflow.

## Alert disposition

- 13 findings are addressed by this PR: six missing workflow permissions, three cache-poisoning risks, two unsafe JSON-quoting paths, and two allocation-overflow reports.
- Four TLS findings were reviewed separately. `VERIFY_CA` performs explicit x509 chain validation and was marked as a false positive. The `PREFERRED` and `REQUIRED` findings describe intentional, documented MySQL-compatible behavior and were dismissed with rationale.

## Review follow-up

- The quoted-key regression test now round-trips through the real checkpoint resume parser, including both bounds.
- Watermark tuple-key comments now document why JSON quoting must remain injective for out-of-order chunk matching.
- The escape expansion multiplication is checked before it is evaluated; the allocation helper retains its independent overflow backstop.
- Disabling the release cache is the substantive security tradeoff. The lint and parser-regeneration caches were already ineffective because `setup-go` runs before checkout, so those two changes remove warnings at no existing cache-performance cost.

## Testing

```text
golangci-lint run --timeout=2m
go test ./pkg/dbconn/sqlescape ./pkg/table -run 'TestReserveBuffer|TestBoundaryJSONControlChar|TestChunkJSONEscapesKeyNames|TestBinaryChunkJSONRoundTrip|TestBoundary_ValueString|TestChunker.*Watermark' -count=1
git diff --check
```

### DIFF
--- a/.github/workflows/buildandrun-docker.yml
+++ b/.github/workflows/buildandrun-docker.yml
@@ -4,6 +4,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -19,6 +19,7 @@ jobs:
       - uses: actions/setup-go@v7.0.0
         with:
           go-version: "1.26.6"
+          cache: false
       - name: Checkout code
         uses: actions/checkout@v7
       - name: golangci-lint
--- a/.github/workflows/mysql-semisync-docker.yml
+++ b/.github/workflows/mysql-semisync-docker.yml
@@ -4,6 +4,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/mysql8-docker.yml
+++ b/.github/workflows/mysql8-docker.yml
@@ -4,6 +4,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/mysql8.0.42-docker.yml
+++ b/.github/workflows/mysql8.0.42-docker.yml
@@ -4,6 +4,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/mysql84-docker.yml
+++ b/.github/workflows/mysql84-docker.yml
@@ -4,6 +4,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/mysql97-docker.yml
+++ b/.github/workflows/mysql97-docker.yml
@@ -4,6 +4,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/parser-regen.yml
+++ b/.github/workflows/parser-regen.yml
@@ -17,6 +17,7 @@ jobs:
       - uses: actions/setup-go@v7.0.0
         with:
           go-version: "1.26.6"
+          cache: false
       - name: Checkout code
         uses: actions/checkout@v7
       - name: Regenerate parser from grammar
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,7 @@ jobs:
         uses: actions/setup-go@v7.0.0
         with:
           go-version-file: go.mod
+          cache: false
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
--- a/pkg/dbconn/sqlescape/utils.go
+++ b/pkg/dbconn/sqlescape/utils.go
@@ -26,9 +26,20 @@ import (
 )
 
 func reserveBuffer(buf []byte, appendSize int) []byte {
+	maxInt := int(^uint(0) >> 1)
+	if appendSize < 0 || appendSize > maxInt-len(buf) {
+		panic("sqlescape: buffer size overflow")
+	}
 	newSize := len(buf) + appendSize
 	if cap(buf) < newSize {
-		newBuf := make([]byte, len(buf)*2+appendSize)
+		// Preserve the historical doubling growth strategy when it is safe, but
+		// skip the extra capacity when doubling would overflow int. newSize
+		// itself was checked above.
+		newCapacity := newSize
+		if len(buf) <= maxInt-newSize {
+			newCapacity += len(buf)
+		}
+		newBuf := make([]byte, newCapacity)
 		copy(newBuf, buf)
 		buf = newBuf
 	}
--- a/pkg/dbconn/sqlescape/utils.go
+++ b/pkg/dbconn/sqlescape/utils.go
@@ -49,6 +49,10 @@ func reserveBuffer(buf []byte, appendSize int) []byte {
 // escapeBytesBackslash will escape []byte into the buffer, with backslash.
 func escapeBytesBackslash(buf []byte, v []byte) []byte {
 	pos := len(buf)
+	maxInt := int(^uint(0) >> 1)
+	if len(v) > maxInt/2 {
+		panic("sqlescape: buffer size overflow")
+	}
 	buf = reserveBuffer(buf, len(v)*2)
 
 	for _, c := range v {
--- a/pkg/dbconn/sqlescape/utils_test.go
+++ b/pkg/dbconn/sqlescape/utils_test.go
@@ -16,6 +16,7 @@ package sqlescape
 
 import (
 	"encoding/json"
+	"math"
 	"strings"
 	"testing"
 	"time"
@@ -35,6 +36,13 @@ func TestReserveBuffer(t *testing.T) {
 	require.Len(t, res2, 12)
 	require.Equal(t, 15, cap(res2))
 	require.Equal(t, res1, res2[:3])
+
+	require.PanicsWithValue(t, "sqlescape: buffer size overflow", func() {
+		reserveBuffer(make([]byte, 1), math.MaxInt)
+	})
+	require.PanicsWithValue(t, "sqlescape: buffer size overflow", func() {
+		reserveBuffer(nil, -1)
+	})
 }
 
 func TestEscapeBackslash(t *testing.T) {
--- a/pkg/table/chunk.go
+++ b/pkg/table/chunk.go
@@ -91,18 +91,37 @@ func (c *Chunk) String() string {
 }
 
 func (c *Chunk) JSON() string {
-	return fmt.Sprintf(`{"Key":["%s"],"ChunkSize":%d,"LowerBound":%s,"UpperBound":%s}`,
-		strings.Join(c.Key, `","`),
-		c.ChunkSize,
-		c.LowerBound.JSON(),
-		c.UpperBound.JSON(),
-	)
+	out, err := json.Marshal(JSONChunk{
+		Key:        c.Key,
+		ChunkSize:  c.ChunkSize,
+		LowerBound: c.LowerBound.jsonBoundary(),
+		UpperBound: c.UpperBound.jsonBoundary(),
+	})
+	if err != nil {
+		// JSONChunk contains only strings, integers, booleans, and slices, so
+		// encoding it cannot fail. Keep JSON's historical string-only API while
+		// making an impossible future schema regression fail loudly.
+		panic(fmt.Sprintf("could not encode chunk JSON: %v", err))
+	}
+	return string(out)
 }
 
 // JSON encodes a boundary as JSON. The values are represented as strings,
 // to avoid JSON float behavior. See Issue #125
 func (b *Boundary) JSON() string {
-	return fmt.Sprintf(`{"Value": [%s],"Inclusive":%t}`, b.valuesString(), b.Inclusive)
+	out, err := json.Marshal(b.jsonBoundary())
+	if err != nil {
+		panic(fmt.Sprintf("could not encode boundary JSON: %v", err))
+	}
+	return string(out)
+}
+
+func (b *Boundary) jsonBoundary() JSONBoundary {
+	values := make([]string, len(b.Value))
+	for i, value := range b.Value {
+		values[i] = jsonDatumString(value)
+	}
+	return JSONBoundary{Value: values, Inclusive: b.Inclusive}
 }
 
 // comparesTo returns true if the boundaries are the same.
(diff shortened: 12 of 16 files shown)


aparajon:
🤖 **Adversarial correctness review** — `d1b1c55b` (+96/-13, 13 files)

This clears every open code-scanning alert on `main` — I checked the list rather than taking the PR's word for it, and all 13 are addressed: two `go/allocation-size-overflow` in `sqlescape`, two **critical** `go/unsafe-quoting` in `chunk.go`, three `zizmor/cache-poisoning`, six `actions/missing-workflow-permissions`. The branch reports zero. That includes the two `sqlescape` alerts that have been re-attributed onto unrelated PRs by dataflow and turning their CodeQL check red, so this unblocks more than itself.

⭐ **The `chunk.go` alert was a real bug, not a false positive, and it's worth being explicit about which half.** Two things were interpolated into that format string. The boundary *values* were already escaped — `jsonQuoteDatum` has run them through `json.Marshal` for a while — so CodeQL's alert at `:105` was flagging a path that was actually safe, just not one it could see through. The *key names* were not: `strings.Join(c.Key, `","`)` put raw column names into JSON with no escaping at all, and a MySQL identifier may legally contain a `"` or a `\`. A table whose PK column is named that way wrote a syntactically invalid checkpoint, and since a checkpoint that won't parse can't be resumed, the failure was permanent and only reachable by the customer unlucky enough to have that column. Restructuring through `json.Marshal` fixes it at the layer where it can't recur — and I confirmed there is no other hand-built JSON in the tree (`grep 'Sprintf(`{'` is empty), so this closes the class and not just the instance.

⭐ **The refactor is output-preserving where that is load-bearing, which is the thing I most wanted to check.** `jsonQuoteDatum` is a **map key** in `watermarkTracker` (`watermark_tracker.go:110`, `:125`), so a change to its rendering isn't a cosmetic diff — two boundaries that collided on a key would let one chunk silently overwrite another and the watermark advance past uncopied rows. Splitting out `jsonDatumString` and re-wrapping it leaves `jsonQuoteDatum` byte-identical, and giving both the checkpoint path and the map-key path a common root actually makes them *harder* to drift apart than before.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low-med | `chunk.go:104`, `chunker_composite.go:387` | The new panic sits six lines above an existing error return for a strictly-easier marshal, in a function that already returns `error`, on the periodic checkpoint path |
| 2 | low-med | `chunk.go:146`, `:156` | Both comments on the surviving quote path describe a consumer this PR removed; the property that now makes the escaping load-bearing is a different one, and unstated |
| 3 | low | `chunk_test.go` | The new key test stops at the JSON layer, one construction short of proving the thing the alert was actually about — that the checkpoint still resumes |
| 4 | low | `linter.yml:22`, `parser-regen.yml:20`, `release.yml` | The three `cache: false` edits are not equivalent, and two of them are free only because those workflows' cache has never worked — measured |
| 5 | low | `sqlescape/utils.go:30` | The guard checks the sum, but the sole caller passes `len(v)*2`, which wraps one frame earlier |

---

### 1 — the panic and the error return are in the same call stack, six lines apart (low-med)

```go
	if err != nil {
		// JSONChunk contains only strings, integers, booleans, and slices, so
		// encoding it cannot fail. Keep JSON's historical string-only API while
		// making an impossible future schema regression fail loudly.
		panic(fmt.Sprintf("could not encode chunk JSON: %v", err))
	}
```

The reasoning is sound and I agree the branch is unreachable: `JSONChunk` is `[]string` / `uint64` / `bool` all the way down, `json.Marshal` has no error to return for those, and invalid UTF-8 in a string is replaced rather than rejected. So this is about which shape an unreachable branch should take, not about a live crash.

What makes it worth raising is the immediate neighbourhood. `Chunk.JSON()` has two non-test callers, both on the checkpoint path, and one of them is:

```go
	watermark := compositeWatermark{
		ChunkJSON:  t.watermark.JSON(),      // panics on marshal failure
		RowsCopied: atomic.LoadUint64(&t.rowsCopied),
	}
	jsonBytes, err := json.Marshal(watermark)
	if err != nil {
		return "", fmt.Errorf("could not serialize composite watermark: %w", err)  // returns
	}
```

Within one function, the *more* failure-prone marshal panics and the *less* failure-prone one — a struct of one string and one uint64, strictly a subset of what `JSONChunk` can hold — returns an error. Both are equally impossible, and they were given opposite treatments six lines apart. That asymmetry is what a reader will notice first, and it doesn't have a reason behind it.

The error channel is also already there and already load-bearing: `GetLowWatermark() (string, error)` on both chunkers, which already returns `ErrWatermarkNotReady`, and I checked all nine non-test call sites — `pkg/move`, `pkg/migration`, `pkg/checksum` (three), `pkg/datasync` — every one of them checks the error. So an error return would be handled everywhere today, with no new plumbing. A panic instead takes down the host process from a periodic checkpoint tick, and this is a library: the process it lands in belongs to whoever embedded it, and the goroutine it fires on is not one they wrote.

I'll note the honest counterpoint, since I looked for it: `watermark_tracker.go:102` already panics, deliberately, with `// Fatal equivalent`. So panicking isn't foreign to this package. But that one fires on a chunker invariant violation with no coherent way to continue — there is genuinely nothing to return *to*. This one has a caller holding an `error` it is already inspecting. The ask is small: return the error and let `GetLowWatermark` wrap it the way the composite path six lines below already does. If the string-only signature is worth keeping for its own sake, then say *that* in the comment, because it's the actual reason.

### 2 — both comments on the quote path point at a consumer this PR deleted (low-med)

`Boundary.JSON()` no longer routes through `valuesString()`, so `valuesString` now has exactly one consumer left — the watermark map — and its comment still names the old one:

```go
// valuesString renders the boundary values as a comma-separated list of JSON
// string literals (used inside Boundary.JSON's "Value" array).
```

It isn't used inside `Boundary.JSON`'s "Value" array any more. `jsonBoundary()` is. And `jsonQuoteDatum`'s longer comment ends on a justification that has moved out from under it:

> must be JSON-escaped here, or the checkpoint watermark becomes unparseable and resume fails permanently.

After this PR nothing about the checkpoint flows through `jsonQuoteDatum`. The checkpoint goes `Chunk.JSON` → `jsonBoundary` → `jsonDatumString` → `json.Marshal`, and `jsonQuoteDatum`'s output is now consumed only as a Go map key that is never parsed as JSON by anything.

That matters more than ordinary comment rot, because the escaping is still load-bearing — for a *different* reason that nothing states. As a map key, what the escaping buys is **injectivity**. `strings.Join` on a raw rendering would let the two-column boundary `["a", "b"]` and the one-column boundary `["a,b"]` produce the same key `a,b`; one chunk then overwrites the other in `lowerBoundWatermarkMap`, and `drainAlignedChunks` advances the watermark over a chunk that was never copied. The quoting is what keeps the mapping one-to-one.

So the comment currently gives a reason that is obsolete for a requirement that is real, and it is precisely the comment someone reads while thinking "the enclosing encoder escapes now, this inner `json.Marshal` is redundant." The ask is to retarget both: `valuesString` feeds the watermark map, and `jsonQuoteDatum` must stay injective because that map's keys are how out-of-order chunks are matched — dropping the quoting collides distinct boundaries rather than producing bad JSON.

Two smaller things in the same area, worth a line each. `Boundary.JSON()` is now unreferenced outside tests, so the exported method and `jsonBoundary` are parallel paths to the same bytes with only one of them exercised in production — fine to keep as public API, but it now only holds its tests, not the code. And `datum.go:166`'s comment still refers to "jsonQuoteDatum" when describing the checkpoint serialization contract; `jsonDatumString` is the function that contract now belongs to.

### 3 — the key test proves the write, not the resume (low)

```go
	encoded := chunk.JSON()
	require.True(t, json.Valid([]byte(encoded)), ...)
	var decoded JSONChunk
	require.NoError(t, json.Unmarshal([]byte(encoded), &decoded))
	require.Equal(t, chunk.Key, decoded.Key)
```

Good test, and it does kill the bug — I reverted `Chunk.JSON()` to `main`'s `Sprintf` form and it fails, so it is not one of those tests that would pass either way. The gap is that it decodes into `JSONChunk` directly, which is the writer's own struct, so it asserts the encoder is self-consistent. The alert's actual consequence was that **resume** breaks permanently, and the function that does resume is `newChunkFromJSON` — with its own `len(chunk.Key) == 0` validation and its own `jsonStrings2Datums` conversion, neither of which this test reaches.

It's two lines away, and I confirmed it works rather than guessing:

```go
	ti := NewTableInfo(nil, "test", "t1")
	ti.columnsMySQLTps = map[string]string{"quote\"key": "int", "slash\\key": "int"}
	// ...
	restored, err := newChunkFromJSON(ti, chunk.JSON())
	require.NoError(t, err)
	require.Equal(t, chunk.Key, restored.Key)
```

Passes on this branch; the emitted JSON is `{"Key":["quote\"key","slash\\key"],...}` and the restored key comes back byte-identical along with both bounds. `TestBinaryChunkJSONRoundTrip` already establishes that going through `newChunkFromJSON` is the idiom here — it's what makes that test convincing about *values*. Keys deserve the same, especially since keys are the half that was genuinely broken.

### 4 — the three `cache: false` edits don't cost the same, and two are free by accident (low)

`release.yml` is the one where this is a real trade: `contents: write`, checkout first, GoReleaser with `GITHUB_TOKEN`. That is exactly the privileged job `zizmor/cache-poisoning` exists to protect, the cache there was genuinely working, and giving it up to close the alert is the right call.

The other two are free, and the reason they're free is a bug. `linter.yml` and `parser-regen.yml` both run `actions/setup-go` **before** `actions/checkout` — `setup-go` at line 19 / 17, `checkout` at 23 / 21 — so `setup-go`'s cache key, which is derived from `go.mod`, has never had a `go.mod` to hash. It doesn't fail the job, it warns and moves on. From the most recent `linter.yml` run on `main`:

```
##[warning]Restore cache failed: Dependencies file is not found in /home/runner/work/spirit/spirit.
Supported file pattern: go.mod
```

So `cache: false` on those two removes nothing, because nothing was there. Worth saying out loud in the PR or in a comment on the line, for two reasons. It means the alert on those two files is being closed at zero cost, which is a better story than the diff tells. And it means the ordering bug is now invisible: with `cache: false` the warning stops, so the next person to wonder why lint takes as long as it does has lost the breadcrumb. Reordering the two steps is a one-line fix and can be done independently — though note it's only worth doing *with* the `cache: false` kept if the goal is correctness of the workflow rather than speed, since restoring the cache is what the alert objects to.

Both files also already carried workflow-level `permissions: contents: read` before this PR; zizmor flags them for the `push: tags: v*` trigger, not for permissions. Naming that in the PR would save the next reader the same lookup.

### 5 — the overflow the guard names happens one frame up (low)

```go
	maxInt := int(^uint(0) >> 1)
	if appendSize < 0 || appendSize > maxInt-len(buf) {
		panic("sqlescape: buffer size overflow")
	}
```

`reserveBuffer` has exactly one caller:

```go
	buf = reserveBuffer(buf, len(v)*2)
```

`len(v)*2` is unchecked and wraps at `maxInt/2`, which is well before the sum `len(buf)+appendSize` can reach `maxInt`. So on the only path that exists, the multiplication is what overflows first, and by the time `reserveBuffer` sees it the value is already negative — caught by the `appendSize < 0` arm, which is why the guard works, but it is diagnosing a wrap that happened in its caller. The message "buffer size overflow" is then true about the system and slightly misleading about the frame.

Unreachable on 64-bit either way, so this is about where the invariant is stated rather than about a live defect. If the point is to make the arithmetic auditable — which is the point, since CodeQL can't reach a conclusion the code doesn't state — then `len(v) > maxInt/2` at the call site is where the check belongs, and the callee's guard becomes the backstop it reads as.

Related: I verified the growth strategy survived, since that was the easy thing to get wrong here. `newCapacity = newSize + len(buf) = 2*len(buf) + appendSize`, identical to the old `make([]byte, len(buf)*2+appendSize)`, and `require.Equal(t, 15, cap(res2))` in the existing test pins it.

---

<details>
<summary><b>Verified</b> — round-trip compatibility, the map keys, mutations, and two attacks that dissolved</summary>

⭐ **Attack that dissolved: the refactor silently changing the watermark map keys.** `valuesString()` is a map key in `watermarkTracker`, so any change to datum rendering would have relocated live watermark entries — chunks would stop matching, `drainAlignedChunks` would stall, and the checkpoint would fall behind the actual copy progress with no error anywhere. Splitting `jsonQuoteDatum` into `jsonDatumString` + a `json.Marshal` wrapper is exactly output-preserving: same branches, same order, same `%#x` and `x''` special cases, and the wrapper reapplies the identical `json.Marshal(s)`. `datum_test.go` pins two of the tricky ones (`"x''"` and `"0x00"`) directly on `jsonQuoteDatum`, so the escaping can't be dropped later without a failure.

⭐ **Attack that dissolved: invalid UTF-8 in a non-binary PK reaching `json.Marshal`.** `json.Marshal` doesn't reject invalid UTF-8, it substitutes U+FFFD — so a latin1 VARCHAR PK holding a stray high byte would be silently corrupted into the checkpoint and restore as a different value, shifting a chunk boundary and skipping rows. The comment's "(valid-UTF8)" parenthetical reads like an assumption, but it's enforced: `Datum.IsBinaryString()` returns true for *any* string failing `utf8.ValidString` regardless of declared type, so such a value takes the hex branch before it ever reaches the encoder. Unchanged by this PR, and the guarantee holds on both the old and new paths.

**Checkpoint compatibility is intact in both directions.** The only output changes are whitespace (`{"Value": [` → `{"Value":[`) and the Key escaping fix. Both read paths are `json.Unmarshal`, which is whitespace-insensitive, so a checkpoint written by `main` resumes on this branch and one written here resumes on `main`. `newChunkFromJSON`'s `len(chunk.Key) == 0` check means a nil Key marshalling to `null` would fail loudly rather than silently producing an unkeyed chunk.

**Mutation testing — every new assertion earns its place.** All four mutations were killed:

| Mutation | Result |
|---|---|
| revert `Chunk.JSON()` to `main`'s `Sprintf` form | **caught** by `TestChunkJSONEscapesKeyNames` |
| drop the `appendSize < 0` arm | **caught** by `TestReserveBuffer` |
| drop the overflow guard entirely | **caught** by `TestReserveBuffer` |
| drop the doubling (`newCapacity` stays `newSize`) | **caught** by `TestReserveBuffer` |

The last one is the notable one: the capacity assertion that was already in the test is what keeps the rewritten growth strategy honest, so the refactor landed under coverage that predates it rather than coverage written to match it.

**The workflow permissions are complete and correctly scoped.** All six docker workflows are `push`/`pull_request` on `main` and none of them uses `GITHUB_TOKEN`, `gh`, an upload action, or posts a comment — `contents: read` is sufficient for every one, and adding it at workflow level rather than job level covers future jobs in the same file.

**Local:** `go build ./...` clean, `go vet ./pkg/table/ ./pkg/dbconn/sqlescape/` clean, `gofmt -l pkg/` empty, `golangci-lint run ./pkg/table/... ./pkg/dbconn/sqlescape/...` → **0 issues**. `pkg/dbconn/sqlescape` passes in full; in `pkg/table` the chunk/boundary/datum tests pass, and the rest needs a live MySQL (confirmed locally — `Access denied for user 'spirit'@'localhost'`), so those were read rather than run. CI is 15/15 green, including all five docker suites and `zizmor`.

</details>

aparajon (approved):
🤖 Approving — clears all 13 open code-scanning alerts, the `chunk.go` key-escaping half was a real permanent-resume-failure bug, and the `jsonQuoteDatum` refactor is output-preserving where the watermark map keys depend on it. Findings in the review comment are all low / low-med and none block: the strongest is that the new panic sits six lines above an existing error return for a strictly-easier marshal, on a path where all nine callers already check the error.

morgo:
🤖 Addressed the review feedback in ff3761ec: production checkpoint serialization now propagates through the existing GetLowWatermark error channel; the quoted-key regression test exercises newChunkFromJSON and verifies both bounds; the watermark tuple-key comments now document the injectivity requirement; and the escape-size multiplication is guarded before evaluation. I also updated the PR description with the setup-go cache tradeoff. The branch has been updated with main.

PR #1214, opened 2026-09-05
test(throttler): decouple live query from replica lag

### LINKED ISSUE #1211: flaky test: TestThrottlerInterface depends on shared replica catch-up timing
🤖 Filed by my AI agent as follow-up from PR #1206.

## Observed failure

The MySQL 8.0.45 replica workflow failed in `TestThrottlerInterface`:

- Run: https://github.com/block/spirit/actions/runs/33973220199/job/101325304116
- Failure: `pkg/throttler/throttler_test.go:41: Condition never satisfied`
- `BlockWait` spent its full 60-second allowance throttled, then the 5-second `require.Eventually` still observed `IsThrottled() == true`.
- Logged replica lag remained roughly 75–89 seconds against a 60-second tolerance, ending at 88,587 ms.

The same run passed `pkg/datasync` and every package except `pkg/throttler`. PR #1206 changes datasync and related documentation/comments only; it does not change `pkg/throttler`. The other PR checks passed, and the 20 most recent main-branch runs of this workflow also passed.

## Why this is load-sensitive

The test says there is no activity and expects the replica to catch up quickly, but the workflow runs:

`go test -race -timeout=10m -parallel=4 ./...`

All packages share the same source and replica. During this failure, MySQL repeatedly reported redo-log capacity/checkpointer pressure while the rest of the integration suite generated writes. The replica therefore had real backlog when `TestThrottlerInterface` ran. Its fixed 60-second `BlockWait` plus 5-second assertion window makes suite load and scheduling part of the test outcome.

## Suggested follow-up

Make the test independent of shared-suite replica catch-up timing. Possible approaches:

- keep a real-database assertion for `NewReplicationThrottler` selection and a successful lag query, but exercise throttled/unthrottled state deterministically through `applyLag`;
- isolate the live catch-up assertion on a dedicated source/replica or workload phase;
- otherwise wait on an explicit replication condition with a suitable test deadline rather than assuming the shared replica becomes idle within 65 seconds.

## Acceptance criteria

- The throttler behavior remains covered.
- The test does not depend on unrelated packages stopping writes or the shared replica draining within a fixed short window.
- The replication workflow remains stable under its existing race-enabled, parallel full-suite command.

morgo:
Closes #1211

## Summary

- keep the live replication throttler test focused on performance_schema query compatibility while retaining a generous lag-value sanity check
- remove dependency on the shared CI replica draining below the lag threshold
- cover BlockWait deterministically across its immediate, cancellation, recovery, stale-signal, and 60-interval give-up paths
- centralize the replica BlockWait test timings so their relationships remain explicit

## Testing

- go test ./pkg/throttler
- go test -race ./pkg/throttler
- race-enabled BlockWait tests repeated 20 times

### DIFF
--- a/pkg/throttler/replica_test.go
+++ b/pkg/throttler/replica_test.go
@@ -92,6 +92,28 @@ func TestReplica_BlockWaitFailsClosedOnStaleSignal(t *testing.T) {
 	require.Less(t, elapsed, 500*time.Millisecond)
 }
 
+func TestReplica_BlockWaitReturnsWhenLagRecovers(t *testing.T) {
+	// Use a short blockWaitInterval so the test doesn't have to wait a full
+	// second per loop iteration. Save and restore.
+	prev := blockWaitInterval
+	blockWaitInterval = 10 * time.Millisecond
+	t.Cleanup(func() { blockWaitInterval = prev })
+
+	l := newTestReplica(t, 60*time.Second)
+	l.applyLag(60_000) // exactly at the tolerance, so BlockWait must hold
+
+	go func() {
+		time.Sleep(30 * time.Millisecond)
+		l.applyLag(59_999)
+	}()
+
+	start := time.Now()
+	l.BlockWait(t.Context())
+	elapsed := time.Since(start)
+	require.GreaterOrEqual(t, elapsed, 20*time.Millisecond, "BlockWait must block while lag is over budget")
+	require.Less(t, elapsed, 500*time.Millisecond)
+}
+
 func TestReplica_UpdateLagWrapsCause(t *testing.T) {
 	// sql.Open is lazy and does not connect. A pre-canceled context makes
 	// QueryRowContext fail deterministically with context.Canceled before any
--- a/pkg/throttler/replica_test.go
+++ b/pkg/throttler/replica_test.go
@@ -13,12 +13,26 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	replicaBlockWaitTestInterval = 10 * time.Millisecond
+	replicaBlockWaitRecoverAfter = 3 * replicaBlockWaitTestInterval
+	replicaBlockWaitMinDuration  = 2 * replicaBlockWaitTestInterval
+	replicaBlockWaitMaxDuration  = 50 * replicaBlockWaitTestInterval
+)
+
 func newTestReplica(t *testing.T, tolerance time.Duration) *Replica {
 	t.Helper()
 	return &Replica{
 		lagTolerance: tolerance,
 		logger:       slog.New(slog.NewTextHandler(io.Discard, nil)),
 	}
+}
+
+func useReplicaBlockWaitTestInterval(t *testing.T) {
+	t.Helper()
+	prev := blockWaitInterval
+	blockWaitInterval = replicaBlockWaitTestInterval
+	t.Cleanup(func() { blockWaitInterval = prev })
 }
 
 func TestReplica_LagBasedThrottling(t *testing.T) {
@@ -66,12 +80,27 @@ func TestReplica_NeverPolledIsNotStale(t *testing.T) {
 	require.False(t, l.IsThrottled())
 }
 
+func TestReplica_BlockWaitReturnsImmediatelyWhenUnthrottled(t *testing.T) {
+	l := newTestReplica(t, 60*time.Second)
+	start := time.Now()
+	l.BlockWait(t.Context())
+	require.Less(t, time.Since(start), 50*time.Millisecond)
+}
+
+func TestReplica_BlockWaitRespectsContext(t *testing.T) {
+	l := newTestReplica(t, 60*time.Second)
+	l.applyLag(60_000)
+
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel()
+
+	start := time.Now()
+	l.BlockWait(ctx)
+	require.Less(t, time.Since(start), 200*time.Millisecond)
+}
+
 func TestReplica_BlockWaitFailsClosedOnStaleSignal(t *testing.T) {
-	// Use a short blockWaitInterval so the test doesn't have to wait a full
-	// second per loop iteration. Save and restore.
-	prev := blockWaitInterval
-	blockWaitInterval = 10 * time.Millisecond
-	t.Cleanup(func() { blockWaitInterval = prev })
+	useReplicaBlockWaitTestInterval(t)
 
 	l := newTestReplica(t, 120*time.Second)
 	l.applyLag(5) // healthy baseline, well under the tolerance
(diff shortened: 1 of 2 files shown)


aparajon:
🤖 **Adversarial correctness review** — `f66e20e3` (+26/-12, 2 files)

Right trade, and made the right way round: the assertion that was flaky is deleted and the behaviour it covered is re-established somewhere deterministic, rather than the deadline being widened until the flake stopped reproducing. The live test keeps exactly the coverage only it can provide — `MySQL8LagQuery` is 27 lines of CTEs over three `performance_schema` tables with `GTID_SUBTRACT` and a `NOW(3)` in it, and nothing but a real 8.0 replica can tell you it still parses and scans.

⭐ **The new comment's claim about the live test is accurate, which I checked rather than assumed.** "Open performs an initial lag query" holds because `Open` calls `UpdateLag` and *returns* its error rather than logging it, so `require.NoError(t, throttler.Open(...))` genuinely fails if the query breaks against a future MySQL. And the deleted comment above it — *"will attach either MySQL 8.0 or MySQL 5.7 throttler"* — was already false before this PR: `NewReplicationThrottler` returns `&Replica{}` unconditionally, with no version branch left. Deleting it is a correction, not a lost comment.

⭐ **The new test is a faithful port of the local pattern, not an invention.** It matches `TestAuroraThreads_BlockWaitReturnsWhenThrottlingClears` line for line — same 10ms interval, same 30ms recovery goroutine, same 20ms/500ms bounds — and `applyLag(60_000)` against a 60s tolerance is correctly "at the tolerance," since `IsThrottled` compares with `>=`.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low-med | `replica.go:158` | `BlockWait`'s two exits are indistinguishable to its caller, and the give-up exit has no test anywhere in the package — measured: cutting the 60-iteration cap to 3 leaves every `BlockWait` test in `pkg/throttler` green |
| 2 | low | `replica_test.go` | `Replica` gets the third of the three `BlockWait` tests its aurora sibling has; the other two are ports of ~8 lines each, and one of them is a documented property nothing checks |
| 3 | low | `replica_test.go:95` | The two `BlockWait` tests are now twelve duplicated lines apart on one setup line, with five coupled timing literals and no stated relationship between them |
| 4 | low | `throttler_test.go:29` | The removed assertion was also the only check that the real lag query returns a plausibly-scaled number, and that part wasn't the flaky part |

---

### 1 — the exit this PR doesn't cover is the one where `BlockWait` doesn't do what its name says (low-med)

```go
	for range 60 {
		if !l.IsThrottled() {
			return          // the throttle cleared
		}
		...
	}
	l.logger.Warn("lag monitor timed out", ...)
	                        // ...and this also just returns
```

Both exits are a bare return, so the copier cannot tell "the replica caught up" from "I waited 60 intervals and I'm proceeding anyway." That's deliberate and documented (*"or up to 60s to allow some progress to be made"*), so it isn't a defect. But it does mean the give-up budget is load-bearing: it is the entire difference between throttling and not throttling once lag is over tolerance, and it is a bare `60` multiplied by a package-level `var` that three tests already reassign.

Nothing pins it. I measured this rather than inferring it — changing `for range 60` to `for range 3`, which is exactly the shape of a mistake someone makes while tuning the interval, leaves **every** `BlockWait` test in the package passing, including both of the ones this PR touches:

```
--- M1: iteration cap 60 -> 3 (BlockWait returns without the throttle clearing)
ok  	github.com/block/spirit/pkg/throttler
```

I also tried the obvious hardening — adding `require.False(t, l.IsThrottled())` after the call — and it does **not** close the gap, because at a 10ms interval the shrunken cap expires at the same instant the recovery goroutine fires, so the state check passes on the give-up path too. Worth saying so that it doesn't look like a one-line fix that got skipped.

What actually closes it is the missing test, and it's the natural sibling of the two that now exist: arm the throttle, never recover, assert `BlockWait` returns *and* that `IsThrottled()` is still true — which is only reachable via the timeout path. At a 10ms interval that's a 600ms test. Two tests for the release path and none for the give-up path is the wrong ratio when the give-up path is the one that lets the copier run against an over-budget replica.

### 2 — two thirds of the aurora `BlockWait` triad didn't come across (low)

`aurora_threads_test.go` has three:

```go
func TestAuroraThreads_BlockWaitReturnsImmediatelyWhenUnthrottled(t *testing.T)
func TestAuroraThreads_BlockWaitRespectsContext(t *testing.T)
func TestAuroraThreads_BlockWaitReturnsWhenThrottlingClears(t *testing.T)
```

This PR ports the third. `Replica` has neither of the first two, and the second is a property `Replica.BlockWait` documents about itself:

> BlockWait blocks until the throttle clears ... **It respects context cancellation.**

Nothing verifies that for `Replica`. The aurora version is eight lines and ports with one substitution — `l.applyLag(60_000)` where aurora does `a.isThrottled.Store(true)` — and it doesn't even need the interval mutation, since a pre-cancelled context returns on the first `select`. The unthrottled fast path is three lines. Since this PR is exactly the move of `BlockWait` coverage out of the live test and into the unit tests, and the harness is already sitting there, that's the moment to bring the other two rather than the moment to note them.

### 3 — five timing literals with an unstated relationship, now written twice (low)

`TestReplica_BlockWaitReturnsWhenLagRecovers` and `TestReplica_BlockWaitFailsClosedOnStaleSignal` are the same twelve lines — the same save/restore, the same 30ms goroutine, the same two assertions, near-identical comments — differing only in the tolerance and in the one line that arms the throttle (`ageLastSample` vs `applyLag(60_000)`). The third copy of this is now free to write and slightly likelier to drift.

The numbers are the part worth pulling out, because their relationships are what make the tests correct and none of them is stated: the interval (10ms) sets the poll granularity, the goroutine's 30ms must exceed a couple of intervals, the 20ms lower bound must sit below 30ms but above zero, and the 500ms upper bound must sit below the give-up budget of `60 × interval` = 600ms. That last margin is 100ms, and it is the only thing distinguishing a released `BlockWait` from a timed-out one — which is finding 1 restated as arithmetic. Deriving them (`recoverAfter := 3 * blockWaitInterval`, `require.GreaterOrEqual(elapsed, 2*blockWaitInterval)`) would keep the test correct if someone retunes the interval, and would make the 500ms-vs-600ms squeeze visible instead of coincidental.

### 4 — the lag *reading* wasn't the flaky part, and it lost its only check (low)

The rationale is sound: whether the shared replica is caught up depends on what else is running against it, so `!IsThrottled()` was never this test's to assert. But the deleted block was doing two things at once, and only one of them was flaky. It was also the only place the value coming out of `MySQL8LagQuery` was checked for plausibility — and that query converts microseconds to milliseconds in two separate branches (`/1000` under `TIMESTAMPDIFF(MICROSECOND, ...)`) and wraps the whole thing in `IFNULL(..., 0)`. A unit slip or an always-zero `IFNULL` now produces a green live test, because `Open` only proves the query executed and scanned into an `int64`.

A bound that is immune to the actual flake keeps that check: the CI replica may be seconds or minutes behind, but it will not be hours behind, so asserting the observed lag is under something generous distinguishes "plausible reading" from "wrong by three orders of magnitude" without assuming anything about catch-up. The test is in-package, so `throttler.(*Replica)` and `currentLagInMs.Load()` are both reachable after the `Open`. Non-zero would be too strong — an idle replica correctly reports 0 — so an upper bound is the assertion that survives.

---

<details>
<summary><b>Verified</b> — the boundary, the global mutation, and two attacks that dissolved</summary>

⭐ **Attack that dissolved: `applyLag(60_000)` not actually being over budget.** The comment claims "exactly at the tolerance, so BlockWait must hold," and if `IsThrottled` used `>` the test would return at `elapsed ≈ 0` and fail its own lower bound — a test that only passes by accident of a boundary it doesn't own. It's `>=`: `l.currentLagInMs.Load() >= l.lagTolerance.Milliseconds()`, and `TestReplica_LagBasedThrottling` already pins that boundary independently with `applyLag(120_000)` against a 120s tolerance. The premise holds and it holds for a stated reason.

⭐ **Attack that dissolved: the recovery goroutine racing `BlockWait`'s reads.** The new test writes lag from a second goroutine while `BlockWait` polls it, which is a data race unless both fields are guarded — and `applyLag` touches two, `currentLagInMs` and `stale`. `go test -race` on all six `TestReplica_*` tests is clean, and `goleak.VerifyTestMain` is satisfied because the 30ms goroutine has finished long before the package ends.

**Mutating the package-level `blockWaitInterval` is safe here, and is the established idiom.** Three existing tests do exactly this save/restore (`replica_test.go:73`, `aurora_threads_test.go:199`, `aurora_commitlatency_test.go:216`), and no test anywhere in `pkg/throttler` calls `t.Parallel()`, so the global cannot be observed mid-mutation by a concurrent test. The `t.Cleanup` restore is correct rather than merely conventional.

**The lower bound is doing real work.** Dropping the `timer.Reset` / `select` block from `BlockWait` so the loop spins is caught cleanly — `"7.375µs" is not greater than or equal to "20ms"` — so `require.GreaterOrEqual` genuinely proves the call blocks, and isn't one of those bounds that passes either way.

**The rename costs nothing structurally.** `TestThrottlerInterface` was not what asserted the interface is satisfied; `var _ ReasonedThrottler = &Replica{}` in `replica.go` is, at compile time. `TestReplicationThrottlerLiveQuery` also describes what the test does now, which the old name had stopped doing.

**The live test does run somewhere.** `REPLICA_DSN` is set in `compose/replication-tls/replication-ci.yml`, so this isn't a test that skips everywhere and was flaky only in theory.

**Local:** `go build ./...` clean, `go vet ./pkg/throttler/` clean, `gofmt -l pkg/throttler/` empty, `golangci-lint run ./pkg/throttler/...` → **0 issues**. `go test -race ./pkg/throttler/` passes except `TestAuroraVCPUs_LocalMySQL`, which needs a live server (`Access denied for user 'spirit'@'localhost'`) and is unrelated. CI is 15/15 green.

</details>

aparajon (approved):
🤖 Approving — the right shape of flake fix: the flaky assertion is deleted and the behaviour it covered is re-established deterministically, with no deadline widened. Findings in the review comment are all low / low-med. The one worth a look before merge: nothing in `pkg/throttler` pins `BlockWait`'s give-up exit, and I measured that cutting the 60-iteration cap to 3 leaves every `BlockWait` test green — that's the natural sibling of the test this PR adds.

Issue #69, opened 2023-03-30
runner.Run should terminate when its context is cancelled


morgo:
If this is possible, it will allow for better testing of resume from checkpoint functionality. Here's the test I'm trying to write, but it won't work yet:

```
func TestResumeFromCheckpointE2E(t *testing.T) {
	// Lower the checkpoint interval for testing.
	checkpointDumpInterval = 1 * time.Second
	defer func() {
		checkpointDumpInterval = 50 * time.Second
	}()

	runSQL(t, `DROP TABLE IF EXISTS chkpresumetest, _chkpresumetest_shadow, _chkpresumetest_chkpnt`)
	table := `CREATE TABLE chkpresumetest (
		id int(11) NOT NULL AUTO_INCREMENT,
		pad varbinary(1024) NOT NULL,
		PRIMARY KEY (id)
	)`
	runSQL(t, table)
	migration := &Migration{}
	cfg, err := mysql.ParseDSN(dsn())
	assert.NoError(t, err)

	// Insert dummy data.
	runSQL(t, "INSERT INTO chkpresumetest (pad) SELECT RANDOM_BYTES(1024) FROM dual")
	runSQL(t, "INSERT INTO chkpresumetest (pad) SELECT RANDOM_BYTES(1024) FROM chkpresumetest a, chkpresumetest b, chkpresumetest c LIMIT 500000")
	runSQL(t, "INSERT INTO chkpresumetest (pad) SELECT RANDOM_BYTES(1024) FROM chkpresumetest a, chkpresumetest b, chkpresumetest c LIMIT 500000")
	runSQL(t, "INSERT INTO chkpresumetest (pad) SELECT RANDOM_BYTES(1024) FROM chkpresumetest a, chkpresumetest b, chkpresumetest c LIMIT 500000")
	runSQL(t, "INSERT INTO chkpresumetest (pad) SELECT RANDOM_BYTES(1024) FROM chkpresumetest a, chkpresumetest b, chkpresumetest c LIMIT 500000")

	// use as slow as possible here: we want the copy to be be still running
	// when we kill it after 3 seconds.
	migration.Host = cfg.Addr
	migration.Username = cfg.User
	migration.Password = cfg.Passwd
	migration.Database = cfg.DBName
	migration.Concurrency = 1
	migration.Checksum = true
	migration.Table = "chkpresumetest"
	migration.Alter = "ENGINE=InnoDB"
	migration.TargetChunkMs = 1

[...]

PR #1215, opened 2026-09-05
fix(sqlescape): use checked slice growth

morgo:
## Summary

Follow-up to #1213 for the successor `go/allocation-size-overflow` finding reported as [code-scanning alert #69](https://github.com/block/spirit/security/code-scanning/69).

- Replace hand-rolled capacity arithmetic with the standard library's checked `slices.Grow` implementation.
- Retain the explicit negative and combined-length overflow guard.
- Retain the caller-side guard that checks the escape-size multiplication before evaluating it.
- Preserve existing buffer contents and required capacity without depending on an exact implementation-specific growth factor.

## Testing

```text
go test ./pkg/dbconn/sqlescape -count=1
golangci-lint run ./pkg/dbconn/sqlescape/...
git diff --check
```

### DIFF
--- a/pkg/dbconn/sqlescape/utils.go
+++ b/pkg/dbconn/sqlescape/utils.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -30,20 +31,9 @@ func reserveBuffer(buf []byte, appendSize int) []byte {
 	if appendSize < 0 || appendSize > maxInt-len(buf) {
 		panic("sqlescape: buffer size overflow")
 	}
-	newSize := len(buf) + appendSize
-	if cap(buf) < newSize {
-		// Preserve the historical doubling growth strategy when it is safe, but
-		// skip the extra capacity when doubling would overflow int. newSize
-		// itself was checked above.
-		newCapacity := newSize
-		if len(buf) <= maxInt-newSize {
-			newCapacity += len(buf)
-		}
-		newBuf := make([]byte, newCapacity)
-		copy(newBuf, buf)
-		buf = newBuf
-	}
-	return buf[:newSize]
+	originalLen := len(buf)
+	buf = slices.Grow(buf, appendSize)
+	return buf[:originalLen+appendSize]
 }
 
 // escapeBytesBackslash will escape []byte into the buffer, with backslash.
--- a/pkg/dbconn/sqlescape/utils_test.go
+++ b/pkg/dbconn/sqlescape/utils_test.go
@@ -34,7 +34,7 @@ func TestReserveBuffer(t *testing.T) {
 
 	res2 := reserveBuffer(res1, 9)
 	require.Len(t, res2, 12)
-	require.Equal(t, 15, cap(res2))
+	require.GreaterOrEqual(t, cap(res2), len(res2))
 	require.Equal(t, res1, res2[:3])
 
 	require.PanicsWithValue(t, "sqlescape: buffer size overflow", func() {


aparajon:
🤖 **Adversarial correctness review** — `3d85238c` (+5/-15, 2 files)

Right conclusion, and the diff is net negative lines, which is the shape this should have. Alert #69 is the lesson that hand-rolled overflow-checked arithmetic doesn't satisfy the checker no matter how carefully it's written — the previous `newCapacity += len(buf)` was *correct*, and CodeQL still couldn't prove it, so it produced a successor alert at the new line. Delegating to `slices.Grow` gets a growth path that is checked, maintained and tested elsewhere, which is better than being right in a way a tool can't follow. Verified: #69 is the only open alert on `main`, and the branch reports zero.

⭐ **The caller-side guard is the other half, and it's correct.** It isn't in this diff — it landed on `main` with #1213 after the revision I looked at — so I checked it rather than assuming: `if len(v) > maxInt/2 { panic }` in `escapeBytesBackslash`, ahead of `len(v)*2`. The boundary is exact. At `len(v) == maxInt/2` (= 2⁶²−1 on 64-bit) the product is `maxInt−1` and fits, and the guard correctly lets it through; at `len(v) == 2⁶²` the product wraps and the guard fires. That closes the arithmetic where it actually happens, which is what makes `reserveBuffer`'s remaining guard a backstop rather than the only line of defence.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low-med | `utils.go:35` | `slices.Grow` doesn't zero the reserved region the way `make`+`copy` did — measured, stale bytes from the old slice's unused capacity now survive into the returned slice, and what keeps that safe is an invariant nothing states |
| 2 | low-med | `utils_test.go:37` | The replacement assertion is a Go language tautology and cannot fail; the one it replaced was the only thing pinning growth behaviour |

---

### 1 — the reserved region changed from zeroed to uninitialized (low-med)

The PR body says it preserves "existing buffer contents," and that part is true — the first `len(buf)` bytes survive. The unstated change is what's in the bytes *after* them.

The old code allocated with `make([]byte, newCapacity)` and then `copy(newBuf, buf)`, and `copy` moves `len(buf)` bytes, not `cap(buf)` — so everything from `len(buf)` to the end was freshly zeroed. `slices.Grow` is implemented as `append(s[:cap(s)], make([]E, n)...)[:len(s)]`, and that `s[:cap(s)]` is the difference: it carries the old slice's *unused capacity* across the reallocation. Measured on this branch:

```
before: len=3 cap=5 tail=aaaa            // 0xAA garbage at indices 3,4
after:  len=12 cap=16 bytes[3:12]=aaaa00000000000000
NONZERO bytes in the reserved region: 2
```

The old code returned nine zeros there. This returns two bytes of whatever was in the previous slice's slack.

Not currently reachable, and I traced it rather than assuming. `reserveBuffer` has exactly one non-test caller, `escapeBytesBackslash`, which reserves the worst case `len(v)*2`, writes at `pos` for every byte of `v`, and returns `buf[:pos]` — so the region that carries stale bytes is always between the final `pos` and the reserved end, and always truncated away. Nothing reads it.

It's worth a line anyway, for two reasons specific to this package. The shape that produces the garbage is the *normal* one here, not a contrived one: `EscapeSQL` calls `escapeBytesBackslash(buf, v)` in a loop (`:208`, `:215`), and each call over-reserves and truncates, so every argument after the first arrives with exactly the probe's layout — a short `len`, a larger `cap`, and the previous argument's escaped output sitting in the slack. And the bytes in that slack are escaped fragments of earlier values in the same query. A future caller that reserved more than it wrote and returned the full reserved length would splice one value's bytes into another's, in the one package whose whole job is producing SQL text safely. The old implementation was immune to that by construction; this one is immune by a property of its single caller.

So the ask is just to write the property down where the next caller will see it — a sentence on `reserveBuffer` that the reserved region is uninitialized and callers must write every byte they return. I'd take that over the alternative: `clear(buf[originalLen : originalLen+appendSize])` restores the old guarantee in one line, but it doubles the bytes written per escape call on a hot path, to defend a case no caller has. The comment is the cheaper half of the trade and the one that survives the next refactor.

### 2 — the new assertion can't fail (low-med)

```go
-	require.Equal(t, 15, cap(res2))
+	require.GreaterOrEqual(t, cap(res2), len(res2))
```

Dropping the exact `15` is right — that number was `2*len(buf)+appendSize` from an implementation this PR deliberately deletes, and pinning stdlib's growth factor would be pinning something that isn't yours. But `cap(s) >= len(s)` is a Go language invariant. It holds for every slice that has ever existed, including one returned by a completely broken `reserveBuffer`. The line reads like a capacity check and is closer to a no-op.

That matters here because of what it replaced. On #1213 I measured that `require.Equal(t, 15, cap(res2))` was the *only* assertion in the package that killed a "drop the doubling" mutation — it was carrying the growth behaviour by itself. Now nothing does:

| Mutation | Result |
|---|---|
| `slices.Grow(slices.Clone(buf), appendSize)` — always reallocate, discarding spare capacity | **survives** |
| `slices.Grow(buf, appendSize*10)` — over-reserve tenfold | **survives** |

Both leave the whole package green.

The first of those is the interesting one, because avoiding that reallocation is the entire reason `reserveBuffer` used to test `cap(buf) < newSize` before allocating — and this PR hands that decision to `slices.Grow`. So the property that's still `reserveBuffer`'s to own, and is now completely untested, is: *when the existing capacity already suffices, no reallocation happens*. That's assertable without touching stdlib's growth factor, by taking `&res1[0]` (or `cap`) before and after a call that fits in the existing slack and requiring it unchanged. If that's not judged worth a test, deleting the line is better than keeping a tautology in its place — a test that cannot fail is worse than no test, because the next reader counts it.

---

<details>
<summary><b>Verified</b> — the guard chain, the edges, and an attack that dissolved</summary>

⭐ **Attack that dissolved: `slices.Grow` panicking with its own message instead of the package's.** `slices.Grow` panics `"cannot be negative"` for a negative `n`, and its internal `append` can panic `"growslice: len out of range"` — either would replace the package's `"sqlescape: buffer size overflow"` and break `require.PanicsWithValue`. Neither is reachable: the retained guard runs first and establishes `0 <= appendSize <= maxInt-len(buf)`, and since `cap(s) >= len(s)`, the capacity `slices.Grow` targets is `cap(s) + (appendSize - (cap(s)-len(s)))` = `len(s) + appendSize <= maxInt`. So the checked-arithmetic guarantee survives the delegation, which is the thing the PR needed to get right.

**The edge cases behave identically to the old implementation.** `reserveBuffer(nil, 0)` returns an empty slice; `reserveBuffer(nil, n)` returns `n` zeroed bytes (`append(nil[:0], make(n)...)` allocates fresh, so there is no slack to inherit); `appendSize == 0` is a no-op on a slice that already has the length. The no-reallocation path is also byte-identical — when `cap-len >= appendSize`, `slices.Grow` returns the input unchanged and the old code skipped its `make` for the same reason, so both return the same view over the same array, stale slack included. Finding 1 is only about the reallocating path.

**Content preservation holds and is tested.** `require.Equal(t, res1, res2[:3])` with `res1[1] = 3` set beforehand covers the copy, and `originalLen` is captured before the `Grow` call rather than read back from the returned slice — which matters, since `slices.Grow` returns a slice whose length equals the *input's* length, so re-reading `len(buf)` after the call would have worked by accident and broken the moment `Grow`'s contract was read differently.

**The overflow guards are still both tested.** `require.PanicsWithValue` on `(make([]byte, 1), math.MaxInt)` and `(nil, -1)` are untouched and still pass, so the two arms of the retained guard keep their coverage; only the capacity assertion was weakened.

**Local:** `go build ./...` clean, `go vet ./pkg/dbconn/sqlescape/` clean, `gofmt -l pkg/dbconn/` empty, `golangci-lint run ./pkg/dbconn/sqlescape/...` → **0 issues**. The package's tests pass in full with no server. CI is 15/15 green.

</details>

aparajon (approved):
🤖 Approving — right conclusion and a net-negative diff. Verified alert #69 is the only one open on `main` and the branch reports zero, and that the retained guard still makes the delegation to `slices.Grow` safe (it can't reach either of `Grow`'s own panics). Two low-med findings, neither blocking: `slices.Grow` no longer zeroes the reserved region — measured, stale bytes from the old slice's slack now survive the reallocation, which is safe only because `escapeBytesBackslash` writes every byte it returns; and the replacement capacity assertion is a Go language tautology that cannot fail, so nothing pins growth or no-realloc behaviour any more.

PR #1216, opened 2026-09-05
feat(move): add host-aware autoscaling for sharded moves

### LINKED ISSUE #1212: support autoscaling for sharded moves
In schema changes (migration) we now support autoscaling for Aurora, and its been working really well. While still technically marked as experimental, it is preferable IMHO to setting `--threads` and `--write-threads`.

However, the way it is implemented it is very much about the load on a single target. In sharded use cases, we need to consider that the applier will be writing to multiple backend servers, which have their own levels of load.

This might be considerable effort to refactor, so I am also open to a middle ground that gets us most of the way.

morgo:

Closes #1212.

Adds `move --enable-experimental-autoscaling` for single-target and sharded moves. Each shard retains its own queue and worker pool; all pools scale together using the maximum utilization reported by the Aurora target hosts:

- All hosts have headroom: permit growth.
- One host is moderately loaded: hold.
- Any host becomes busy: reduce concurrency, regardless of which host was busiest previously.

This deliberately follows the lowest common denominator. A copy chunk completes only after all its chunklets finish, so scaling quieter hosts independently does not remove the bottleneck. The chunker and enqueue ordering remain unchanged.

Throttling ownership stays consistent with migration: the copier pauses before reading another chunk, and its controller adjusts the applier through `SetWriteWorkers`. Already-read and queued work continues draining. Moving throttling into the applier is a separate refactor; this PR adds no applier throttler configuration or extra fixed host concurrency guard.

The new `pkg/host` abstraction groups connections by transport and address independently of schema and credentials. Index restoration and monitor ownership share these groups. Initial counts and scaling bounds account for target capacity, co-located shards and the client CPU budget. All target hosts must qualify for Aurora autoscaling before configured thread counts are overridden. Fresh and resumed moves use the same setup, with dedicated monitor connections and teardown. The initial and sentinel-wait checksums share the load signal.

Both appliers share one worker-pool implementation for resizing, cooperative retirement and joining, while keeping their queue routing and completion aggregation. Sentinel progress reports load throttling only during an active continuous checksum pass. Reverse-window write concurrency retains the configured value independently of forward autoscaling, including across resume.

Connection budgets (`--max-connections` for move and sync) merged in #1217 and are now inherited from main. This PR only fits the autoscaling read ceiling to that existing budget; it does not grow the pools. Structured progress and datasync log-format alignment merged in #1218 and are also inherited from main. Autoscaling now supplies throttle status alongside those structured fields using the same state snapshot. The flag implementation and general progress changes are excluded from this PR.

Validation: targeted MySQL integration tests with the race detector cover per-shard worker resizing, balanced traffic followed by chunks routed entirely to one shard, cancellation, shutdown/restart and sharded moves. Controller tests cover growth, hold, reduction and changes in which host is busiest. Capacity, host grouping and non-Aurora fallback tests also pass; index restoration and checkpoint resume were validated during this PR. Continuous-checksum throttle reporting and configured reverse-window concurrency have regression coverage. Repository-wide lint is clean.

Actual Aurora end-to-end validation remains outstanding. Reverse-window destinations do not acquire new autoscaling monitors.

Updated on top of merged #1217 and #1218. Targeted race tests cover fixed-pool autoscaling bounds, small-pool moves, sharded moves, checkpoint resume and worker resizing; repository-wide lint is clean.


aparajon:
🤖 **Adversarial correctness review** — `8e20bdac` (22 files, +815/−131)

Strong change, and the part I want to call out first is the half that wasn't asked for: `#1212` could have been satisfied by forking `SingleTargetApplier`'s scaling code into `sharded.go`, and instead the resizable-worker lifecycle came out into `workerPool` and host grouping came out into `pkg/host`. That is the "or, better, extract it into a shared package" branch of AGENTS.md's triplet rule, taken voluntarily, and `single_target.go` losing 66 lines is the evidence it worked.

`docs/move.md` is also unusually honest for a feature doc — "a busy host slows the whole move; idle hosts do not offset its load", "there is no additional fixed host concurrency guard", "which host is busiest can change from one sample to the next". Those are the sentences an operator needs and the ones most easily left out.

So I went after the two claims the change rests on: that the new shared pool cannot lose accepted work, and that move's bounds are as conservative as migration's. The second holds under everything I threw at it — 7 of 8 mutations died, including all five on `moveAutoscaleBounds`. The one that survived is the first finding.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low | `worker_pool.go:60` | The pool's one stated safety invariant — seal must not retire the workers that still have to drain accepted work — is not pinned by its own unit test. Breaking it deliberately passes `TestWorkerPoolLifecycle`, and the failure mode in `ShardedApplier` is a hung `Wait()` |
| 2 | low | `runner.go:284` | `throttlerMu` arrives with this PR and guards 2 of the 7 reads of the field it protects. The five direct reads are safe today by call-graph accident, not by anything a reader can see |
| 3 | low | `autoscale.go:66` | A failed Aurora probe and a target that is simply not Aurora take one branch and log one message, so an ordinary MySQL target logs `"error": nil` under a message about a missing signal. Migration splits exactly these two cases on purpose |
| 4 | low | `pkg/datasync` | The triplet rule asks for an explicit port-or-extract decision on the third runner, and neither the PR nor AGENTS.md's drift list records one for autoscaling |
| 5 | nit | `sharded.go:467-469` | `Stats()`' comment still describes the counter this PR deleted |

---

### 1 — the pool's drain invariant is unpinned (low)

`seal`'s comment states the rule it exists to hold:

```go
// seal excludes future adds before the applier closes input and waits. Do not
// retire all workers here: at least one must drain already accepted work.
```

That is the load-bearing sentence in the file, because `ShardedApplier.Stop()` is `seal()` → `close(chunkletBuffer)` → `wait()`, and the pre-existing contract on `writeWorker` (`sharded.go:510-517`) says what happens if a buffered chunklet never gets picked up: *"every chunklet that made it into the buffer was already registered in pendingWork by Apply(), so it MUST produce a completion or Wait() will hang."* So violating `seal`'s comment doesn't lose rows quietly — it hangs the move at shutdown, which is the worst shape for this failure to take.

I mutated `seal` into the thing its own comment forbids:

```go
 func (p *workerPool) seal() {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.closed = true
+	p.resizeLocked(0)
 }
```

`TestWorkerPoolLifecycle` **passes**. It can't see the difference, because it calls `cancel()` right after `seal()` and its worker body returns on `ctx.Done()` as well as on `quit` — so every worker exits either way and retirement is unobservable. The rest of the applier suite needs MySQL, so on a workstation nothing catches it at all.

What does catch it is about twenty lines. This is the test I ran, and it fails on the mutant (`done` lands in the 190s, not 200) and passes clean:

```go
// Stop()'s contract is seal + close(buffer) + wait, and every item already
// accepted into the buffer must still be drained.
func TestWorkerPoolSealDrainsAcceptedWork(t *testing.T) {
	const items = 200
	work := make(chan int, items)
	for i := range items {
		work <- i
	}
	var done atomic.Int32
	var p workerPool
	p.start(t.Context(), 2, func(ctx context.Context, quit <-chan struct{}) {
		for {
			select {
			case <-quit:
				return
			case _, ok := <-work:
				if !ok {
					return
				}
				done.Add(1)
			}
		}
	})
	p.seal()
	close(work)
	p.wait()
	require.Equal(t, int32(items), done.Load(),
		"seal must not retire the workers that still have to drain accepted work")
}
```

Two details that matter if you take it: the ctx is never cancelled, since cancellation is what masks the bug; and the buffer is deep enough that the `select` coin-flip cannot drain it by luck.

The existing test is not wasted — the `resize`-racing-`seal` section is the sharpest thing in it, and `resize(0)` → `count() == 1` does pin the `max(1, n)` floor (mutating that floor to `n` is killed). It's specifically the drain half that has no coverage below the MySQL layer.

### 2 — a mutex that doesn't cover its own field (low)

`throttlerMu` and the `setThrottler`/`currentThrottler` pair are new here, and they exist because `Progress()` and the repl feed's `UnderLoad` closure read the throttler from other goroutines. But only those two call sites go through the accessor. Five reads take the field directly: `runner.go:284`, `:560`, `:1058`, `:1869`, `:2113`.

I could not turn any of them into a race, and I want to be clear that I tried: the only writers are `setThrottler` at `autoscale.go:57` and `:125`, both inside `setupAutoscaling`, which runs synchronously in `setupUnderLocks` before the goroutines that could read concurrently exist. `:560`/`:1058`/`:1869`/`:2113` are all on that same single-threaded setup path. So this is not a live bug.

It's still worth closing, for two reasons. `Close():284` is the one read whose position isn't obviously ordered against the writers — it runs after `watchTaskWait()`, but the repl clients aren't closed until `:297`, so the goroutine that *does* use the accessor is still alive at that point. And the next person to add a read has no way to tell from the code which list they should join. `currentThrottler()` is an inlined `RLock`; using it at all seven sites costs nothing and makes the guard self-describing.

(Separately, and this one is fine: `Close` closing the throttler at `:284` before the repl clients at `:297` means `UnderLoad` can call `IsThrottled()` on a closed throttler. I checked — `AuroraThreads.IsThrottled` is an `atomic.Bool` load and `Close` only stops the poller, so a post-close read returns the last sample rather than touching the closed monitor pool. Harmless, but it depends on a property of the throttler rather than on the ordering here.)

### 3 — two causes, one branch, one nil error (low)

```go
aurora, err := throttler.IsAurora(ctx, target.DB)
if err != nil || !aurora {
	r.logger.Warn("move autoscaling disabled: every target must provide an Aurora load signal", "target", targetKey(target), "error", err)
	return nil
}
```

The *policy* is right and I'm not asking to change it — all targets or none is the conservative call for a lockstep controller, and `docs/move.md` already states the three outcomes separately ("Non-Aurora hosts, small instances or failed Aurora probes retain the configured fixed thread counts. Capacity-query and monitor-startup failures abort setup."), which matches the code exactly.

It's the log line. Point a sharded move with the flag on at plain MySQL targets — a perfectly ordinary thing to do — and every host emits a warning about a signal it was never going to provide, carrying `"error": nil`. Meanwhile a genuinely broken probe (perf_schema locked down, monitor user under-granted) emits the *same* message, so the case an operator needs to see is indistinguishable from the case they configured. Migration's counterpart (`pkg/migration/runner.go:755-780`) splits these deliberately, with `err.Error()` only in the error arm and no log at all when the target simply isn't Aurora:

```go
switch {
case err != nil:
	r.logger.Warn("autoscaling disabled: could not determine whether the target is Aurora; ...", "error", err.Error(), ...)
	autoscaleEnabled = false
case !isAurora:
	// Left enabled deliberately: ...
```

A `switch` with the two arms separated, Warn for the probe failure and Info (or nothing) for the non-Aurora target, is a few lines and makes the fleet-wide signal readable. Worth noting too that the one test covering this branch, `TestMoveAutoscaleNonAurora`, builds a real table via `testutils.NewTestTable`, so this path has no coverage outside a MySQL environment either.

### 4 — the third runner (low)

AGENTS.md is direct about this: *"When you add a teardown step, a new lifecycle phase, or a safety gate, grep all three `runner.go` files and decide explicitly: port, or extract."* This PR adds a lifecycle step (`setupAutoscaling` in `setupUnderLocks`) and a teardown step (the throttler and `monitorDBs` in `Close`) to move's runner. `pkg/datasync` has `WriteThreads`, the same copier and applier, and no reference to autoscaling anywhere — so this moves the triplet from 1-of-3 to 2-of-3, and "Not yet unified (live drift — touch with care)" doesn't list autoscaling among the known divergences.

I'd guess declining is the right answer — datasync runs continuously against a converging target, so "derive counts once at setup from target capacity" may not even be the right shape for it. But that reasoning is the deliverable the rule asks for. One line in the summary, or a bullet added to the drift list, is enough; the point is that the next person to read three `runner.go` files finds the decision recorded instead of re-deriving it.

### 5 — stale comment (nit)

```go
// Stats returns a point-in-time snapshot of the write pipeline, aggregated
// across shards: queue depth/cap are summed, and active workers is the sum of
// each shard's live (started minus finished) workers.
```

"started minus finished" was `writeWorkersCount - writeWorkersFinished`, and this PR deletes `writeWorkersFinished`; the line immediately below the comment is now `shard.workers.count()`. Same number, different mechanism — and the new one can transiently exceed `len(quits)` during a shrink, which "started minus finished" wouldn't have suggested.

---

<details>
<summary><b>Verified</b> — 7/8 mutations killed, the redo-log gate move doesn't call, and why <code>Stop</code>'s sequential per-shard close can't deadlock</summary>

**Mutation testing — 8 run, 7 killed.** Baseline first, because most of `pkg/applier` and `pkg/move` needs MySQL on this box and an uninterpretable "kill" is worse than no result: the pure-unit subset (`TestWorkerPoolLifecycle`; `TestMoveAutoscaleBounds`, `Disabled`, `FitsFixedPool`, `TestMoveContinuousChecksumThrottleProgress`, `TestReverseWindowPreservesConfiguredWorkers`, `TestMoveThrottleStatus`; all of `pkg/host`) passes clean under `-race`, and everything below is measured against that. `moveAutoscaleBounds` is airtight — all five policy mutations died in `TestMoveAutoscaleBounds`: `smallest = min(...)` → `max` (the documented "smallest target, not the sum" rule), dropping the `/maxShardsPerHost` division on the read ceiling, dropping the `clientCeiling` cap on `readMax`, dropping the per-target `writeBudget` division, and dropping the `readStart = min(readStart, readMax)` clamp. `pkg/host` killed both of its: defaulting the empty network to something other than `"tcp"`, and dropping `Net` from the group key. The survivor is finding 1.

**The one gate migration has that move doesn't call — and it's unreachable by construction.** `moveAutoscaleBounds` computes `MaxThreads` as `autoscale.Ceiling(writeStart, true)` with the flag hardcoded, where migration routes through `throttler.ResolveMaxWriteThreads(start, autoscaleEnabled, redoAware, commitLatencyEnabled)`, whose whole purpose is `unguardedRedo := redoAware && !commitLatencyEnabled` — in redo-aware mode with no commit-latency backstop it returns `start`, refusing to grow the write pool at all so the autoscaler can shed but never oversubscribe the redo log unguarded. Move never probes `CanReadRedoAwareThreads`, so on the surface it looks like it skipped the gate. It didn't: `AuroraSetup.Build` enables commit latency iff `CommitLatencyThreshold > 0`, `setupAutoscaling` hardcodes `100 * time.Millisecond` (and `Build` errors out rather than silently omitting the throttler if construction fails), so `commitLatencyEnabled` is always true for move, `unguardedRedo` is always false, and `ResolveMaxWriteThreads` would return exactly the `2 * writeStart` that `Ceiling(writeStart, true)` returns. The comment above the threshold — "growth above the initial write count requires a redo-log backstop; keep the same default as migration, without adding another tuning knob" — is precisely this argument, and it checks out. Since the equality depends on a constant three files away, `ResolveMaxWriteThreads(writeStart, true, false, true)` would state it in code, but the behavior is right as written.

**`Stop`'s sequential per-shard `wait()` + `close(chunkletCompletions)` cannot deadlock**, which was my main worry about moving completion-channel ownership out of the workers' defer. A worker's `chunkletCompletions <- ...` is a blocking send on a bounded channel, so if the coordinator drained shards in order, shard 1's workers could wedge on a full buffer while `Stop` was still waiting on them. It doesn't: `feedbackCoordinator` spawns one forwarder goroutine **per shard**, each a plain `range` over that shard's channel into `mergedCompletions`, all live concurrently until their channel closes. Every shard drains regardless of which one `Stop` is currently joining. Relatedly, `Stop` releasing `a.Lock()` before `workers.wait()` is what keeps `wait()` from blocking a concurrent `SetWriteWorkers` that holds `p.mu` under `a.Lock()` — the one lock-ordering hazard the new per-shard `Lock()` in `SetWriteWorkers` introduces, and it's already handled.

**The extraction does drop observability, but only redundantly — I checked before writing it up.** `SingleTargetApplier.SetWriteWorkers` used to log `"scaled write workers up"/"down"` with from/to at Info, and `workerPool.resize` logs nothing. But the only caller is `copier/autoscaler.go:390`, which logs `"autoscaler adjusting write threads"` with `from`/`to`/`min`/`max` immediately before calling it — strictly more than the removed line, and it now covers the sharded fan-out too. The sharded applier also loses two Debug lines from `writeWorker`'s deleted defer; `Stats().ActiveWorkers` and the emitted gauges still carry the live count.

**`reverseWriteThreads` is genuinely decoupled**, and the `pkg/move/README.md` paragraph describing it is accurate. There is exactly one `&Runner{}` literal in the package (`runner.go:238`), it sets the field from `m.WriteThreads`, and the normalization immediately above guarantees a non-zero value (an explicit `--write-threads 0` warns and becomes the default), so `reversewindow.go:267` can never pass `Threads: 0`. The copy happens at construction, before `setupAutoscaling` overwrites `r.move.WriteThreads` at `autoscale.go:131`, which is what makes the exemption hold for a resumed window as well as a fresh one. `TestReverseWindowPreservesConfiguredWorkers` proves the field is a copy rather than an alias; the ordering argument is what proves the behavior, and it holds.

**Autoscaling's interaction with the fixed connection budget is consistent.** `setupUnderLocks` runs `setupAutoscaling` before `fitReadThreadsToPools`, so `pool.go:48-50` sees the derived numbers: `MaxReadThreads` gets capped to `available` (the per-copy budget after the checksum-snapshot reserve) without raising `--max-connections`, matching the comment. The `max(r.move.Threads, ...)` inside it is defensive rather than load-bearing — once autoscaling has engaged, `readStart = min(readStart, readMax)` already guarantees `Threads <= MaxReadThreads` — but it's the right direction for the disabled case and a ceiling below its start would be uncontrollable anyway.

**Setup failures don't leak monitors.** Every early return in `setupAutoscaling` after the `engaged` sentinel is declared — the `AuroraSetup.Build` error, the partial-coverage bail when a second probe fails after capacity discovery, and the `composite.Open` failure — runs the deferred close over both `signals` and `monitors`. `AuroraResult`'s contract (`Throttlers` empty ⟺ `MonitorDB` nil) means the partial-coverage case has nothing to close for that target and closes the earlier ones, and `multiThrottler.Open` already unwinds its own children on a mid-list failure, so no throttler is left open on any path.

**`pkg/host`'s doc comments state the limits rather than glossing them** — "DNS aliases and proxies cannot be resolved to physical servers here; callers must use consistent direct endpoints" is the assumption that would silently make two schemas on one machine look like two machines, and `docs/move.md` repeats it for operators. `GroupConfigs` dereferences its elements without a nil check, which the comment declares as a precondition ("normalized, non-nil"); `targetHosts()` is the only caller and builds the slice from `target.Config` directly, so it holds today.

**Local run at `8e20bdac`:** `gofmt -l pkg/` clean, `go vet ./pkg/applier/ ./pkg/move/ ./pkg/host/ ./pkg/copier/` clean. The pure-unit subset above passes under `-race`. Everything else in `pkg/applier`, plus `TestMoveAutoscaleNonAurora` and `pkg/copier`'s `TestAutoScalerIntegrationEngaged`, needs a MySQL this box doesn't have — including `TestShardedApplierScaling`, the new 132-line test that is the real coverage for the sharded pool, so treat CI as the authority on those. All cited line numbers verified against the head commit; the diff was read against the true merge base (`a31c0ecc`), not a stale local `main`.

</details>

aparajon (approved):
🤖 Approving — the extraction is the right call and `moveAutoscaleBounds` held under every mutation I could aim at it (7/8 killed). The five findings are all low or nit and none of them block: the one with hard evidence behind it is that `workerPool`'s seal-must-not-retire invariant passes its own unit test when deliberately broken, and the ~20-line test that catches it is in the review comment.

Full review: https://github.com/block/spirit/pull/1216#issuecomment-5572280046

aparajon (approved):
🤖 **Re-approving at `b4a150fd`** — all five findings addressed, and I verified each on the new head rather than taking the commit message for it.

- **1 (test gap) — closed, empirically.** `TestWorkerPoolSealDrainsAcceptedWork` now **kills** the mutation that survived before: `seal()` also calling `resizeLocked(0)` fails it, where it previously passed the whole pure-unit layer. Your comment on the test is better than mine — it names *why* the lifecycle test cannot see it, which is the part a future reader would otherwise re-derive.
- **2 (`throttlerMu`) — complete.** The only remaining references to the field are inside `setThrottler`/`currentThrottler` themselves; all five direct reads now go through the accessor, and the struct comment states the rule.
- **3 (conflated causes) — split as suggested**, with `err.Error()` confined to the Warn arm and the ordinary non-Aurora target dropped to Info, so a plain MySQL fleet stops warning about a signal it was never going to provide.
- **4 (datasync) — recorded, and the citation resolves.** `pkg/datasync/runner.go:1173` is `Throttler: &throttler.Noop{}`, and it is the *only* `Noop` site in the package, so the drift entry is exact rather than approximate. "No equivalent setup moment and no run end to size for" is a better reason than the one I guessed at.
- **5 (stale comment) — fixed**, and it now documents the transient overshoot rather than just dropping the old mechanism.

`gofmt -l pkg/` and `go vet ./pkg/applier/ ./pkg/move/ ./pkg/datasync/` clean at `b4a150fd`; the pure-unit subset passes under `-race`. CI is 15 green with three `build` legs still running and nothing red. The MySQL-gated tests (`TestShardedApplierScaling`, `TestMoveAutoscaleNonAurora`) remain CI's call.

aparajon:
🤖 **CI note** — the `build` leg of *MySQL 9.7 /w docker-compose* went red at `b4a150fd` ([run 34136296013](https://github.com/block/spirit/actions/runs/34136296013/job/101787953865)). I dug into it rather than assuming, and I don't think `b4a150fd` caused it — but it isn't nothing either, so here is what the dump actually says.

**It's a hang, not a slow test.** `pkg/move` hit the 10-minute package timeout at `600.031s`. That same leg ran `pkg/move` in **52.7s** at `8e20bdac` fifty minutes earlier and **39.4s** on `main` at `a31c0ecc`, so this is a 12× overshoot rather than a package sitting near its ceiling. One test was running: `TestMoveForceRecoversCheckpointMissingSource (8m42s)`, blocked inside `testutils.RunSQL` → `db.ExecContext` at `runner_test.go:1808` — the `UPDATE <dstDB>._spirit_checkpoint SET binlog_position = '{}'` that `corrupt()` performs immediately after `checkpointAndStop`. A plain `UPDATE` waiting eight minutes is a row or metadata lock still held by a connection that outlived the stopped runner.

**Three things rule out this commit:**

- **No goroutine in the dump is blocked on a mutex or semaphore** — no `semacquire`, no `sync.(*RWMutex)`, nothing on `throttlerMu`. That directly exonerates the only non-comment change in `runner.go`, the five reads moved onto the `currentThrottler()` accessor.
- **The hung test never enables autoscaling.** Its `Move` literal sets only `SourceDSN`/`TargetDSN`/`Threads: 1`/`WriteThreads: 1`, so `setupAutoscaling` returns at the `!r.move.EnableExperimentalAutoscaling` guard and every line `b4a150fd` changed in `pkg/move/autoscale.go` is unreachable on this path.
- **The commit adds no `pkg/move` test** — the new test is in `pkg/applier` — and the parent commit passed this leg with the identical `pkg/move` test set.

So I'd call it a flake in that leg, most likely a lingering connection on the shared MySQL container holding a lock. I'm deliberately not re-running it: a red check believed spurious is worth reporting, not clearing. Two things worth doing with it rather than just retrying, though — the goroutine dump names the culprit for free, so it's cheap evidence while it's fresh; and if `checkpointAndStop`'s runner can leave a connection holding a lock on the checkpoint table, that's a teardown gap that will keep surfacing as an unrelated timeout in whichever test happens to run next. Per AGENTS.md the answer there is the root cause, not a larger deadline.

My approval at `b4a150fd` stands.

*This note was generated by Claude Code (claude-opus-5).*

PR #1217, opened 2026-09-06
feat: add --max-connections to move and sync

morgo:

Related to #1212; split out of #1216.

Adds `move --max-connections` and `sync --max-connections`, both defaulting to 128 per source/target SQL pool, matching migration. Worker counts no longer expand move's pools, and supplied target handles receive the same limit. Additional connections owned by custom appliers remain outside this budget.

Migration and move share validation in `pkg/dbconn`, which also hosts migration’s two-bound reader-budget helper. Move fits its single concurrency count directly. Move fits checksum readers to the fixed pool, including shared handles and per-table headroom. Datasync uses ordinary reads without pinned checksum snapshots or cutover, so it accepts smaller pools without applying the finite migration's reader reserve.

Validation: race-enabled MySQL tests cover an eight-connection move with 16 write workers, a two-connection sync with 16 write workers through initial copy and continuous INSERT/UPDATE/DELETE replication, checkpoint resume, configuration defaults/validation, and migration budget regressions. Repository-wide lint is clean.

Review follow-up: regression tests pin the lock reserve for five targets sharing one handle and the additional headroom discovered for a 20-table move. The fitter reports only the concurrency it actually computes; documentation explains runtime reserve growth and links the shared helper to its consumers. Targeted race tests and repository-wide lint pass.

### DIFF
--- a/docs/move.md
+++ b/docs/move.md
@@ -83,7 +83,7 @@ Passing `--force` changes that recovery behaviour: instead of failing, Move wipe
 
 Sets the fixed size of each source and target connection pool, matching `migrate`. Read and write workers share these pools; increasing worker counts does not grow them. Dedicated monitoring and advisory-lock connections are separate.
 
-The explicit budget must cover `--threads` plus six connections of checksum headroom. During setup, read concurrency is fitted to the pool, accounting for table statistics queries and checksum snapshots. Write workers may wait for a connection. This is a per-pool limit, not a total across the move.
+The explicit budget must cover `--threads` plus six connections of checksum headroom. During setup, the reserve grows by one connection per source table beyond the first. For example, 20 tables reserve 25 connections, and a smaller pool lowers read concurrency to a minimum of one, allowing background queries to queue. Reusing a connection handle across targets also requires room for each target’s checksum snapshots and locks. Any reduction in read concurrency is logged at INFO. Write workers may wait for a connection. This is a per-pool limit, not a total across the move.
 
 ### reverse-window
 
--- a/docs/move.md
+++ b/docs/move.md
@@ -83,7 +83,7 @@ Passing `--force` changes that recovery behaviour: instead of failing, Move wipe
 
 Sets the fixed size of each source and target connection pool, matching `migrate`. Read and write workers share these pools; increasing worker counts does not grow them. Dedicated monitoring and advisory-lock connections are separate.
 
-The explicit budget must cover `--threads` plus six connections of checksum headroom. During setup, read concurrency are fitted to the pool, accounting for table statistics queries and checksum snapshots. Write workers may wait for a connection. This is a per-pool limit, not a total across the move.
+The explicit budget must cover `--threads` plus six connections of checksum headroom. During setup, read concurrency is fitted to the pool, accounting for table statistics queries and checksum snapshots. Write workers may wait for a connection. This is a per-pool limit, not a total across the move.
 
 ### reverse-window
 
--- a/docs/move.md
+++ b/docs/move.md
@@ -17,6 +17,7 @@ This will copy all tables from the source database to the target database, verif
 - [defer-cutover](#defer-cutover)
 - [defer-secondary-indexes](#defer-secondary-indexes)
 - [force](#force)
+- [max-connections](#max-connections)
 - [reverse-window](#reverse-window)
 - [source-dsn](#source-dsn)
 - [target-chunk-size](#target-chunk-size)
@@ -74,6 +75,15 @@ When set to `true`, target tables are created without secondary indexes. The ind
 When Move cannot resume from an existing checkpoint — for example the checkpoint was written by an incompatible Spirit version, or the target is in a state the resume path cannot validate — it fails rather than risk corrupting a partially-copied target (see [checkpoint-max-age](#checkpoint-max-age)).
 
 Passing `--force` changes that recovery behaviour: instead of failing, Move wipes the target tables and starts the copy fresh, checking for source-side failures before wiping and re-running the full post-setup checks against the cleaned target. Expired checkpoints and malformed or missing source positions are eligible for forced recovery; transient read or connection failures are not. Source and target must refer to different databases, even if different hostnames or credentials are used. Use it only when the target's current contents can safely be discarded.
+
+### max-connections
+
+- Type: Integer
+- Default value: `128`
+
+Sets the fixed size of each source and target connection pool, matching `migrate`. Read and write workers share these pools; increasing worker counts does not grow them. Dedicated monitoring and advisory-lock connections are separate.
+
+The explicit budget must cover `--threads` plus six connections of checksum headroom. During setup, read concurrency are fitted to the pool, accounting for table statistics queries and checksum snapshots. Write workers may wait for a connection. This is a per-pool limit, not a total across the move.
 
 ### reverse-window
 
--- a/docs/sync.md
+++ b/docs/sync.md
@@ -191,3 +191,9 @@ File+offset checkpoints also record the source's `@@server_uuid`. Resume refuses
 coordinates from a different server or an older checkpoint without identity;
 use `--force` to discard the partial copy and start fresh. GTID checkpoints
 remain portable across servers, subject to the normal GTID resume checks.
+
+### max-connections
+
+`--max-connections` sets the fixed size of each source and target SQL pool (default `128`, matching `migrate` and `move`). Worker counts may exceed the budget and wait for connections. It also applies to a supplied target handle; additional connections owned by a custom applier are outside this limit. Zero in the Go API selects the default; negative values are rejected.
+
+Sync’s continuous checker uses ordinary reads rather than pinned snapshot pools, and sync has no cutover. It therefore does not require move’s checksum/cutover headroom or lower configured read concurrency to fit that headroom.
--- a/pkg/datasync/pool_test.go
+++ b/pkg/datasync/pool_test.go
@@ -1,0 +1,27 @@
+package datasync
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/block/spirit/pkg/dbconn"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSyncMaxConnections(t *testing.T) {
+	r, err := NewRunner(&Sync{})
+	require.NoError(t, err)
+	require.Equal(t, dbconn.DefaultMaxConnections, r.sync.MaxConnections)
+	field, ok := reflect.TypeFor[Sync]().FieldByName("MaxConnections")
+	require.True(t, ok)
+	require.Equal(t, "128", field.Tag.Get("default"))
+	_, err = NewRunner(&Sync{MaxConnections: -1})
+	require.ErrorContains(t, err, "--max-connections must be non-negative")
+	// Sync has neither cutover nor pinned checksum snapshots. Workers may
+	// outnumber its connection budget without requiring snapshot headroom.
+	r, err = NewRunner(&Sync{MaxConnections: 2, Threads: 16, WriteThreads: 16})
+	require.NoError(t, err)
+	require.Equal(t, 2, r.sync.MaxConnections)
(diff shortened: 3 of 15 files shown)


aparajon:
🤖 **Adversarial correctness review** — `56f623f4` (+287/-116, 15 files)

The split of the budget question into two halves is the right shape, and it is the part I'd most want preserved in review. `Validate` covers what is knowable at parse time — the configured thread count against a fixed reserve — and `fitReadThreadsToPools` covers what is only knowable after discovery: the table count and whether a programmatic caller reused one `*sql.DB` across targets. That is the same split migration already uses for autoscaling, where the instance-derived thread count only exists after the probe. It also means the fitter is not dead code sitting behind a validator that already rejected its inputs, which was my first suspicion: a 20-table move at `--threads 4 --max-connections 12` passes `Validate` (12 ≥ 4+6) and still needs the runtime fit, because `reserve` has grown to 25 by then.

⭐ **Making `NewRunner` call `Validate` is what lets the `WriteThreads` guard relax safely, and I checked that rather than assuming it.** The old `if m.WriteThreads <= 0` carried a comment explaining that it existed because `MaxOpenConnections = Threads + WriteThreads + 2` — a negative count made that negative, which `SetMaxOpenConns` reads as *unlimited*. This PR removes that arithmetic, so the reason is genuinely gone, and narrowing to `== 0` would still leak a negative through on the programmatic path if nothing rejected it. `move.go:94` does (`--write-threads must be non-negative`), and `NewRunner` now runs it before touching the field. The two halves of that change depend on each other and both landed.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low-med | `pkg/move/pool.go:36` | The `copies+2` reserve floor survives deletion — measured, both new pool tests stay green — and it is not cosmetic: it changes the fitted reader count in 1,585 of the configurations I enumerated |
| 2 | low-med | `pkg/move/pool.go:41` | Move calls the two-bound `ReadBoundsForPool` with the same value for both bounds and a zero reserve, so `start == ceiling` by construction; the log line reports `max_read_threads` as a number that can never differ from `threads` |
| 3 | low-med | `pkg/dbconn/pool_budget.go:31` | The 25-line rationale on `readBoundsForPool` lost every pointer to the consumers that make its central claim checkable |
| 4 | low | `docs/move.md:85` | "read concurrency **are** fitted"; and the stated rule (`--threads` plus six) omits the runtime reserve growth, which is the case an operator will actually hit |
| 5 | low | `pkg/dbconn/pool_budget.go:47` | `ValidateConnectionLimit`'s doc ends with a claim about one specific caller, on a general-purpose helper |

---

### 1 — the one line in the new fitter that nothing pins (low-med)

```go
reserve = max(reserve, copies+2) // snapshot creation holds a lock per occurrence
```

Deleting it leaves both new tests passing:

```
=== M1 drop copies+2 reserve floor
ok  	github.com/block/spirit/pkg/move	0.343s
```

That is not true of its neighbours, which is what makes this worth raising rather than shrugging at. I ran the same treatment on the two lines either side and both die cleanly:

| Mutation | Result |
|---|---|
| `available := max(1, (MaxConnections-reserve)/copies)` → drop the `/copies` | **killed** |
| `if MaxConnections < 2*copies+1` → `< copies+1` | **killed** |
| `reserve = max(reserve, copies+2)` → delete | **survives** |

The reason it survives is arithmetic rather than oversight: in both tests `copies+2` is ≤ the reserve already computed from `minChecksumPhaseReserve`, so the `max` is the identity. `TestMoveConnectionBudget` has `copies == 1` (no sources, no targets) so it compares 3 against 6; `TestMoveSharedSnapshotPoolBudget` reaches `copies == 4` and compares 6 against 6, landing exactly on the boundary where the line still does nothing. Every configuration where the floor binds needs `copies ≥ 5`, and neither test goes there.

I wanted to know whether that gap was worth a test or whether the line is just belt-and-braces, so I enumerated the space rather than guessing — every `copies` in 1..12, `tables` in 1..3, and every `--max-connections` from the guard's own minimum (`2*copies+1`) up to 200, comparing the reader count with and without the line:

```
configs where the copies+2 floor changes the reader count: 1585; first: copies=5 tables=1 max=16 with=1 without=2
```

So it is load-bearing across a wide band — at `copies=5`, one table, `--max-connections 16`, the line is the difference between one read thread and two — and it is also the *only* thing in the function that encodes the claim in its own comment, that snapshot creation holds a lock per occurrence. A third case in `TestMoveSharedSnapshotPoolBudget` with five shared targets and a budget above the guard would pin it, and it is a two-line addition to a test that already builds exactly that shape.

Worth noting what does *not* close this: the existing four-target case is one short of binding, so bumping the existing assertions won't do it — the test needs a `copies ≥ 5` arm with enough budget to clear `2*copies+1`, which is why it reads as a gap rather than a loose assertion.

### 2 — the two-bound helper is called with one bound (low-med)

```go
available := max(1, (r.move.MaxConnections-reserve)/copies)
ceiling := r.move.Threads
start, ceiling := dbconn.ReadBoundsForPool(r.move.Threads, ceiling, available, 0)
```

`ceiling` is initialised to `r.move.Threads` on one line purely to be passed as the second argument on the next, where it is immediately overwritten. Since both bounds go in equal and the reserve is 0, `fit` is `available`, and the helper returns `(min(Threads, available), min(Threads, available))` — the two outputs cannot differ. I confirmed the equivalence by substitution rather than by reading:

```go
start := min(r.move.Threads, available)
ceiling := start
```

builds, passes both pool tests, and drops the `dbconn` import from the file entirely.

This matters for two reasons beyond tidiness. The helper's whole documented reason for taking both bounds is that in migration they are genuinely different numbers — an instance-derived start and an autoscaling ceiling — and that fitting only the ceiling gets floored back up by consumers. Move has no autoscaler and one number, so borrowing the two-bound signature invites the next reader to believe move has that shape too. And the log line:

```go
r.logger.Info("fitting read threads to the connection pool", "threads", start, "max_read_threads", ceiling, ...)
```

emits two fields that are equal in every possible run. An operator reading `threads=1 max_read_threads=1` will reasonably read that as two independently-fitted numbers that happened to coincide, and will draw a conclusion about headroom that the code never computed. Either drop `max_read_threads` from the line, or if move is expected to grow a distinct ceiling later, say so where the temp is created.

### 3 — the moved comment lost the pointers that made it checkable (low-med)

The `readBoundsForPool` that moved from `pkg/migration/runner.go` into `pkg/dbconn` arrived with a 5-line comment in place of a 25-line one. The compression itself is defensible — a shared helper should not carry migration-specific narrative — and the load-bearing sentence did survive: *"fitting just the ceiling fails because consumers floor it back to the start."*

What did not survive is every pointer that let a reader verify that sentence. The original named `SingleChecker.initConnPool` for why the transactions must all open under one lock, and `checksum.NewChecker` and `resolveReadCeiling` as the two consumers that resolve `max(Autoscale.MaxThreads, Concurrency)` and thereby do the flooring. Those are the three places someone changing this function has to look, and now nothing points at them. That is the specific way this comment rots: the claim stays readable and stays unverifiable, so the next person either trusts it or re-derives it from scratch.

One line — "consumers floor the ceiling back to the start; see `checksum.NewChecker` and the copier's `resolveReadCeiling`" — restores the checkability without restoring the length. The deleted paragraph about `Migration.Validate` not covering the autoscaling case is genuinely migration-specific and is right to go.

### 4 — the docs describe the rule the operator won't hit (low)

```
During setup, read concurrency are fitted to the pool
```

Grammar aside, the section states the budget "must cover `--threads` plus six connections of checksum headroom", which is `Validate`'s rule. The rule that will actually surprise someone is the runtime one, where the reserve grows by one per additional source table: a 20-table move needs 25 connections of headroom, not six, and finds out during setup by quietly lowering read concurrency and logging at Info. The following sentence gestures at it ("accounting for table statistics queries") without giving the operator anything to compute with. `+1 per source table beyond the first` is short enough to state outright, and it turns an Info-level surprise into something sizeable in advance.

### 5 — a caller's rule documented on a shared helper (low)

```go
// ValidateConnectionLimit rejects negative limits, which database/sql would
// otherwise interpret as unlimited. Zero means the runner should apply its
// default. Continuous syncs do not require cutover or pinned-snapshot headroom.
func ValidateConnectionLimit(limit int) error {
```

The first two sentences describe the function. The third describes why `datasync` calls this one instead of `ValidateMaxConnections` — true, and useful, but it belongs at the call site in `Sync.Validate`, where the choice is actually made. On the helper it reads as a constraint the helper enforces, which it does not; a third caller with different headroom needs would be perfectly entitled to use this and would find the doc arguing against itself.

---

<details>
<summary><b>Verified</b> — the ordering, the pool ownership, and two attacks that dissolved</summary>

⭐ **Attack that dissolved: `SetPoolSize` on a nil injected target handle.** `datasync/runner.go:315` calls `dbconn.SetPoolSize(r.target.DB, ...)` unconditionally, while `move/runner.go` guards the equivalent loop with `if target.DB != nil` — an asymmetry that looked like a missing nil check, and `SetPoolSize(nil, 2)` does panic (`invalid memory address or nil pointer dereference`, measured). It isn't a regression: nothing validates `Target.DB` in either runner, and on the pre-existing line immediately below, `RequireDifferentDatabase` issues an identity query against the same handle and panics identically. So a nil injected DB was already fatal one line later; this moves the panic, it doesn't create it. The inconsistency is worth resolving in one direction eventually, but not in this PR.

⭐ **Attack that dissolved: the fit landing after its own consumers.** Lowering `r.move.Threads` inside `setupUnderLocks` is only meaningful if nothing has read it yet. All three consumers — `resumeFromCheckpoint`, `newCopy`, `postCopyPhase`, each passing it as `Concurrency` — are called at lines 682, 721/724 and 1319, all downstream of the fit at 653, and `r.sourceTables` (which the reserve depends on) is populated at 623 in `setupDiscovery`, which `Run` calls at 1234 before `setupUnderLocks` at 1289. The ordering holds on every path.

**The target-only `SetPoolSize` loop is complete, not partial.** It covers `r.targets` and not `r.sources`, which looked one-sided until I checked the struct: sources are DSN-only (`SourceDSN`, `SourceDSNs`), so spirit opens them itself with `r.dbConfig.MaxOpenConnections = r.move.MaxConnections` already applied. `Targets []applier.Target` is the only field carrying a caller-supplied `*sql.DB`, so it is the only one that needs the explicit resize.

**The shared helpers are well covered; only move's local one isn't.** Both mutations against `pkg/dbconn/pool_budget.go` die in both dependent packages: weakening `maxConnections < readers+reserve` to `< readers` fails `pkg/move` and `pkg/migration`, and reducing `ReadBoundsForPool` to fitting the ceiling alone fails both as well — which is the migration behaviour the original comment was written to protect, still pinned after the move.

**`defaultThreads = 2` matches the Kong tag.** `Threads` carries `default:"2"`, and `TestMoveConnectionBudget` asserts the `MaxConnections` tag against `DefaultMaxConnections` reflectively, so the constant-vs-tag drift that this pattern usually invites is checked for the new field. The same reflective check is in `TestSyncMaxConnections`.

**`Sync`'s budget is proven end to end where it can be.** `TestSyncE2E` now asserts `MaxOpenConnections == 2` on both the source and target `*sql.DB` *and* both `DBConfig`s, after a run with 16 write threads — so the "workers contend rather than expand the pool" claim is checked against the live handle, not just the config struct.

**Local:** `go build ./...` clean, `go vet` clean on all four changed packages, `gofmt -l` empty, `golangci-lint run ./pkg/move/... ./pkg/dbconn/... ./pkg/datasync/...` → **0 issues**. The tests that don't need a server pass; the failures in `pkg/move` and `pkg/dbconn` are all `Access denied for user 'spirit'@'localhost'` and fail identically on `origin/main`. CI is 18/18 green.

</details>

aparajon (approved):
🤖 Approving — the Validate/runtime split is the right shape and the `WriteThreads` relaxation is safe because `NewRunner` now validates first. Findings are in [my review comment](https://github.com/block/spirit/pull/1217#issuecomment-5556472423); the lead one is a fitter line that no test pins.

PR #1218, opened 2026-09-06
refactor: align move and sync progress and status reporting

morgo:

Related to #1212; split out of #1216. Independent of the connection-budget PR.

Aligns move's structured progress with migration: copy ETA, initial checksum counts, and per-table copy progress. Progress reads one state snapshot, safely accesses the copy chunker during setup/resume, and returns a sentinel-wait summary without logging from a polling API.

Datasync now exposes structured copy ETA and includes copier elapsed time and phase timing in its existing multiline status blocks. All three runners share the table-progress conversion, with stable ordering and source-qualified identifiers for multi-source moves. Datasync's continuous verifier has no finite initial-checksum phase, so it does not invent progress for one.

Per-table row counts use actual settled rows and estimated table cardinality, rather than optimistic keyspace distances. Estimates may change; resumed counts follow the chunker’s watermark semantics. The live polling regression uses a ticker rather than spinning.

Validation: targeted race-enabled tests cover live progress polling during move, migration and datasync, structured fields and phase transitions, multi-source table identity/order, and datasync log formatting. Repository-wide lint is clean.


aparajon:
🤖 **Adversarial correctness review** — `1c328c96` (+357/-76, 12 files)

The central change is the one worth defending in review: `RowsCopied`/`RowsTotal` stop being the optimistic chunker's keyspace-distance counters and become settled rows plus a cardinality estimate. Those two things have the same shape and completely different meanings, and the old code took the first and third returns of `Progress()` — which is keyspace numerator and keyspace denominator for a sparse integer PK — and published them to API consumers under names that promise row counts. `sparseProgressChunker` in the new test makes the difference concrete (50000/100000 keyspace against 2/4 rows), and both halves of the fix are pinned: reverting `CopyRowCounts` to `Progress()`'s copied count fails `pkg/status`, and so does reading the estimate from `tables[len-1]` instead of `tables[0]`.

⭐ **The mutex is correctly scoped, and I checked the boundary rather than trusting the comment.** `chunkerMu` guards `copyChunker` for `Progress()` callers only, which looked under-inclusive: `DumpCheckpoint` reads `r.copyChunker.GetLowWatermark()` on the `continuallyDumpCheckpoint` goroutine under a *different* lock (`checkpointMu`). It's not a gap. `Run` calls `setupUnderLocks` — which is where both publications happen, via `resumeFromCheckpoint` and `newCopy` — strictly before `startBackgroundRoutines` starts `WatchTask`, so the dumper cannot observe the field mid-write. The only genuinely concurrent reader is an external poller, and that is exactly what the mutex and `TestMoveProgressPolledConcurrently` cover.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low-med | `pkg/datasync/runner.go:1677` | The new `Status()` case handles `AnalyzeTable`, a state datasync never enters — copied from move, where it is reachable — and the new test pins the dead half by setting the state directly |
| 2 | low-med | `pkg/move/runner.go:2011` | `Throttle` is the one field where move still differs from migration, and this diff deletes the only comment in the codebase explaining why; datasync's identical comment survives |
| 3 | low | `pkg/datasync/runner.go:1594`, `:1642` | The `//nolint:exhaustive` justification says sync uses three states; this PR adds a case for a fourth two lines below one of them |
| 4 | low | `pkg/status/README.md:131` | `Summary` is the one field the alignment doesn't cover, and the new section describing the alignment doesn't say so |

---

### 1 — half the new `Status()` case is unreachable, and the test makes it look covered (low-med)

```go
case status.RestoreSecondaryIndexes, status.AnalyzeTable:
	b := status.NewBlock("sync status: state=%s total-time=%s state-time=%s", ...)
```

That pair is verbatim `pkg/move/runner.go:1594`, where both states are reachable (`status.Do(status.RestoreSecondaryIndexes, …)` at move:1780, `status.Do(status.AnalyzeTable, …)` at move:1790). Datasync sets exactly three states, and I enumerated them rather than reading past them:

```
pkg/datasync/runner.go:194  r.status.Do(status.CopyRows, ...)
pkg/datasync/runner.go:369  r.status.Do(status.RestoreSecondaryIndexes, ...)
pkg/datasync/runner.go:376  r.status.Do(status.ApplyChangeset, ...)
```

`status.AnalyzeTable` is set in exactly two places repo-wide — `pkg/move/runner.go:1790` and `pkg/migration/runner.go:655` — and neither is on a sync path. So a sync run can never take that half of the case.

What makes it worth a line rather than a shrug is that the new test appears to disprove this:

```go
for _, phase := range []status.State{status.RestoreSecondaryIndexes, status.AnalyzeTable} {
	r.status.Set(phase)
	block = r.Status()
	require.Contains(t, block, "state-time=")
```

`r.status.Set(status.AnalyzeTable)` is a call the sync runner never makes. The assertion passes, coverage marks the line, and the branch reads as required behaviour rather than as speculative. `pkg/status/README.md:133` then repeats the claim in prose — *"`state-time` while restoring indexes or analyzing tables"* — so the documentation asserts a phase sync does not have. Dropping `status.AnalyzeTable` from both the case and the test loop is the whole change; if it is deliberate headroom for a future sync analyze phase, saying so in the case comment costs one line and stops the next reader from re-deriving what I just did.

### 2 — the one place move still diverges from migration is now the one place with no explanation (low-med)

This PR makes move's `Progress()` a close copy of migration's, and I mean that literally rather than approximately — the state-snapshot comment is word-for-word, `"Waiting on Sentinel Table"` is word-for-word, even the `//nolint: exhaustive` spacing matches migration's rather than move's previous `//nolint:exhaustive`. The field lists then differ in exactly one entry:

```go
// pkg/migration/runner.go
	Throttle:     r.throttleStatus(state),
// pkg/move/runner.go — no Throttle field at all
```

The deleted lines are the only thing that ever explained that:

```go
-		// Throttle is deliberately left zero: a move always copies through a
-		// Noop throttler, so there is nothing to report yet. Populate it here
-		// when move learns to throttle (issue #831).
```

The behaviour is unchanged — move still reports no throttling — so this is purely about what a reader can recover. Before the diff, move's `Progress()` looked different from migration's in several ways and one of them carried a note. After it, the two functions are near-identical and the single remaining difference is a silently absent field, which is precisely the shape that reads as an oversight during the next alignment pass. The tell that this is the wrong half to delete is three files over: `pkg/datasync/runner.go` keeps its identical *"Throttle is deliberately left zero"* comment, untouched by this diff. So the PR left the note on the runner where the divergence is obvious and removed it from the one where it is now invisible.

Restoring the two-line version (`Noop` throttler, populate when move learns to throttle) puts it back in the file where someone diffing move against migration will look.

### 3 — the nolint justification is falsified by this diff (low)

```go
switch state { //nolint:exhaustive // sync only uses Initial/CopyRows/ApplyChangeset
```

on both `Progress()` (:1594) and `Status()` (:1642). `RestoreSecondaryIndexes` at :369 has been reachable since before this PR, so the comment was already wrong; what changes here is that the `Status()` switch now carries an explicit case for that fourth state, two cases below the comment denying it exists. A suppression's justification is the only thing a reviewer has to judge whether the suppression is still earned, and this one is now checkably false on its own line. `// sync does not reach the cutover/checksum states` would be true and would keep doing the job.

It also has a consequence for `Progress()`. That switch has no case for `RestoreSecondaryIndexes` and falls to `default`, which is defensible — but the reason a reader would accept the omission without checking is the comment saying the state can't occur.

### 4 — `Summary` is outside the alignment, and the new docs don't scope it (low)

The new README section describes what the three runners now share. Measured on this branch, `Summary` is not among it:

```
move   state=restoreSecondaryIndexes  Summary=""                        Status()!=""=true
sync   state=restoreSecondaryIndexes  Summary="restoreSecondaryIndexes" Status()!=""=true
```

Move matches migration (neither has a `default` arm, so the summary stays empty); datasync keeps `default: summary = state.String()`. Both are reasonable in isolation, and I'm not asking for one to change — move being byte-compatible with migration is clearly the intent. The narrow ask is that the README section, which enumerates what was aligned, is where a reader forms their expectation of what "align move and sync progress and status reporting" delivered. One clause noting that `Summary` remains runner-specific keeps the next person from filing this as a bug. It matters slightly more for the state in finding 1: `RestoreSecondaryIndexes` is the phase this PR just taught `Status()` about, and it is the phase where the two runners' summaries most visibly disagree.

---

<details>
<summary><b>Verified</b> — the mutation results, and four attacks that dissolved</summary>

**Every mutation I ran against the new logic was killed.** Six mutations, each built and run against the pure tests only:

| Mutation | Result |
|---|---|
| `CopyRowCounts` returns `chunker.Progress()`'s copied count again | **killed** (`pkg/status`) |
| `CopyRowCounts` reads the estimate from `tables[len-1]` (the shadow) | **killed** (`pkg/status`) |
| Drop `slices.SortFunc` from `TablesFromChunker` | **killed** — 30 of 30 independent runs |
| Drop `eta = r.copier.GetETAState()` in move | **killed** (`pkg/move`) |
| Drop `eta = cp.GetETAState()` in datasync | **killed** (`pkg/datasync`) |
| Drop the new `RestoreSecondaryIndexes` `Status()` case | **killed** (`pkg/datasync`) |

The sort one is the one I expected to survive. Both ordering assertions use two-element multi-chunkers, and map iteration over two entries is a coin flip, so a single draw would detect the regression half the time — flaky-shaped, which is worse than either extreme. It isn't, because `TestTablesFromChunkerExcludesShadowEstimate` draws four more times across its subtests. I measured rather than reasoned about it: 30 separate `go test -count=1` invocations, 30 failures. Worth knowing that the reliability comes from the shadow test's subtests rather than from the assertion that looks like it owns the ordering.

⭐ **Attack that dissolved: `slices.SortFunc` is not stable, and the PR body says "stable ordering."** Instability can only bite on equal keys, and equal keys are impossible here — the multi path's names are `multiChunker.chunkers`' map keys, unique by construction, and the single path emits one row. `SortStableFunc` would be a real cost for no reachable case.

⭐ **Attack that dissolved: mixed atomic and non-atomic access to `EstimatedRows`.** `CopyRowCounts` does an `atomic.LoadUint64`, which is only meaningful if every writer agrees. Every runtime access does: the sole write is `atomic.StoreUint64` at `tableinfo.go:210`, and `tableinfo.go:560`, `chunker_composite.go:446` and `chunker_optimistic.go:719` all load atomically. The one plain assignment is `chunker_mock.go:53`, inside a constructor before the value is shared.

**Attack that dissolved: same-named tables colliding in the multi-chunker map.** A move with N sources can legitimately carry the same table name more than once, which would have collapsed two entries into one. `NewMultiChunker` keys by `tables[0].QualifiedName()`, which prefixes `Host` when set, and `TestTablesFromChunker` pins it by giving two `items` chunkers distinct hosts and asserting both survive.

**Attack that dissolved: the deleted sentinel log.** Move's `Progress()` used to emit an Info line with `sentinel-wait-time` and `sentinel-max-wait-time` on every call, and this PR replaces it with a summary string. Removing logging from a polled accessor is the right call on its own — the line fired once per poll, at whatever cadence the caller chose — but it only holds if the information survives elsewhere, and it does, in two places: `Runner.Status()` already has a `WaitingOnSentinelTable` case emitting `waiting=` and `max-wait=`, and `status.WatchTask`'s `continuallyDumpStatus` calls it on a fixed interval for the whole sentinel wait (`stopWatchTask` runs only after `sentinel.Wait` returns). `sentinel.Wait` also logs a Warn with `wait-limit` on entry. The README's claim that "periodic logging remains the responsibility of `WatchTask`" checks out.

**`RowsCopied` can exceed `RowsTotal`, and the new docs say so.** The two values are read at different instants and the estimate is an InnoDB statistic that is routinely low, so a consumer dividing them can see over 100%. `CopyRowCounts`' comment states it outright ("not an upper bound") and `TableProgress`' field comments repeat it, which is the right place for it — there is no in-repo consumer of `Progress().Tables`, so this is an external API contract and the contract is now written down. `fraction()` doesn't clamp, but it is only used for `CopyProgress`, not these fields.

**`PerTableProgress` has exactly one caller.** Sorting in `TablesFromChunker` rather than in the chunker looked like it would leave other consumers on map order; there are none outside `pkg/status/tables.go`, so the ordering guarantee is complete where it is.

**Local:** `go build ./...` clean, `go vet` clean on `pkg/status`, `pkg/table`, `pkg/move`, `pkg/datasync`, `pkg/migration`, `gofmt -l` empty across all five, `golangci-lint run` → **0 issues**. `pkg/status` passes in full; the move and datasync progress tests pass, including `TestMoveProgressChunkerPublication` under `-race`. CI is 18/18 green.

</details>

aparajon (approved):
🤖 Approving — the semantic fix is right and well pinned. Four findings in the review comment above, all documentation/consistency level.

PR #1221, opened 2026-09-06
dbconn: take RDS TLS from the driver, and drop rejectReadOnly

morgo:
## What

Bumps `github.com/block/mysql` to merged master and retires the RDS TLS machinery spirit and the driver now both have. **+228 / −3076**, the bulk of it `rdsGlobalBundle.pem`.

Deleted: the embedded bundle, the `x509.CertPool` built from it, and the local `rdsAddr` regex. `NewTLSConfig` and `IsRDSHost` stay as thin forwarders to `mysql.RDSTLSConfig()` / `mysql.IsRDSAddr`, so there is one bundle to refresh instead of two that can drift.

They stay rather than going away because two callers need them and neither is a `database/sql` connection the driver covers:

- `GetTLSConfigForBinlog` builds a `*tls.Config` for the **go-mysql binlog client** — Vitess-lineage code, not the driver.
- **block/schemabot** calls `dbconn.IsRDSHost` to pick a TLS mode (`pkg/mysqlconn/mysqlconn.go:49`).

`initRDSTLS` keeps registering under the name `"rds"` for the same reason: `EnhanceDSNWithTLS` returns DSNs carrying `tls=rds`, and schemabot opens them.

## Two changes here are forced by the bump, not chosen

Both are behaviour. They are the reason this is not a pure deletion.

### 1. `--tls-mode=DISABLED` must now say `tls=false` out loud

The driver's auto-TLS fires when the DSN asked for nothing — which is exactly what `cfg.TLSConfig = ""` meant. So on an RDS host, `DISABLED` would have produced a **TLS connection**: the opposite of what the operator asked for, arriving from a dependency bump, with no error to notice.

Measured, with the old assignment restored:

```
mode=DISABLED  host=db.cxyz.us-east-1.rds.amazonaws.com:3306  tls_param=""  effective_tls=true
mode=DISABLED  host=mysql.internal:3306                       tls_param=""  effective_tls=false
```

It is **RDS-only**, so no test against a local MySQL could have caught it. And the tests that did cover `DISABLED` asserted the DSN string *"should not contain `tls=`"* — the spelling, not the effect. Seven such assertions across four files now go through a `requireNoEffectiveTLS` helper that parses the DSN and checks `cfg.TLS == nil`, and `TestNewDSNDisabledMode` covers an RDS address explicitly.

### 2. `AllowCleartextPasswords` had to grow a second condition in the same breath

It was `cfg.TLSConfig != ""`. With fix 1 in place that reads the new `"false"` as *"TLS is on"* — and starts sending passwords in the clear over a plaintext connection. Fixing the first without the second would have been a straight security regression.

Each guard fails on its own mutation:

```
KILLED  DISABLED writes empty tls= (the original bug)
KILLED  cleartext guard drops the tlsDisabled check
KILLED  NewTLSConfig returns an empty root pool
KILLED  NewCustomTLSConfig ignores the empty-certData fallback
```

## `DBConfig.RejectReadOnly` is deleted

Not a choice either: [block/mysql#5](https://github.com/block/mysql/pull/5) removed `Config.RejectReadOnly` and made the rejection unconditional, so `conn.go:341` stopped compiling the moment the pin moved.

The interesting half is the **opt-out**. `pkg/datasync/runner.go` set `RejectReadOnly = false` for its read-only source, on the grounds that *"sync's source is read-only by design, so it must not fire."* That reasons from the wrong axis: 1290/1792/1836 are raised by **writes**, not by connecting to a read-only server. A pure reader cannot provoke them.

Verified rather than argued — a `super_read_only` MySQL 8.0, as the restricted source user (`SELECT`, `REPLICATION SLAVE/CLIENT`, `RELOAD`):

| statement | result |
|---|---|
| every session `SET` `newDSN` adds (`sql_mode`, `time_zone`, `transaction_isolation`, ...) | OK |
| `SHOW TABLES`, `SHOW CREATE TABLE`, `SELECT` | OK |
| `SHOW MASTER STATUS` | OK |
| **`FLUSH BINARY LOGS`** (the one that looked risky) | OK — and it genuinely rotated the binlog |
| `INSERT` — negative control | **ERROR 1290** |

So the probe can see a failure, and sync's actual workload does not trigger one. If sync ever *does* write to its source that is a bug, and the rejection now surfaces it instead of hiding it. The target side keeps the protection, which is where the Aurora-failover case actually lives.

## Behaviour deltas from delegating the host match

- **Case-insensitive now.** A fix: DNS is case-insensitive, nothing normalizes the host, and an uppercased hostname is the same endpoint.
- **GovCloud and China report `false`.** The bundle holds no roots for either partition, so verifying against it could only ever fail. Reach them with `--tls-certificate-path` and that partition's own bundle.
- **`MinVersion: tls.VersionTLS12`** is now pinned; spirit's config left it to the Go default.

## Not changed, but worth knowing

`NewCustomTLSConfig(nil, ...)` still falls back to the RDS roots for a **non-RDS** host with no `--tls-certificate-path`. That is inherited behaviour and it is close to useless — a non-RDS server will not present an RDS-issued certificate, so `VERIFY_CA` / `VERIFY_IDENTITY` fail by construction. Preserved deliberately: tightening it is a separate behaviour change and does not belong in a dependency bump. Flagged in the doc comment.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l`, `go mod tidy` — all clean
- `go test ./...` against MySQL 8.0.45: **identical pass/fail set to `origin/main`**, no new failures. `pkg/dbconn` fully green.
- The 12 pre-existing failures (FK integration, minimal-RBR, transaction-compression) reproduce unchanged on a stashed `origin/main` tree — local compose configuration, not this branch. One `pkg/status` flake appeared once and passes 5/5 in isolation.

## Related

Same retirement in the other two consumers: [block/vitess#23](https://github.com/block/vitess/pull/23), and strata's to follow. The driver side is [block/mysql#4](https://github.com/block/mysql/pull/4) and [#5](https://github.com/block/mysql/pull/5).


aparajon:
🤖 **Adversarial correctness review** — `f224250b` (+228/-3076, 11 files)

The two forced behaviour changes are correctly identified and correctly paired. `--tls-mode=DISABLED` leaving `TLSConfig` empty would have handed RDS users TLS from a dependency bump, and the `AllowCleartextPasswords` guard genuinely had to move in the same commit — `!= ""` reading `"false"` as "TLS is on" is a security regression, not a cosmetic one. I re-ran both mutations independently and both die:

```
KILLED  cfg.TLSConfig = tlsDisabledConfigName  →  ""
KILLED  cleartext guard drops the tlsDisabled check
KILLED  NewCustomTLSConfig's empty-certData fallback → x509.NewCertPool()
```

The `RejectReadOnly` reasoning is right too, and for the reason given rather than by luck: 1290/1792/1836 come from writes, and the guard was protecting a reader from an error a reader cannot provoke. Retiring it while keeping the target side's protection is the correct shape.

Two findings, both about what the retained surface promises callers.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | med | `conn.go:208-217` | The `"rds"` registration is kept because *"block/schemabot opens them"* — but after the block/mysql switch, schemabot opening them is precisely what fails. Measured: `invalid value / unknown config name: rds`, raised at the `ParseDSN` schemabot calls, before `sql.Open` |
| 2 | low-med | `conn.go:513`, `:556` | `EnhanceDSNWithTLS` is the other DSN producer and did not get fix 1. For `DISABLED` it returns a DSN with no `tls=` at all — the exact input the driver's auto-TLS overrides. Spirit's own caller survives only because `newDSN` runs afterwards |

---

### 1 — the registration's stated beneficiary is the consumer it breaks (med)

```go
// The registration survives the driver's own auto-TLS because the name is part
// of this package's contract, not an internal detail: EnhanceDSNWithTLS returns
// DSNs carrying tls=rds, and block/schemabot opens them.
```

The first half is right and the retention is right. The second half is the part I'd change, because it reads as reassurance and the situation is the opposite.

Measured, with a scratch module importing this branch's `pkg/dbconn` alongside upstream `go-sql-driver/mysql` — the pairing schemabot has today:

```
enhanced          = "u:p@tcp(db.cxyz.us-east-1.rds.amazonaws.com:3306)/app?tls=rds"
upstream.ParseDSN → err = invalid value / unknown config name: rds   (cfg == nil)
sql.Open("mysql") → err = invalid value / unknown config name: rds
```

It fails at `ParseDSN`, which matters for where the error surfaces. `pkg/mysqlconn/mysqlconn.go` in schemabot does exactly this sequence — `dbconn.EnhanceDSNWithTLS` at :157, then `mysql.ParseDSN` at :163 with upstream imported at the top of the file — so the failure arrives as `parse enhanced DSN: invalid value / unknown config name: rds`, on every RDS MySQL connection, before any dial. Its `tlsModeForHost` returns `REQUIRED` for exactly the hosts `IsRDSHost` matches, so the RDS path is the only one affected and it is affected completely.

To be clear about blame: this is not introduced here. Spirit's move to `github.com/block/mysql` did it, and schemabot is fine *today* only because its pin (`v0.16.1-0.20260903162727-fc5f1dfb0a40`) still imports upstream in `pkg/dbconn/conn.go`. The break lands the moment schemabot bumps — which the `block/mysql` wave makes imminent, and which nothing in either repo will catch at compile time. The `DriverName` doc comment two dozen lines up already states the hazard precisely; it is only this comment, and the PR summary, that phrase it as a reason the registration still works.

So: keep the registration, and say what actually has to happen. Something like *"the name is what `EnhanceDSNWithTLS`'s DSNs carry, so a consumer must open them with `DriverName` — block/schemabot currently opens with upstream go-sql-driver and will need to switch when it bumps past this."* That turns a line a schemabot reader would take as an all-clear into the one-line migration note they need.

The alternative worth considering, since `EnhanceDSNWithTLS` exists mainly to hand a DSN to someone else: export the `*tls.Config` (or a `func(*mysql.Config)`) so a consumer can apply TLS without routing through a registry name that only one driver package can resolve. That removes the class rather than documenting it, but it is a bigger change than this PR and I would not hold the bump for it.

### 2 — the DISABLED fix landed on one of the two DSN producers (low-med)

`newDSN` now writes `tls=false`. `addTLSParametersToDSN`, reached through the exported `EnhanceDSNWithTLS`, still has:

```go
case "DISABLED":
	return dsn, nil // No TLS needed
```

and `EnhanceDSNWithTLS` short-circuits above it with `return inputDSN, nil` for the same mode. Both hand back a DSN with no `tls=` — which is the exact condition fix 1 identifies as unsafe, since that is what the driver's auto-TLS treats as "the DSN asked for nothing."

Spirit's one internal caller is safe, and I checked rather than assumed: `runner.go:1276` passes the enhanced replica DSN straight into `NewWithConnectionType`, which calls `newDSN`, which applies the fix. So there is no live bug in this repo.

But the export is documented as producing a DSN to open, the reason it exists is to hand one to a caller that will open it themselves, and for `DISABLED` that caller gets TLS on RDS. Making the two producers agree is small — return `cfg.TLSConfig = tlsDisabledConfigName` from the DISABLED branch instead of the untouched DSN — and it keeps the invariant in one place rather than depending on which entry point a consumer happened to use. If the asymmetry is deliberate (enhance means "add TLS, never remove it"), that is a reasonable position, but then it belongs in the doc comment, because the mode's name promises the opposite.

---

<details>
<summary><b>Verified</b> — the delegation, and three attacks that dissolved</summary>

**The delegation is sound at the pinned version.** `mysql.IsRDSAddr` is `rdsAddr.MatchString(addr) && !govCloudAddr.MatchString(addr)`, both `(?i)`, so the case-insensitivity and GovCloud/China claims in the summary hold exactly as written. `rdsRootCAs` is a `sync.OnceValue` the driver never hands out directly and `RDSTLSConfig` clones — so `NewCustomTLSConfig`'s "a private copy of the pool" comment is true rather than hopeful.

**Attack that dissolved: the private-pool claim resting on an untested dependency property.** `TestValidCertificateBundle` already pins it from spirit's side — `require.NotSame(cfg.RootCAs, other.RootCAs)` plus a non-empty check on the `NewCustomTLSConfig(nil, …)` fallback. If the driver ever regressed to sharing, this test fails here rather than silently widening trust. Good test to have written.

**Attack that dissolved: `--tls-mode=DISABLED` on the replica path.** `replicaDBConfig.TLSMode` does inherit `DISABLED` from the main config, and `EnhanceDSNWithTLS` returns the DSN untouched — but `NewWithConnectionType` runs `newDSN` on it immediately after, so the replica gets `tls=false` like everything else. Finding 2 is the exported contract only.

**Attack that dissolved: the `canRetryError` comment rewrap.** The reflowed line runs long, but `pkg/dbconn/dbconn.go` already carries pre-existing comment lines at 110, 141, 170 and 222 characters, so this is house style rather than a regression.

**Local.** `gofmt -l`, `go build ./...`, `go vet ./pkg/dbconn/ ./pkg/datasync/` clean. `go test ./pkg/dbconn/` green on the TLS/DSN subset that runs without a server (the rest of the package needs MySQL, so I did not use full-package runs to judge mutations). All mutations restored from backup; the worktree is clean at `f224250b`.

**Leak check: clean.** The probe hostname is invented, and the schemabot paths cited are public.

</details>

aparajon (approved):
🤖 Approving — the two forced behaviour changes are correctly identified and correctly paired, and I re-ran the mutations that prove it. Findings are in the review comment above; the one worth acting on before the bump propagates is finding 1 — the `tls=rds` registration is retained on the grounds that schemabot opens those DSNs, and after the block/mysql switch that is exactly the case that fails (`invalid value / unknown config name: rds`, at the `ParseDSN` schemabot calls).

Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for spirit/pull/1221, f224250b.

**Verdict:** 8 findings — 0 blocking, 6 non-blocking (one missed downstream symbol, and four unasserted security properties on paths this PR rewired), 2 suggestions. The DISABLED-mode fix is real and its new tests are individually load-bearing.

### Non-blocking

**1. The PR keeps three symbols for schemabot's sake and misses a fourth.** `GetEmbeddedRDSBundle` is deleted, but `block/schemabot` calls it at `pkg/engine/postgres/postgresconn.go:402` — and the PR body explicitly preserves `NewTLSConfig`, `NewCustomTLSConfig` and `IsRDSHost` *because* schemabot uses them. Low blast radius (schemabot pins a released version and the call site's migration is `nil`), but the consumer note should name it so the bump isn't a surprise compile break.

**2. The binlog TLS path — the stated reason `NewTLSConfig` survives — has no verification-strength assertion.** Rewriting [`conn.go:683`](https://github.com/block/spirit/blob/f224250b/pkg/dbconn/conn.go#L683)'s `VERIFY_IDENTITY` branch to build a `PREFERRED` config (`InsecureSkipVerify: true`) passes the whole suite. The branch *is* under test — deleting `tlsConfig.ServerName = host` at `:708` is caught by every `enabled_*` subtest of `TestGetTLSConfigForBinlogCaseInsensitive` — so the tests reach it and assert only `ServerName`. A regression here drops verification on the replication stream while the `database/sql` pool stays correctly verified, so nothing looks wrong. One `require.False(cfg.InsecureSkipVerify)` per mode closes it.

**3. `initCustomTLS` can register a skip-verify config under the `verify_identity` name undetected.** Making [`conn.go:235`](https://github.com/block/spirit/blob/f224250b/pkg/dbconn/conn.go#L235) register `NewCustomTLSConfig(certData, "PREFERRED")` for all three strict names survives every test: the DSN still spells `tls=verify_identity`, the connection still succeeds, and zero certificate verification happens. This is the exact function whose `certData` sourcing the PR rewired, so it is the one place worth pinning by strength rather than by name.

**4. A 0-byte `--tls-certificate-path` file silently falls back to Amazon's RDS roots.** [`conn.go:126`](https://github.com/block/spirit/blob/f224250b/pkg/dbconn/conn.go#L126) treats empty `certData` as "use the driver's roots", which is correct for the `nil` call but indistinguishable from an operator pointing at a truncated or empty CA file for a *private* CA. That user asked for `VERIFY_IDENTITY` against their own root and gets Amazon's instead, with no error. Erroring on a non-nil-but-empty read would separate the two intents.

**5. The deleted bundle's per-certificate validation is not replaced.** The old `TestValidCertificateBundle` looped every PEM block in `rdsGlobalBundle` and required `x509.ParseCertificate` to succeed; the rewrite asserts only that the pool is non-empty, per-call, and not skip-verify. Spirit will now catch an entirely empty root pool but not a truncated or wrong-partition bundle arriving via a future `block/mysql` bump — which surfaces at connect time in production as an `unknown authority` error naming nothing in this repo. Narrow loss (the old test never proved the roots were *right*), but asserting a known subject such as `Amazon RDS Root 2019 CA` would restore it cheaply.

**6. The datasync source-config relaxation this PR rewrites is unpinned.** Deleting `r.sourceDBConfig.ForceKill = false` at [`datasync/runner.go:265`](https://github.com/block/spirit/blob/f224250b/pkg/datasync/runner.go#L265) passes all 41 datasync tests, E2E included, because the local test MySQL grants `tsandbox` the kill privilege — so the least-privilege path is never exercised. The PR's central claim (everything sync sends the source succeeds against a `super_read_only` MySQL) is verified in the commit message but has no regression test; `sync_test.go:169-170` asserts only `MaxOpenConnections` on these configs.

### General suggestions

**7. `VERIFY_CA`'s hand-rolled chain check is never invoked by a test.** [`conn.go:152`](https://github.com/block/spirit/blob/f224250b/pkg/dbconn/conn.go#L152)'s `VerifyPeerCertificate` closure can be replaced wholesale with `return nil` and the suite stays green — tests assert the struct shape (`InsecureSkipVerify == true`, `RootCAs != nil`) but never the callback's behaviour. Since `VERIFY_CA` is implemented as skip-verify *plus* that callback, a regression there degrades it to no verification at all, strictly worse than `PREFERRED`. Pre-existing, but this PR changes which pool the closure captures.

**8. These security assertions only ever run inside the Docker+MySQL matrix.** The new TLS/DSN tests need no database — they pass standalone — but there is no unit-only job: `linter.yml` is golangci-lint only, and every `go test` lives in `mysql*-docker.yml`. That workflow also has a documentation-only short-circuit, so a PR classified as docs-only skips them entirely. A cheap always-on `go test ./pkg/dbconn/... -run 'TLS|DSN|Certificate|RDS'` job would gate the highest-value assertions in this PR.

The whole PR rests on `applyRDSAutoTLS` being genuinely new in the `block/mysql` pin — if it weren't, deleting the embedded bundle would remove RDS roots with nothing replacing them. Verified against the pinned module source in `GOMODCACHE`: the driver hook is new, and restoring the pre-PR `cfg.TLSConfig = ""` for `--tls-mode=DISABLED` is killed by exactly one test — the RDS subtest of `TestNewDSNDisabledMode`, not the `127.0.0.1` one. That is the bug: before this PR, `DISABLED` against an RDS host had TLS silently forced back on by the driver.

### Verified correct

- Each new assertion in `TestValidCertificateBundle` is individually load-bearing: non-empty pool, `InsecureSkipVerify == false`, and per-call pool distinctness each kill a distinct mutant alone.
- The empty-`certData` fallback is pinned — building an empty pool instead of the driver's roots is caught, and notably `pkg/change/tls_test.go` (which now passes `nil`) does *not* catch it.
- `requireNoEffectiveTLS` is a genuine strengthening over the old `NotContains(dsn, "tls=")`: it catches both `tls=skip-verify` smuggled in as "disabled" and the dropped half of the cleartext guard.
- Removing `require.True(cfg.RejectReadOnly)` and the `rejectReadOnly=true` DSN param is forced by the driver field's deletion, not a weakening — the code would not compile otherwise.
- `IsRDSHost` is pinned in both directions (five tests catch always-false; always-true breaks the pooled integration tests).
- The `PREFERRED` default is pinned by four independent tests.
- No `//go:build` tag and no `t.Skip` in any of the five modified test files, so nothing new is silently excluded.

morgo:
🤖 Addressed. Both reviews converged on the same weak spot from different directions, and it turned out to be a real bug rather than a doc nit.

**@aparajon finding 2 / @Kiran01bm's framing — fixed, and it was the same bug twice.** `EnhanceDSNWithTLS` *and* `addTLSParametersToDSN` both returned an untouched DSN for `DISABLED`, which is exactly the condition the driver reads as "apply RDS auto-TLS." Both now write `tls=false`. Precedence is preserved in both directions, which I checked rather than assumed: an explicit `tls=` in the DSN still outranks the mode, and a nil config is still a no-op — "nothing was said about TLS" is not "no TLS", and only the second should write anything.

The three tests that broke were all asserting the DSN *spelling* (`should not contain "tls="`, `returns original DSN unchanged`). That is precisely why the original bug was invisible, so the same treatment as last round: the table rows pin `tls=false`, the replica test asserts no *effective* TLS.

**@aparajon finding 1 — comment rewritten, and you were right that it read as an all-clear.** It no longer cites schemabot as evidence the registration works, because after the switch schemabot is the thing that breaks. It now states the requirement (open those DSNs with `DriverName`) and notes schemabot is switching to block/mysql alongside this. I confirmed your measurement independently, and the reason it resolves is that once schemabot imports block/mysql it shares *that* driver's registry — so `tls=rds` resolves again. Same-registry, not same-name.

**@Kiran01bm 7 — best finding in either review, fixed.** `VERIFY_CA` is `InsecureSkipVerify` plus the callback, so the callback is the only thing verifying anything, and `return nil` left the struct identical while degrading the mode below `PREFERRED`. It is now exercised against real generated certificates: a leaf from the configured CA is accepted *despite a deliberately mismatched hostname* (that is the mode's whole point), and a leaf from an unknown CA is rejected.

**@Kiran01bm 2, 3, 5 — fixed**, all as strength assertions rather than shape:

| # | now pinned by |
|---|---|
| 2 | `InsecureSkipVerify` false on the binlog path for `REQUIRED` and `VERIFY_IDENTITY` |
| 3 | the registered config read back through `ParseDSN` and checked for its mode's strength |
| 5 | the RDS pool checked for `Amazon RDS Root 2019 CA`, so a future bump cannot hand over a non-empty pool that is not the RDS trust store |

**@Kiran01bm 4 — fixed, and it needed to move.** A configured-but-empty CA file now errors. The check has to sit at the read site in `initCustomTLS`, because by the time `NewCustomTLSConfig` sees `len(certData) == 0` the two intents are indistinguishable — only the *path* separates "named no CA" from "named a CA file that is empty."

**@Kiran01bm 8 — fixed.** Added `.github/workflows/unit-tls.yml`. You were right that every `go test` lived in the `mysql*-docker` matrices behind a docs-only short-circuit. I verified the subset is genuinely database-free (every matched test is a pure DSN/TLS-config function, whole set runs in under a second) and it now runs on every PR with nothing to provision.

Mutations, each killed:

```
KILLED  VERIFY_CA callback body -> return nil
KILLED  VERIFY_IDENTITY InsecureSkipVerify false -> true
KILLED  empty-certificate-file check removed
KILLED  EnhanceDSNWithTLS DISABLED -> untouched DSN
```

**Not doing two, deliberately:**

- **@Kiran01bm 1 (`GetEmbeddedRDSBundle`)** — real break, but it is being fixed rather than accommodated. The one call site is schemabot's *Postgres* trust pool, which wants an `*x509.CertPool`; it now takes `mysql.RDSTLSConfig().RootCAs` directly, so no byte-slice accessor has to survive anywhere. That is in the companion schemabot PR, which also moves it to `block-mysql` and handles the fact that the hot-swap driver keeps upstream go-sql-driver linked.
- **@Kiran01bm 6 (datasync `ForceKill`)** — genuine coverage gap and correctly diagnosed (the local test MySQL grants `tsandbox` the kill privilege, so the least-privilege path never runs). Pinning it needs a restricted user in the test fixture, which is a fixture change rather than a TLS change, and the field is untouched by this PR. Leaving it out rather than half-testing it.

Full `pkg/dbconn`, `pkg/datasync` and `pkg/change` green locally.

aparajon:
🤖 **Re-review** — `f224250b` → `ff47088d` (approval stands)

Both findings are addressed, and the two extras you found on the way are the better half of the commit. I re-ran your mutation table against the **workflow's own filter** rather than the package, so the result also answers "does the new job actually protect this":

```
KILLED  VERIFY_CA callback -> return nil          TestVerifyCAActuallyVerifiesTheChain
                                                  TestVerifyCARejectsRDSChainWhenPrivateCAConfigured
KILLED  EnhanceDSNWithTLS DISABLED -> untouched   TestEnhanceDSNWithTLS
                                                  TestDisabledModeProducesNoTLSFromEitherDSNProducer
KILLED  empty-certificate-file check removed      TestEmptyCertificateFileIsAnError
```

Finding 1 is closed properly rather than reworded — the comment now states the requirement, and `block/schemabot#1320` makes the "switching alongside this change" clause true instead of aspirational. Finding 2 landed on both producers, and separating "the caller said nothing about TLS" (nil config, leave it alone) from "the caller asked for none" (DISABLED, say `tls=false` positively) is the right cut; the DSN-outranks-mode ordering is preserved, and `tls=false` is a value upstream go-sql-driver parses too, so the fix does not widen the blast radius of the one remaining incompatible value.

The empty-CA-file catch is the one I'd have missed. "Named a path to a truncated private CA" and "named no path" are genuinely only distinguishable at the read site, and silently verifying against Amazon's roots on a connection that *succeeds* is the worst shape that class of bug comes in.

One finding on the new workflow.

| # | Sev | Where | What |
|---|---|---|---|
| 1 | low | `.github/workflows/unit-tls.yml:47` | The `-run` filter is case-sensitive, and this package names TLS modes in ALL CAPS — so `Verify` misses `VERIFY_CA` and `Disabled` misses `DISABLED`. Two pure tests are dropped, one of them covering DISABLED DSN behaviour |

### 1 — the filter misses the repo's own naming convention (low)

The comment above it is right about the failure mode and calls the shot: *"a new TLS test whose name this misses is silently unprotected."* That case already exists, because the filter's vocabulary is title-case and the mode names are not:

```
$ go test -list 'TLS|DSN|Certificate|RDS|Verify|Strict|Disabled|Registered' ./pkg/dbconn/   → 44
$ go test -list '(?i)TLS|DSN|Certificate|RDS|Verify|Strict|Disabled|Registered' ./pkg/dbconn/ → 46

gained by (?i):
  TestVERIFY_CAHostnameFlexibility
  TestPREFERREDModeDISABLEDFallback
```

`VERIFY_CA` does not contain `Verify`; `DISABLED` does not contain `Disabled`. Both gained tests are pure `newDSN` assertions needing no server, and `TestPREFERREDModeDISABLEDFallback` covers DISABLED DSN construction — the behaviour this PR changes.

I'm calling it low rather than medium because neither test is *uniquely* protecting anything today: I could not construct a mutation that survives the current filter but dies under `(?i)`, since `TestDisabledModeProducesNoTLSFromEitherDSNProducer` already covers the DISABLED path from both producers. The cost is future coverage, which is exactly what the comment is trying to buy. `(?i)` at the front of the pattern is the whole fix and it is free — I ran it with `-race` and no database:

```
go test -race -count=1 -run 'TLS|DSN|…'      ok  github.com/block/spirit/pkg/dbconn  2.638s
go test -race -count=1 -run '(?i)TLS|DSN|…'  ok  github.com/block/spirit/pkg/dbconn  2.112s
```

Residual after that: `TestPREFERREDModeConfigConsistency` is pure and still missed, because it carries no word from the vocabulary at all. Adding `Preferred|Mode` would pull it in along with the rest of `tls_mode_test.go`, which by the comment's own logic is the safe direction.

<details>
<summary><b>Also verified</b></summary>

**The cache commit is a real fix, not a workaround.** I looked at it expecting a test-masking change and it isn't one: `cache: false` on `setup-go` for a job that also runs on release tags, matching `linter.yml`, on a job that compiles one package. Nothing is skipped.

**Three DISABLED assertions moved from spelling to effect.** That was the actual reason the original bug was invisible — `"should not contain tls="` is satisfied by exactly the DSN that triggers auto-TLS. Asserting `tls=false` and "no effective TLS" is the axis that can fail.

**The RDS-pool test earns its place for the reason given.** Spirit no longer owns the bundle, so "non-empty" is not the property worth pinning — "is actually the Amazon trust store" is, and that is what a bad `block/mysql` bump would break.

`gofmt -l` and `go vet ./pkg/dbconn/` clean at `ff47088d`. All mutations restored from backup; worktree clean.

</details>

aparajon (approved):
🤖 Re-approving on `ff47088d` so the approval points at what is actually there — the previous one was on `f224250b`.

Both findings from the first round are closed, and I re-ran your mutation table against the new workflow's own filter rather than the package, so all three kills also confirm the job protects them. One low finding on that filter in the [re-review](https://github.com/block/spirit/pull/1221#issuecomment-5563516935): it is case-sensitive, so `Verify` misses `VERIFY_CA` and `Disabled` misses `DISABLED`, dropping two pure tests. `(?i)` recovers both and costs nothing.

PR #1226, opened 2026-09-10
fix(dbconn): read tinyint(1) as an integer, not a bool

aparajon:
## Why

A signed `tinyint(1)` never reaches spirit as an integer. The driver maps it to a
Go `bool` by default (`tinyInt1IsBool`), and the copier reads rows back into Go to
build its `INSERT`, so `NewDatum` is handed a `bool` for a column it expects to be
numeric:

```
failed to convert value to datum for column is_active:
  could not convert datum to int64: strconv.ParseInt: parsing "false": invalid syntax
```

A table that merely *contains* a `tinyint(1)` therefore fails any copy-path ALTER,
whichever columns are being altered. It fails closed, so nothing is left
half-applied.

## Where the fix goes

Converting the `bool` to 1/0 inside `NewDatum` is the obvious fix and it is wrong.
The `(1)` in `tinyint(1)` is a *display width*, not a range: the column holds the
whole signed tinyint range, and the driver collapses every non-zero value to
`true`. By the time spirit sees the value the distinction is already gone, so
nothing above the driver can recover it — mapping `true` back to `1` would write
`1` over a stored `2`.

```
Before                              After

 tinyint(1) = 2                      tinyint(1) = 2
       │                                   │
       ▼                                   ▼
 driver: tinyInt1IsBool               driver: tinyInt1IsBool
   (default on)                         disabled
       │                                   │
       ▼                                   ▼
    Go bool true                       Go int64 2
       │                                   │
       ▼                                   ▼
 NewDatum: no bool case              NewDatum: signed arm
       │                                   │
       ▼                                   ▼
   ParseInt("true")                     Datum 2
     → migration fails                     │
                                           ▼
                                   _t_new gets 2
```

So the fix is to stop the coercion at the driver: spirit asks for integers on
every connection it opens. `NewDatum` is left alone — with the driver reporting
integers, a `bool` arriving there would mean an unknown new source of lossy
values, and hitting the `default` arm and failing loudly is the behavior we want.

## What

- `newDSN` applies `TinyInt1IsBool(false)`, so every spirit connection — copier,
  checksum recopy, binlog applier — reads `tinyint(1)` as an integer.
- The conversion error now reports its input. `val` was reassigned by `ParseInt`
  before the message was built, so it always read `value=0` and the offending
  value survived only inside the wrapped `strconv` error.

## Tests

`TestTinyInt1ValuesSurviveCopy` copies a table whose `tinyint(1)` column holds
`0, 1, 2, 127, -128` plus 2000 rows of `3`, and asserts every value is unchanged
afterwards. It distinguishes all three states:

| code | result |
|---|---|
| before this PR | `parsing "false": invalid syntax` — the reported failure |
| coercing bool→1/0 in `NewDatum` | checksum differs on all 3 attempts; migration fails after the full copy |
| this PR | passes |

`TestNewDSNDisablesTinyInt1IsBool` pins the DSN property directly, on both the
TLS-enabled and TLS-disabled paths, since that is the whole safety argument.
`TestDatumConversionErrorReportsValue` covers the error text.

---

*Opened by Claude (Opus 5).*

### DIFF
--- a/pkg/dbconn/conn.go
+++ b/pkg/dbconn/conn.go
@@ -538,6 +538,12 @@ func NewWithConnectionType(inputDSN string, config *DBConfig, connectionType str
 // different one fails at connect time with "invalid value / unknown config
 // name", because TLS registries are per-driver package globals rather than
 // anything the DSN carries.
+//
+// This settles TLS and nothing else. The settings a spirit connection also
+// needs — interpolation, cleartext-password gating, integer tinyint(1) — are
+// applied by newDSN, so they reach a connection only via [New]. Spirit's own
+// replica path passes this result to [NewWithConnectionType] for exactly that
+// reason; a consumer that opens the result directly owns the rest itself.
 func EnhanceDSNWithTLS(inputDSN string, config *DBConfig) (string, error) {
 	// A nil config is "the caller said nothing about TLS", which is not the
 	// same as asking for none: leave the DSN alone and let whatever opens it
--- a/pkg/dbconn/conn.go
+++ b/pkg/dbconn/conn.go
@@ -392,6 +392,16 @@ func newDSN(dsn string, config *DBConfig) (string, error) {
 	// and its default-off are gone — so the blue/green and Aurora-failover
 	// protection spirit used to opt into is now simply how the driver behaves.
 	cfg.InterpolateParams = config.InterpolateParams
+	// Report tinyint(1) as an integer rather than a bool. The (1) is a display
+	// width, not a range — the column holds the whole signed tinyint range —
+	// but the driver maps every signed tinyint(1) to a Go bool by default,
+	// collapsing any non-zero value to true. The copier reads rows back into
+	// Go to build its INSERT, so that collapse would write 1 in place of a
+	// stored 2, and the information is gone by the time spirit sees the value:
+	// no amount of care further up can recover it.
+	if err := cfg.Apply(mysql.TinyInt1IsBool(false)); err != nil {
+		return "", fmt.Errorf("could not disable tinyInt1IsBool: %w", err)
+	}
 	// Allow cleartext password authentication only when the connection is
 	// actually encrypted (required for AWS RDS IAM auth, safe because the
 	// connection uses TLS).
--- a/pkg/dbconn/conn_test.go
+++ b/pkg/dbconn/conn_test.go
@@ -111,6 +111,32 @@ func TestNewDSNAllowNativePasswords(t *testing.T) {
 		"DSN must not contain allowNativePasswords=false")
 }
 
+func TestNewDSNDisablesTinyInt1IsBool(t *testing.T) {
+	// The driver maps a signed tinyint(1) to a Go bool by default, which
+	// collapses every non-zero value to true. The copier reads rows back into
+	// Go to build its INSERT, so spirit must always ask for integers: a stored
+	// 2 would otherwise be written as 1, and the value is unrecoverable by the
+	// time spirit sees it.
+	dsn := "root:password@tcp(127.0.0.1:3306)/test"
+
+	for _, tlsMode := range []string{"PREFERRED", "DISABLED"} {
+		t.Run(tlsMode, func(t *testing.T) {
+			config := NewDBConfig()
+			config.TLSMode = tlsMode
+			resp, err := newDSN(dsn, config)
+			require.NoError(t, err)
+			require.Contains(t, resp, "tinyInt1IsBool=false",
+				"DSN must disable the driver's tinyint(1)-to-bool mapping")
+
+			// Round-trip it: the flag has to survive parsing, since that is how
+			// the driver actually receives it.
+			cfg, err := mysql.ParseDSN(resp)
+			require.NoError(t, err)
+			require.Contains(t, cfg.FormatDSN(), "tinyInt1IsBool=false")
+		})
+	}
+}
+
 func TestNewDSNAllowCleartextPasswords(t *testing.T) {
 	// With TLS enabled (default PREFERRED mode), AllowCleartextPasswords should be true
 	dsn := "root:password@tcp(127.0.0.1:3306)/test"
--- a/pkg/dbconn/conn_test.go
+++ b/pkg/dbconn/conn_test.go
@@ -117,24 +117,72 @@ func TestNewDSNDisablesTinyInt1IsBool(t *testing.T) {
 	// Go to build its INSERT, so spirit must always ask for integers: a stored
 	// 2 would otherwise be written as 1, and the value is unrecoverable by the
 	// time spirit sees it.
-	dsn := "root:password@tcp(127.0.0.1:3306)/test"
+	//
+	// Assert on the presence of tinyInt1IsBool=false rather than the absence of
+	// =true: the driver's default is true, so FormatDSN only writes the
+	// parameter when it is false. An untouched DSN carries no parameter at all,
+	// which reads as "bool mapping on".
+	inputs := []struct {
+		name string
+		dsn  string
+	}{
+		{"unset", "root:password@tcp(127.0.0.1:3306)/test"},
+		// Spirit requires exact values, so it overrides a caller who asked for
+		// the bool mapping rather than deferring to them.
+		{"caller_asked_for_bool", "root:password@tcp(127.0.0.1:3306)/test?tinyInt1IsBool=true"},
+	}
 
 	for _, tlsMode := range []string{"PREFERRED", "DISABLED"} {
-		t.Run(tlsMode, func(t *testing.T) {
-			config := NewDBConfig()
-			config.TLSMode = tlsMode
-			resp, err := newDSN(dsn, config)
-			require.NoError(t, err)
-			require.Contains(t, resp, "tinyInt1IsBool=false",
-				"DSN must disable the driver's tinyint(1)-to-bool mapping")
+		for _, input := range inputs {
+			t.Run(tlsMode+"/"+input.name, func(t *testing.T) {
+				config := NewDBConfig()
+				config.TLSMode = tlsMode
+				resp, err := newDSN(input.dsn, config)
+				require.NoError(t, err)
+				require.Contains(t, resp, "tinyInt1IsBool=false",
+					"DSN must disable the driver's tinyint(1)-to-bool mapping")
 
-			// Round-trip it: the flag has to survive parsing, since that is how
-			// the driver actually receives it.
-			cfg, err := mysql.ParseDSN(resp)
-			require.NoError(t, err)
-			require.Contains(t, cfg.FormatDSN(), "tinyInt1IsBool=false")
-		})
+				// Round-trip it: the flag has to survive parsing, since that is
+				// how the driver actually receives it.
+				cfg, err := mysql.ParseDSN(resp)
+				require.NoError(t, err)
+				require.Contains(t, cfg.FormatDSN(), "tinyInt1IsBool=false")
+			})
+		}
 	}
+}
+
+// TestConnectionReadsTinyInt1AsInteger proves the setting changes what the
+// driver returns, not just what the DSN says. The assertions above check the
+// DSN text; this checks the behavior every connection spirit opens gets, since
+// New is the single door onto newDSN.
+func TestConnectionReadsTinyInt1AsInteger(t *testing.T) {
+	testutils.RunSQL(t, "DROP TABLE IF EXISTS conn_tinyint1_scan")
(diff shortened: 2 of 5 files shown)


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for spirit/pull/1226, a4934fe2.
**Verdict:** 4 findings — 1 blocking (bool→1/0 rewrites out-of-range TINYINT(1)), 2 non-blocking, 1 suggestion.

### Blocking

**Mapping bool→1/0 silently rewrites signed TINYINT(1) columns holding values outside {0,1}.** [datum.go:103](https://github.com/block/spirit/blob/a4934fe2/pkg/table/datum.go#L103) — with `tinyInt1IsBool` defaulting to true and spirit never overriding it, block/mysql collapses a stored `flags = 2` to Go `true` before `NewDatum` sees it, so `_t_new` gets `1`. Previously this hit the `default` arm and aborted the migration; now the checksum flags the chunk and `SingleChecker.replaceChunk` re-reads through the same driver, so the repair never converges and the migration hard-fails after the whole table is copied. The information is destroyed at the driver layer — no fix inside `NewDatum` can recover it; set `tinyInt1IsBool=false` on spirit's DSN in `pkg/dbconn` so the driver returns int64.

### Non-blocking

**The comment justifying the new bool arms names the wrong source.** [datum.go:100](https://github.com/block/spirit/blob/a4934fe2/pkg/table/datum.go#L100) — the binlog decoder never emits bool (go-mysql parses `MYSQL_TYPE_TINY` as int8/uint8, and `DecodeBinlogRow` only rewrites binary-width and ENUM/SET columns); the sole producer of bools is block/mysql on the copier's read path. A maintainer debugging the next TINYINT(1) issue will look in `pkg/change/binlog.go` and find nothing. The same wrong premise is repeated at [datum.go:133](https://github.com/block/spirit/blob/a4934fe2/pkg/table/datum.go#L133), `datum_test.go:187` and `datatype_test.go:1385`.

**`TestBooleanColumnConcurrentDML`'s concurrent-DML apparatus targets a path the fix does not touch.** [datatype_test.go:1447](https://github.com/block/spirit/blob/a4934fe2/pkg/migration/datatype_test.go#L1447) — the regression is reproduced entirely by copying the 2000 seeded rows; the DML goroutine's rows reach the applier as int8, so deleting the goroutine would leave the test failing-on-revert identically. The message `no concurrent INSERTs reached the binlog path — test is vacuous` is wrong on both halves, and the same misdirection appears at lines 1421 and 1434.

### General suggestions

**`require.Positive(t, insertCount)` cannot establish what its comment claims.** [datatype_test.go:1447](https://github.com/block/spirit/blob/a4934fe2/pkg/migration/datatype_test.go#L1447) — the applier deliberately drops rows above the high watermark, and the optimistic chunker's terminal chunk is open-ended (`id >= X`), so concurrently-inserted high-id rows arrive via the copier's re-SELECT regardless of whether the binlog INSERT path ever ran. Counting them therefore cannot detect the vacuity described. Pre-existing pattern (same at lines 932 and 1361), so not introduced here.

Whether the new bool arm can misroute a non-boolean type: `mySQLTypeToDatumTp("tinyint(1)")` → `removeWidth` → `"TINYINT"` → `signedType`, so the driver's bool lands on the arm the fix added, never on `binaryType`/`unknownType`; and no type routed to those arms (VARCHAR/TEXT/JSON/DATETIME/FLOAT) is converted to bool by block/mysql, so the missing bool case there is not a gap.

### Verified correct

- The `original := val` change is a real fix — the error message previously always read `value=0` because `ParseInt`'s zero return overwrote `val` first.
- Both `default` arms are byte-for-byte unchanged, so no parse behaviour changed for int8/uint8/string inputs from the binlog.
- For genuine 0/1 booleans the mapping is right: `Datum.String()` takes the `IsNumeric()` branch and emits a bare `1`/`0`, valid in both the copy and apply statements.
- `d5`/`d6` exercise the tinyint(1) routing end-to-end via `NewDatumFromValue`.
- `Datum.compare`/`Add`/`Range`/`MaxValue` are unaffected — `NewDatum` normalises bool to int64/uint64 before the Datum is constructed.
- The new test follows AGENTS.md conventions: FROM-less `INSERT…SELECT` for `SeedRows`, `t.Parallel()`, testify-free `waitForCopyRows` off the test goroutine.
- The `is_active = is_archived` mismatch query is well-formed for the inserted (1,0)/(0,1) pairs, so a coerce-to-constant regression would be caught.

aparajon:
🤖 Confirmed the blocking finding and moved the fix to the driver layer — thank you, this was right, and the original patch was worse than the bug it fixed.

**Blocking — bool→1/0 rewrites out-of-range TINYINT(1).** Reproduced against the real driver: with `tinyInt1IsBool` at its default, `0`→`false` but `2`, `127` and `-128` all →`true`. Mapping that back to `1` writes `1` over a stored `2`, so the patch turned a loud fail-closed error into a silent rewrite. The checksum does catch it — differences on all 3 attempts, aborting after the whole table is copied — which is exactly the non-convergence you described.

Fixed at the layer you pointed to: `newDSN` now applies `TinyInt1IsBool(false)`, so every spirit connection reads `tinyint(1)` as an integer. Both `case bool` arms are gone; `NewDatum` is back to `main` plus the `original := val` error fix. A bool arriving there now would mean an unknown lossy source, and the `default` arm failing loudly is the behaviour we want.

**Non-blocking — the comment named the wrong source.** Correct: the binlog path (`single_target.go:769`, from `RowImage`) hands over int8/uint8 and never produces a bool; the producer is the driver on the copier's read path (`:499`). Both sites emit the same error string, which is what made the binlog reading look right. All four comments now describe the copier and the driver.

**Non-blocking — the concurrent-DML apparatus tested a path the fix doesn't touch.** Also correct; the DML goroutine is gone. `TestTinyInt1ValuesSurviveCopy` is a plain copy of `0, 1, 2, 127, -128` plus 2000 rows of `3`, which separates all three states: on `main` it reproduces the reported error verbatim, on the bool→1/0 version it fails the checksum, and on this fix it passes.

**Suggestion — `require.Positive(t, insertCount)`.** Removed along with the goroutine. Agreed it could not establish what it claimed; I have left the pre-existing instances at lines 932 and 1361 alone.

Since the safety claim now lives in the DSN, `TestNewDSNDisablesTinyInt1IsBool` pins it directly on both the TLS-enabled and TLS-disabled paths.

`3d269002`.

*Replied by Claude (Opus 5).*

Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for spirit/pull/1226, 3d269002.

Re-review of the delta only: `a4934fe2..3d269002`, this PR's one new commit since the previous review. It moves the tinyint(1) fix down to the DSN (`tinyInt1IsBool=false`) and drops the bool handling from `NewDatum`, which resolves the blocking finding raised there; the branch fast-forwarded, so nothing is left uncovered. Findings from the earlier review are not repeated here.

**Verdict:** clean — approve.

Deleting the `bool` arms from `NewDatum` while flipping `tinyInt1IsBool=false` on the DSN
could have left a live path that still produces a Go `bool` and now falls through to the
default arm.

- Text protocol (copier/checksum): the block/mysql fork parses `fieldTypeTiny` into `int64`
  (packets.go:941) and only then collapses it in `convertTinyInt1ToBool` (rows.go:131-141), so with
  the option off no `bool` is ever produced and
  [NewDatum](https://github.com/block/spirit/blob/3d269002/pkg/table/datum.go)'s `int64` arm passes
  the value through untouched.
- Binlog protocol: go-mysql decodes `MYSQL_TYPE_TINY` to `int8`/`uint8`, never `bool`, and the
  applier's call sites
  ([single_target.go:767](https://github.com/block/spirit/blob/3d269002/pkg/repl/single_target.go#L767),
  [sharded.go:1019](https://github.com/block/spirit/blob/3d269002/pkg/repl/sharded.go#L1019)) feed
  exactly those, which the signed/unsigned default arms already handle.
- Scan destinations: no production code scans a `tinyint(1)` column into a `*bool`, and nothing
  calls `rows.ColumnTypes()`/`ColumnTypeScanType`, the other behaviour the flag drives.

### Verified correct

- Fix is at the right layer: the DSN option is the only place the original `tinyint(1)` value still exists before the driver collapses it.
- [newDSN](https://github.com/block/spirit/blob/3d269002/pkg/dbconn/conn.go#L402) applies the option *after* `mysql.ParseDSN`, so a user-supplied `tinyInt1IsBool=true` is overridden, not honoured.
- DSN round-trip holds — the driver emits the param only when non-default and reads it back in `ParseDSN` — so the new `conn_test.go` assertions are meaningful, not tautological; the tls/dsn CI job is green.
- Coverage is complete: every production `sql.DB` that reads row values comes from `dbconn.New` → `newDSN` ([buffered.go:152](https://github.com/block/spirit/blob/3d269002/pkg/copier/buffered.go#L152), [single.go:552](https://github.com/block/spirit/blob/3d269002/pkg/checksum/single.go#L552), [recopier.go:134](https://github.com/block/spirit/blob/3d269002/pkg/checksum/recopier.go#L134)).
- The only raw `sql.Open` sites ([lint/load.go:21](https://github.com/block/spirit/blob/3d269002/pkg/lint/load.go#L21), [fmt/fmt.go:65](https://github.com/block/spirit/blob/3d269002/pkg/fmt/fmt.go#L65)) read `SHOW CREATE TABLE` only and never surface a `tinyint(1)` value.
- `NewDatumFromValueWithType`'s `[]byte`→`string` path and the `isBit`-gated BIT decode are unaffected: `ct.isBit` is false for `tinyint`.
- Net `datum.go` change vs. base `1ab2595e` is only the error-message fix — capturing `original` before `ParseInt`/`ParseUint` zeroes `val` — a genuine bug fix covered by `TestDatumConversionErrorReportsValue`.

Kiran01bm (approved):
🤖 Approved on Kiran's (@kmuddukrishna) behalf by the scheduled review agent — no blocking findings at 3d269002. See the review comment above; non-blocking findings and suggestions, if any, are not merge gates.

morgo (approved):

Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for spirit/pull/1226, f4e26dbd.

Re-review of the delta only: `3d269002..f4e26dbd`, this PR's one new commit since the previous review. The branch fast-forwarded, so nothing is left uncovered. Findings from the earlier reviews are not repeated here.

**Verdict:** clean — approve.

**The new `caller_asked_for_bool` case would be vacuous if `TinyInt1IsBool(false)` were applied before `ParseDSN`.** It isn't: [`conn.go:402`](https://github.com/block/spirit/blob/f4e26dbd/pkg/dbconn/conn.go#L402) runs `cfg.Apply(mysql.TinyInt1IsBool(false))` unconditionally *after* parsing, so an input DSN of `?tinyInt1IsBool=true` sets the unexported flag true during parse and then gets flipped back to false. The case genuinely proves spirit overrides a caller who asked for bool conversion.

### Verified correct

- `require.Contains(resp, "tinyInt1IsBool=false")` is not vacuous — block/mysql's `FormatDSN` emits the parameter only when the flag is false, so asserting presence-of-false is the right form.
- No duplicate `tinyInt1IsBool` in the round-tripped DSN: the `paramOrder` replay in `FormatDSN` only covers unknown params, and `tinyInt1IsBool` has its own parse case, so it is emitted exactly once.
- `TestConnectionReadsTinyInt1AsInteger` discriminates: block/mysql converts signed `TINYINT(1) NOT NULL` to bool when the flag is on, so `require.IsType(t, int64(0), flags)` fails if the DSN setting is ever dropped.
- Both wire protocols are covered by the same hook — `textRows.Next` and `binaryRows.Next` both call `convertTinyInt1ToBool` — so the test's text-protocol query matches what the copier's prepared/interpolated queries get.
- `require.Len(t, want, seen)` reads oddly (args swapped in intent) but is functionally correct: it fails unless `seen == 5`, so a short read is still caught.
- The test checks `rows.Err()` and closes both `db` and `rows` via `utils.CloseAndLog`.
- The replica path funnels through the same constructor: [`runner.go:1276`](https://github.com/block/spirit/blob/f4e26dbd/pkg/migration/runner.go#L1276) calls `EnhanceDSNWithTLS` and [`runner.go:1285`](https://github.com/block/spirit/blob/f4e26dbd/pkg/migration/runner.go#L1285) passes the result to `NewWithConnectionType`, which re-runs [`newDSN`](https://github.com/block/spirit/blob/f4e26dbd/pkg/dbconn/conn.go#L453).
- The PREFERRED→plaintext fallback keeps the setting: [`conn.go:501`](https://github.com/block/spirit/blob/f4e26dbd/pkg/dbconn/conn.go#L501) rebuilds the DSN via `newDSN` on a config copy rather than patching the raw input DSN.

Kiran01bm (approved):
🤖 Approved on Kiran's (@kmuddukrishna) behalf by the scheduled review agent — no blocking findings at f4e26dbd. See the review comment above; non-blocking findings and suggestions, if any, are not merge gates.

PR #1227, opened 2026-09-11
fix(migration): size thread pools from the instance below the autoscaling minimum

aparajon:
## TL;DR

**On an Aurora instance below `autoscale.MinVCPUs` (4), a spirit copy throttles on its own workers and stops making progress, with no other load on the target at all.** The threads hard-stop compares the copy's thread count against a threshold derived from the instance, but below that gate the pool was not derived — it fell back to the `--threads` / `--write-threads` defaults, constants that describe a much larger box. At 2 vCPUs that is 8 threads against a threshold of 3, and since the autoscaling controller is correctly disabled there, nothing ever sheds a thread. So the copy advances about one chunk per backoff **at any table size**: a few hundred thousand rows that a native index build finishes in about a second can fail to complete inside a caller's deadline of minutes.

The instance-derived sizing this needs already exists and is already correct; it was simply never reached below `MinVCPUs`. This change reaches it — 2 + 1 = 3 at 2 vCPUs, which fits — and leaves the controller off. That repairs the copy for every consumer on such an instance, not for one caller.

## How the two got separated

`--enable-experimental-autoscaling` does two separable jobs. It **derives** each pool's starting size from the instance vCPU count (`autoscale.ReadBounds`, `autoscale.WriteStart` — pure functions, evaluated once at setup), and it **scales** those pools at runtime on the throttler's utilization signal. Below `autoscale.MinVCPUs` (4), a single `break` turned off both.

Keeping the controller off there is right, and this change keeps it off: the utilization signal's denominator is the vCPU count, so at 2 vCPUs one thread is half the whole scale and the controller can only ping-pong 1↔2. The sizing was off only because it sat below the same `break` — which left the two halves of the threads hard-stop's own comparison coming from different inputs. The threshold is derived from the instance (`vCPUs + headroom`); the pool measured against it fell back to `--threads` / `--write-threads`, constants whose defaults describe a much larger box.

A table of a few hundred thousand rows on a 2-vCPU Aurora target, flags left at their default of 4, redo-aware signal (headroom 1):

```
Before

  vCPUs = 2 ──┬──► threshold = 2 + 1 = 3 ─────────────┐
              │                                       │
              └──► 2 < MinVCPUs: break                │
                   │                                  │
                   ▼                                  ▼
                   read 4 + write 4 = 8 ────────►     8 > 3   ──►  throttled at 0 rows/s  ✗

After

  vCPUs = 2 ──┬──► threshold = 2 + 1 = 3 ─────────────┐
              │                                       │
              └──► 2 < MinVCPUs: no controller        │
                   │                                  │
                   ▼                                  ▼
                   read 2 + write 1 = 3 ────────►     3 <= 3  ──►  copy runs to the end  ✓
```

The fix removes the `break`. `sizePoolsFromInstance` now holds both decisions and returns them separately: the derived sizes (always) and `growable` (only at or above `MinVCPUs`, which is what decides whether a read ceiling exists at all). The arithmetic, the client-CPU bound on it, and the drain shape are unchanged — they just run at every instance size now.

**Decided deliberately: the flag owns both thread counts wherever there is an instance to size them from.** `--threads` and `--write-threads` are now ignored at every instance size, not just at or above `MinVCPUs`. Honoring them below `MinVCPUs` would make the outcome depend on a number the caller usually left at its default, and would leave exactly this case broken. Help text updated to match.

This repairs every consumer's copy on an instance under `MinVCPUs`, not one caller's.

<details>
<summary><b>Why an oversized pool stalls rather than slows</b></summary>

The cost of throttling is not proportional to how far over the threshold a pool is. `BlockWait` parks a caller in `blockWaitInterval` steps for up to 60 of them, and the sample it waits on is the copy's own thread count, so the pool size selects one of two regimes:

```
        pool fits (2 + 1 = 3 <= 3)          pool does not fit (4 + 4 = 8 > 3)

        sample = 3  ──► not throttled       sample = 8  ──► throttled
             │                                   │
             ▼                                   ▼
        BlockWait returns                   BlockWait parks, re-checks 60x
             │                                   │  nothing sheds a thread
             ▼                                   ▼  below MinVCPUs
        next chunk immediately              sample still 8 ──► still throttled
             │                                   │
             ▼                                   ▼
        no backoff term at all              one chunk per 60s, every chunk
```

The right-hand column is a fixed point: the copy's own threads are what hold the sample over, and below `MinVCPUs` no controller sheds one, so the state never clears on its own. That is why the symptom is a copy that does not finish rather than one that takes longer, why it is independent of host and storage, and why it does not care how few rows the table holds. On the smallest instance class it is the difference between a few hundred thousand rows finishing in under a second and not finishing inside a caller's deadline of minutes.

The read side is only accidentally right for a consumer that already passes `--threads 2`: `ReadBounds(2)` happens to equal that default. The write side is wrong at every setting — `WriteStart(2)` is 1, against a configured 4 — so such a consumer still overshoots 6 against a threshold of 3 before this change.

</details>

<details>
<summary><b>Residuals, stated honestly</b></summary>

**At 2 vCPUs the fit is exact, with no slack.** The derived total is 3 and the tighter threshold is 3, because `MinReadStartThreads` floors the read side at 2 where the proportional derivation would give less — so the copy consumes the headroom the tighter mode carries for spirit's own sibling connections. A sample that catches one of those mid-query reads 4 and trips the gate, and a long enough copy does see that: over ~500 samples of a 500k-row copy, once, cleared at the next sample. It cannot hold, because the copy's own three threads are under the threshold whatever else is running, so the cost is bounded by one poll interval rather than by the 60-poll backoff. That is why the copy test bounds the sized case at one trip instead of zero: zero holds over a short sample window and is not the property the sizing guarantees.

**The derived pool is also a smaller pool.** One write thread instead of four is less copy parallelism, and on a host with spare cores that is slower. On an instance whose vCPU count sets a threshold of 3 it is the difference between a copy that runs and a copy that spends its life in backoff — and two of those four threads had no core to run on regardless.

**`globalStatusMode` goes from always-throttled to marginal, not fixed.** The two signals carry different self-monitoring headroom, and the looser one is looser precisely because its signal counts spirit's own monitoring threads, which a pool total does not include:

| signal | threshold at 2 vCPUs | copy threads before | copy threads after | monitors also counted | observed after |
| --- | --- | --- | --- | --- | --- |
| `redoAwareMode` (perf_schema) | 3 | 8 | 3 | no | 3 — fits |
| `globalStatusMode` (`Threads_running`) | 4 | 8 | 3 | 1–2 | 4–5 — can still trip |

Fitting under the tightest threshold is necessary on both and sufficient only on the redo-aware one. Closing the `globalStatusMode` gap means either widening `selfMonitoringHeadroom` or excluding spirit's own monitoring connections from that signal, both of which change throttling behavior at every instance size — out of scope here.

**Without the flag, a small instance still oversubscribes.** The Aurora threads throttler is built unconditionally on an Aurora target, independent of `--enable-experimental-autoscaling`, so a 2-vCPU migration with no flag runs its configured 4 + 4 against a threshold of 3–4. Overriding an operator's explicit `--threads` when nothing asked for derived sizing is a much larger behavior change and is not part of this.

**The fit covers every phase that holds a pool, and the change-feed drain is not one.** `FlushBounds` floors concurrency at `MinFlushConcurrency` — the historical `change.DefaultFlushConcurrency`, so the derivation is a no-op below `4xlarge` and does not size it down, and a flush's batches run on their own pool connections rather than through the applier's write workers. Two things keep that out of the fit rather than in violation of it: the number is a cap on the batches *one flush keeps in flight*, not a resident pool, so it holds connections only while there are pending changes; and the drain reads this same signal through `ClientConfig.UnderLoad` (`flushUnderLoad` → `GradualOnly`, and `AuroraThreads` is a `GradualThrottler`), halving its concurrency toward a floor of 1 while the signal is over. So the drain can carry a sample over on a small instance, but only against real concurrent writes, and it narrows itself when it does. Sizing it down from the instance would change behavior that predates any sizing, and would not close the gap either: at 2 vCPUs a copy of 3 leaves no room for a concurrent writer at all.

**`pkg/move` is unaffected.** It has a parallel `MinVCPUs` bail-out, but it installs `throttler.Noop{}` when autoscaling is off, so there is no hard-stop for an oversized pool to lose to. `pkg/datasync` has no autoscaling at all.

</details>

<details>
<summary><b>Tests</b></summary>

- `TestSmallInstancePoolsFitTheThreshold` (`pkg/throttler`, the one package that can see both halves of the comparison) asserts that for every vCPU count below `MinVCPUs`, in **both** throttle modes, the derived pool total fits under `throttleThreshold()`; and pins `MinThreadsThrottleThreshold` against both modes' thresholds across 1–128 vCPUs, with 2 and 3 vCPUs asserted explicitly. The range deliberately stops at `MinVCPUs`: above it the write pool oversubscribes the redo log on purpose and the controller can shed.
- `TestSmallInstancePoolsLeaveNoBackoffTerm` (`pkg/throttler`) asserts the two regimes on the real hard-stop, without a server: a derived pool's own steady-state sample, repeated, leaves `IsThrottled` false and `BlockWait` returning without waiting out one interval, in both modes and at every vCPU count below `MinVCPUs`; a sample one thread over parks the caller until the run is cancelled. This is the consequence of the fit rather than a regression test for it — the arithmetic was always correct, it just was not reached — and it is the claim that holds on any hardware.
- `TestSizePoolsFromInstance` (`pkg/migration`) covers 2 and 3 vCPUs: not growable, no read ceiling, the fit under `throttler.MinThreadsThrottleThreshold`, and the graceful-degradation floors (`MinReadStartThreads`, and `WriteStart`'s floor of 1, so nothing derives to zero threads). Plus the `MinVCPUs` boundary and a client-CPU-bound case.
- `TestSmallestInstancePoolsNeverOutgrowTheHardStop` (`pkg/migration`) is the smallest instance class at 2 vCPUs. It resolves the ceilings the copy actually runs with — `copier.ResolveMaxReadThreads` and `throttler.ResolveMaxWriteThreads`, over both privilege-probe answers, since those are settled after the pools are sized — and asserts that the *widest* the pools can ever reach still fits, not just the size they start at. It also asserts the other half of the comparison directly: `defaultThreads + defaultWriteThreads` does **not** fit under that threshold, so the two sides stay comparable if either the defaults or the derivation moves.
- `TestEveryPhasePoolFitsTheHardStopOnTheSmallestInstance` (`pkg/migration`) enumerates the phases that hold threads open against the target — the copy's readers plus the applier, and the checksum's workers plus the applier a repair writes through — and checks each against the threshold, with the drain excluded and the reason stated, so a new phase has to be added to the table rather than quietly skipping the check.
- `TestSmallInstanceCopyDoesNotTripItsOwnThrottle` (`pkg/migration`) runs the failure rather than computing it. It builds an index on a table of a hundred thousand or so narrow rows behind a throttler that samples the copy's own running threads — the redo-aware signal's predicate, `PROCESSLIST_COMMAND = 'Query'` minus the sampler's own connection, filtered to this migration's database — and hard-stops at `MinThreadsThrottleThreshold(2)`, with production's backoff loop. Both runs copy the same rows under the same gate; the thread counts are the only difference:

  | pools | peak sample | threshold | throttle entries |
  | --- | --- | --- | --- |
  | sized from the instance (2 read + 1 write) | 2–3 | 3 | **0–1** |
  | thread flag defaults (4 read + 4 write) | 5–6 | 3 | 4–11 |

  The simulated part is the instance, not the mechanism: a test server has no `@@innodb_buffer_pool_instances` to read a vCPU count from and no redo-log wait event to subtract (a no-op off Aurora anyway), so the vCPU count is named and the threshold is derived from it by production's own function.

  Neither wall-clock time nor time parked is asserted, in either run, beyond an upper bound on what a fitting pool may pay. Both belong to the host: one with cores to spare refills the pool as soon as a backoff ends, so eight throttled threads still finish, and finish sooner there than three unthrottled ones — and it can clear the gate between one chunk and the next call into `BlockWait`, so that copy trips the gate nine times and is charged nothing at all. Whether the gate is tripped is what a test on any host can pin; what tripping it costs belongs to the target.

  Off the suite, the same harness was run against servers pinned to 2 CPUs to check the wall clock the sizing leaves behind, at roughly 100k and 500k rows. The sized pool tripped the gate at most once per copy and finished in 3.4s and 16.5–22.1s respectively, against a native `ALGORITHM=INPLACE` build of the same index at 0.9s and 3.7–5.8s. The unsized pool tripped it 16–22 times on the same copies, once per chunk in the limit, since each chunk that starts while the signal is over pays `BlockWait`. That is the shape that reaches a caller's deadline of minutes: the trip count scales with the chunk count, and each trip costs up to 60 polls when the signal does not clear.

Reverting the sizing reproduces the defect verbatim: `4 read + 4 write threads at their widest, against a hard-stop that trips above 3` in the resolution test, and in the copy test the sized-from-the-instance run turns into `peak_sample=5 throttle_entries=9`.

One production change supports the last test: an unexported `testThrottler` field on `Migration`, substituted through the same copier-only path as the existing `useTestThrottler` mock, so a test can supply the signal a non-Aurora target cannot produce.

</details>

<details>
<summary><b>Docs</b></summary>

`pkg/autoscale/README.md` and `docs/migrate.md` now say that `MinVCPUs` bars the controller and not the derivations, and the `db.r6g.large` row of the sizing table carries its real numbers instead of a footnote marker. `ReadBounds`' and `FlushBounds`' doc comments no longer require callers to have established `MinVCPUs` first — the bounds are defined at every instance size.

</details>

---

*This PR was prepared by Claude Code (Opus 5).*


JashLal:
🤖 Reviewed the diff and the runner configuration at `2e7d65d215078e1f9ae8662f5fdc535bbc90cc93`.

The sizing change correctly separates deriving fixed pool sizes from enabling the controller. Below MinVCPUs, the derived read/write counts reach the applier, copier, and checksum; autoscaling is disabled and the zero read ceiling resolves to the starting count. The existing client CPU cap and larger-instance growth guards remain in place. I found no production-code blocker in this review.

Suggestion — `pkg/migration/runner_small_instance_test.go:225`: add coverage of the Aurora setup branch itself. This test computes `sizePoolsFromInstance` and passes those values through `WithThreads`/`WithWriteThreads` on a non-Aurora server. It therefore verifies the derived pools' behavior, but would still pass if the original early exit were reintroduced ahead of the helper call in `setupCopierCheckerAndReplClient`. A stubbed Aurora detection/vCPU probe that drives setup would protect the wiring that caused this regression.

The documented global-status monitoring headroom and concurrent drain limitations remain relevant; this does not guarantee zero throttling under all activity. Static review only; I did not run the test suite or an Aurora migration.

JashLal (approved):

Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for spirit/pull/1227, 2e7d65d2.

**Verdict:** 1 finding — 1 non-blocking (stale test-helper doc).

Both finder lenses completed. Adversarial verification was capped at 8 candidates, so 3 lower-ranked candidates were never verified either way and are not reported here.

### Non-blocking

**Stale doc comment on `NewTestRunner` contradicts the new `WithTargetChunkSize` option.** [helpers_test.go:302](https://github.com/block/spirit/blob/2e7d65d2/pkg/migration/helpers_test.go#L302) still says copy chunk sizing uses `table.ChunkerDefaultTarget` and "neither is settable per-test here", but this PR adds [`WithTargetChunkSize`](https://github.com/block/spirit/blob/2e7d65d2/pkg/migration/helpers_test.go#L363) which sets exactly that field, and the PR's own new test passes it into `NewTestRunner`. A test author trusting the comment will hand-roll a `Migration` struct to get a small chunk size instead of using the option.

`sizePoolsFromInstance` replaces the old inline read/write clamping, so a sign error there would silently change production concurrency. Verified algebraically that `p.read` / `p.write` / `p.readCeiling` reproduce the old `min(readStart, clientCeiling)` and `max(min(readCeiling, clientCeiling), readStart)` exactly, and that `hostBound()` is equivalent to the old `cappedRead != readStart || cappedWrite != writeStart` condition — behaviour-preserving above `MinVCPUs`.

### Verified correct

- `ReadBounds` at 2/3/4/96 vCPUs and `WriteStart` at 2/4/96 match every `TestSizePoolsFromInstance` expectation.
- `readCeiling == 0` does not leak a 2x ceiling: `maxRead` falls back to `ResolveMaxReadThreads(Threads, false) = 2`, so the copier takes the `configured > 0` branch.
- With `growable == false`, `autoscaleEnabled` is cleared before maxWrite/copier/checksum wiring, and `ResolveMaxWriteThreads(1,false,false,_) = 1` — no pool can grow on a small instance.
- `FlushBounds(2) == (8, 1000)` equals `change.DefaultFlushConcurrency`/`DefaultBatchSize`, so deriving the flush shape below `MinVCPUs` is a behavioural no-op (only the comment is stale).
- `MinThreadsThrottleThreshold` is the tighter mode at every size (`min(1,2)=1`), matching `throttleThreshold() == vCPUs + mode.headroom`, so the new 1..128 equality loop is sound.
- `setupThrottler` ordering is safe: `useTestThrottler` checked before `testThrottler`, neither installed by default, and it runs after `setupCopierCheckerAndReplClient` so `r.copier` is non-nil at `SetThrottler`.
- `copyThreadThrottler` satisfies `throttler.Throttler` and its poller stops reliably: `Runner.Close` closes the throttler, and LIFO `t.Cleanup` ordering runs `m.Close` before the test DB is dropped.

Kiran01bm (approved):
🤖 Approved on Kiran's (@kmuddukrishna) behalf by the scheduled review agent — no blocking findings at 2e7d65d2. See the review comment above; non-blocking findings and suggestions, if any, are not merge gates.

aparajon:
Closing for now, may re-open if we find we need this later. For now upsizing is the better answer, Aurora users should scale up to at least `t4g.large` if the sort of issues in this PR emerge

PR #1230, opened 2026-09-17
Expose checksum progress and bound hot chunks

morgo:
Adaptive checksum chunks are discovered while a pass runs, so `chunks passed / chunks emitted` looks almost complete even while the walker still has substantial keyspace left. A small number of continuously changing chunks can also cycle through the retry queue forever without reaching the queue-size guard, leaving one pass permanently at its final few chunks.

This change:
- reports stable 0–100% chunker progress;
- includes retry queue depth, hot chunks, in-flight work, mismatches, recopies, and deferred hot chunks in the periodic `sync status` block;
- defers a chunk after ten changing-source observations, marks the pass not clean, and revisits it from a fresh walk next pass. Deferred chunks never count as verified and cannot fire `FirstCleanPass`.

Validation:
- `go test -race ./pkg/checksum ./pkg/datasync -run 'TestStatsReports|TestSyncProgressAndLogFormat|TestHotChunkConverges|TestPermanentlyHotChunkDefersToNextPass' -count=10`
- `golangci-lint run ./pkg/checksum/... ./pkg/datasync/...`

The full checksum package suite additionally requires its local MySQL fixture.


aparajon:
## 🤖 What the new deferral bound and progress estimate are pinned by, and where the fake stands in for the chunker

Reviewed `32e48ec0f9c08d62ad0b7f9e3f667c09088ad340` against its merge base `b056adb81d` — `pkg/checksum/continuous.go` (+94/-7), `pkg/checksum/continuous_test.go` (+75), `pkg/datasync/runner.go` (+16), `pkg/datasync/progress_test.go` (+8). **0 blocking, 4 non-blocking.** `go build ./...` is clean and all 23 fake-driven tests in `continuous_test.go` pass at this head; the `throttle_test.go` failures I saw locally are a missing MySQL, and CI is green. This package has no `docs/invariants.md` to cite against.

Two things are right in ways worth naming, because both look like weakenings on a diff read and are not.

**The `FirstCleanPass` gate got stricter, not looser.** `continuous.go:598` now requires `recopies == 0 && deferredHot == 0`, so a pass that deferred a hot chunk can never fire the signal that gates cutover. Deferral resolves a work item without claiming the chunk passed — `handleResult`'s defer branch (`:1009`) increments only the deferral counter, and `chunksPassedThisPass` and the attempt histogram are both left alone. The ordering in `executeWork` is also correct: the two catch-up checks run before the hot check, so a chunk that converged is never deferred instead.

**The package-doc rewrite at `:45-50` corrects a claim that was already false.** The old text said a permanently hot row "will eventually fill the retry queue; the configurable MaxQueueSize then trips an error". At the base, `enqueueRetry` never refused a retry and `MaxQueueSize` only stalled the walker — the same file said so four hundred lines down ("a soft backpressure threshold rather than a hard cap"). So this PR is not trading an error for a log; it is replacing a stale guarantee with the bound that actually exists.

I also checked whether `hotChunkCount` leaks on the new path, since a deferred entry leaves the dispatcher without being re-enqueued: it does not. The `+1` on enqueue and the `-1` on emit are both gated on `consecutiveSrcChanged >= 2`, and a deferred item was already decremented when it was emitted.

---

## Non-blocking

### 1. The per-pass reset of `HotChunksDeferredThisPass` is now load-bearing for cutover, and nothing pins it

`continuous.go:570` resets the new counter at the top of each pass, alongside the nine existing ones. That line is doing much more work than the others: because `:598` gates `FirstCleanPass` on `deferredHot == 0`, a stale value does not merely misreport a statistic — it withholds the cutover signal permanently, for the life of the process.

Deleting `:570` leaves the whole suite green. Here is what that costs, from the probe below run against that mutation:

```
PassesCompleted:1820337  ChunksPassedThisPass:0  MismatchesThisPass:0  RecopiesThisPass:0
HotChunksDeferredThisPass:1  FirstCleanPassAt:0001-01-01 00:00:00 +0000 UTC
```

1.8 million passes, every one of them substantively clean, and the signal never fires — held down by one count carried over from pass 1. The operator-visible symptom is a verifier that looks perfectly healthy in the `verify` status row while cutover never unblocks, and the only clue is `hot-deferred=1` on a pass that deferred nothing.

`TestMultiplePassesResetCounters` (`continuous_test.go:965`) is the test whose name promises this, and it is worth a look while you are here: its body waits for two passes and then asserts `chunker.resets >= 1`. It pins the chunker's reset, not any counter's, so it would not have caught a missing reset on the nine pre-existing counters either. The new counter is the first one where that gap has a consequence beyond a wrong number.

<details>
<summary>Test that passes on <code>32e48ec</code> and fails with the reset at <code>continuous.go:570</code> removed</summary>

```go
// A chunk that was hot enough to be deferred in one pass, and then settles,
// must not keep the verifier from ever reporting a clean pass: the deferral
// count belongs to the pass that observed it.
func TestDeferredHotChunkDoesNotBlockLaterCleanPass(t *testing.T) {
	chunker := newTestChunker(1)
	cfg := fastConfig()
	cfg.MaxHotAttempts = 3
	c := newTestChecker(t, chunker, cfg,
		func(_ context.Context, _ *table.Chunk, attempt int) (int64, int64, uint64, error) {
			if attempt < 4 {
				return int64(attempt), 0, 1000, nil // source moving, target behind
			}
			return 42, 42, 1000, nil // settled
		},
	)

	stop, _ := runUntil(t, c)
	require.Eventually(t, func() bool { return c.Stats().PassesCompleted >= 1 }, 3*time.Second, time.Millisecond)

	select {
	case <-c.FirstCleanPass():
	case <-time.After(3 * time.Second):
		t.Fatalf("no clean pass after the hot chunk settled; stats=%+v", c.Stats())
	}

	err := stop()
	require.True(t, errors.Is(err, context.Canceled) || err == nil)
}
```

`--- PASS: TestDeferredHotChunkDoesNotBlockLaterCleanPass (0.10s)` at this head. With `c.hotChunksDeferredThisPass.Store(0)` deleted:

```
--- FAIL: TestDeferredHotChunkDoesNotBlockLaterCleanPass (3.10s)
    no clean pass after the hot chunk settled; stats={PassesCompleted:1820337 ... HotChunksDeferredThisPass:1 FirstCleanPassAt:0001-01-01 ...}
```
</details>

The same gap covers `c.inFlight.Store(0)` on the next line (`:571`) — deleting it is also green — though a stale `InFlight` only misreports, so it is the milder half.

### 2. The only test of `ProgressBasisPoints` measures the fake, because the fake makes rows and chunks the same number

`TestStatsReportsChunkerProgress` (`continuous_test.go:989`) drives `chunker.Next()` four times and asserts 2500 then 10000. It passes, but it cannot distinguish the two candidate sources, because `testChunker.Progress()` returns the cursor for both slots (`continuous_test.go:88-92`):

```go
return uint64(c.cursor), uint64(c.cursor), uint64(len(c.chunks))
```

So changing `continuous.go:1182` from the first return value to the second — `progress, _, total :=` → `_, progress, total :=`, i.e. reading chunks emitted instead of rows walked — leaves the entire suite green. Against a production chunker the two are not interchangeable: `chunkerComposite.Progress` (`pkg/table/chunker_composite.go:445`) returns `rowsCopied`, accumulated from `Feedback` (`:345`), against `Ti.EstimatedRows`. A chunker three chunks into four can be a quarter of the way through the rows.

Worth stating alongside that, since the field doc at `:227-232` is careful to call the number an estimate: in the continuous flow the numerator is the sum of **target** row counts (`executeWork` passes `tgtCount` as `actualRows`, and only for fresh-walk reads, which is the right call for sizing feedback) while the denominator is the **source** table's `information_schema` estimate. Those two can disagree by a large factor in either direction, so the value can reach 10000 well before the walk ends — the `min` clamp hides the overshoot, and dropping the clamp is also green — or top out well short of it. That is defensible for a progress hint, but `runner.go:1693` renders it as `progress=%.2f%%`, two decimal places of precision the number does not have. One decimal, or none, would stop an operator reading stalled hundredths as a stalled pass.

<details>
<summary>Test that passes on <code>32e48ec</code> and fails when the estimate reads chunks instead of rows</summary>

```go
// rowsChunksChunker reports rows read and chunks emitted as distinguishable
// values, the way every production chunker does.
type rowsChunksChunker struct {
	*testChunker
	rows, chunksDone, totalRows uint64
}

func (c *rowsChunksChunker) Progress() (uint64, uint64, uint64) {
	return c.rows, c.chunksDone, c.totalRows
}

func TestProgressIsMeasuredInRowsNotChunks(t *testing.T) {
	chunker := &rowsChunksChunker{
		testChunker: newTestChunker(4),
		rows:        250,  // a quarter of the rows
		chunksDone:  3,    // but three of four chunks emitted
		totalRows:   1000,
	}
	c := newTestChecker(t, chunker, fastConfig(),
		func(_ context.Context, _ *table.Chunk, _ int) (int64, int64, uint64, error) {
			return 42, 42, 1000, nil
		},
	)
	require.Equal(t, uint64(2500), c.Stats().ProgressBasisPoints)
}
```

`--- PASS` at this head. With `continuous.go:1182` reading the chunks slot instead:

```
--- FAIL: TestProgressIsMeasuredInRowsNotChunks (0.00s)
    Error: Not equal: expected: 0x9c4  actual: 0x1e
```
</details>

### 3. Nothing pins the bound's exact threshold, and at `MaxHotAttempts = 1` the documented counting is off by one

`continuous.go:922` defers when `item.attempts+1 >= c.cfg.MaxHotAttempts`. Relaxing that `>=` to `>` — one extra observation of every hot chunk before it is deferred — leaves the suite green, because `TestPermanentlyHotChunkDefersToNextPass` (`:361`) asserts that a deferral eventually happens, not when. Since the whole point of the constant is the size of the bound, the assertion worth adding is on the count of reads, not on the deferral alone.

Measuring it also turns up a boundary case. The docs say the same thing twice, at `:124-128` and `:164-168`: "the initial read counts as attempt one", and `MaxHotAttempts` is "the number of observations allowed". Counting actual reads before deferral:

| `MaxHotAttempts` | reads before deferral | docs imply |
|---|---|---|
| 1 | **2** | 1 |
| 2 | 2 | 2 |
| 3 | 3 | 3 |
| 5 | 5 | 5 |

The hot branch is only reachable from a retry, so a value of 1 cannot cost fewer than two reads and behaves identically to 2. `NewContinuousChecker` (`:475-477`) only rejects `<= 0`, so 1 is accepted and silently means 2. Either clamp it to a minimum of 2, or say in the field doc that the bound counts the fresh read plus at least one retry.

### 4. `MismatchesThisPass` keeps an identity the deferral path breaks, and "clean" now means two things

The field doc at `:234-240` states an identity that the PR does not revisit:

> On a clean pass this equals PassedSecondAttemptThisPass + PassedUnder5AttemptsThisPass + PassedUnder10AttemptsThisPass + RecopiesThisPass

A deferred chunk mismatched on its fresh-walk read, so it incremented `mismatchesThisPass`, but it lands in no histogram bucket and is not a recopy. Measured on the PR's own deferral scenario:

```
MismatchesThisPass=1; documented sum=0; HotChunksDeferredThisPass=1; corrected sum=1
```

Whether that makes the doc false depends on which sense of "clean pass" it means, and the file now uses both. Under the `FirstCleanPass` sense the identity survives trivially, since this PR makes a pass with deferrals not clean — but that reading also makes the identity's own `RecopiesThisPass` term dead, because a pass with recopies is not clean either. Under the looser "pass that completed without error" sense, which is the reading that makes the `RecopiesThisPass` term meaningful, the identity is now short by exactly the deferral count. Adding `+ HotChunksDeferredThisPass` and saying which sense is meant fixes both halves.

One naming point on the operator surface, since it is the same word: `runner.go:1693` renders `clean=%d/%d` from `ChunksPassedThisPass`/`ChunksThisPass`. Chunk-level "passed" and pass-level "clean" are different properties here — a chunk repaired by the recopier counts in `ChunksPassedThisPass`, on a pass that by this file's own rule cannot be clean. `passed=` or `verified=` would keep the row from implying the pass is on track to be clean. While that row is being tuned, `WalkerStalls` and `PermanentFailures` are the two counters that say the verifier is falling behind or has given up, and the row carries neither.

aparajon (approved):
🤖 Approving. The bound is the right shape and lands on the safe side of the gate it touches: a deferred chunk is never counted as verified, and `FirstCleanPass` now needs zero deferrals as well as zero recopies, so bounding the pass cannot shorten the path to cutover. The package-doc rewrite also retires a `MaxQueueSize` guarantee the code never had.

Four non-blocking items in [the review](https://github.com/block/spirit/pull/1230#issuecomment-5720220461), the two with tests attached being: the new counter's per-pass reset is now load-bearing for cutover and deleting it stays green (one stale deferral withholds `FirstCleanPass` for the life of the process), and the only test of `ProgressBasisPoints` cannot tell rows from chunks because the test chunker returns the cursor for both.

PR #1232, opened 2026-09-17
lint(rename_column): report a case change apart from a rename

aparajon:
## Summary

`rename_column` reported every new column name as a rename, at error severity, including a name that differs only in case. `ALTER TABLE users CHANGE COLUMN phone PHONE VARCHAR(40)` got the rename message: that a rename cannot be done atomically across application pods, and that ORMs generating column names at compile time break until recompiled. Neither applies, because MySQL resolves the column identifier under either case and no query stops matching.

But the clause is not harmless either. Result-set metadata returns the column in its stored case, so a client that indexes returned rows by column name — in a language whose keys are case-sensitive, which is most of them — stops finding the column under the name it asked for.

So it is reported apart from a rename, with its own message, as a warning: the exposure is narrower and less certain than a name no client resolves at all. A name differing by more than case stays an error with the rename message.

```sql
-- error (no client resolves the old name)
ALTER TABLE users RENAME COLUMN phone TO phone_number;
ALTER TABLE users CHANGE COLUMN phone phone_number VARCHAR(40);

-- warning (queries keep matching; the returned column name changes)
ALTER TABLE users CHANGE COLUMN phone PHONE VARCHAR(40);
ALTER TABLE users RENAME COLUMN phone TO PHONE;
```

Both grammars that can give a column a new name now go through one renderer, so `RENAME COLUMN` gets the same distinction `CHANGE COLUMN` does. Worth flagging for review: `RENAME COLUMN` previously reported every clause as a rename without comparing the two names at all, so this is a behavior change for `RENAME COLUMN phone TO PHONE` specifically. No existing test pinned the `RENAME COLUMN` message, which is why adding the grammar name to it broke nothing.

## Testing

`pkg/lint` passes. Cases cover a case change in both grammars and in both directions, that it is a warning rather than an error, that it does not carry the rename message, and that a name differing by more than case stays an error in both grammars. One pre-existing failure in this package, `TestAutoIncCapacityIntegration`, needs a local MySQL and fails identically on `main`.

---

*This PR was authored by Claude Code (claude-opus-5).*


morgo:
> It changes the case the name is stored in, and every query and ORM mapping that referenced the column still resolves to it.

I don't think this is verifiably true about ORMs. If the ORM retrieves a row into a hash table/associative array indexed by the returned column name, this would cause a breakage if/when that index is case sensitive, which in a lot of languages is likely to be the case.

If we relax it, it should be to a warning with a different error message, never to no flag.

morgo (approved):

PR #1235, opened 2026-09-17
feat(sync): add Aurora target-aware autoscaling

### LINKED ISSUE #1234: Add Aurora-aware autoscaling to datasync
🤖 `spirit sync` currently keeps copy, applier, and continuous-checksum concurrency fixed. Consumers injecting a SingleTargetApplier cannot opt into the Aurora load-aware controllers already available to migrate and move.

Add opt-in target-aware autoscaling across the initial copy and continuous verification, with fixed connection budgets, safe phase transitions and cancellation, and fallback for unsupported targets. Built-in replication should consume the target load signal for adaptive flush narrowing; injected change sources need an equivalent callback. Cover injected appliers and checkpoint resume with regression tests.

morgo:

Closes #1234.

`spirit sync` keeps its copy and verification worker counts fixed, including when a caller injects an applier with an unrelated thread count. Add `--enable-experimental-autoscaling` to use Aurora target load throughout the sync.

- Derive bounded read/write concurrency from target capacity, client CPU capacity, and the fixed connection budget. Non-Aurora, undersized, or unsupported injected targets retain their configured counts.
- Reuse the copier's dual controller during initial copy and the checksum controller during continuous verification. Keep write scaling active for checksum repairs after the copy completes.
- Reconcile injected SingleTargetApplier workers with the derived starting count. Join controllers across phase transitions and on shutdown; close monitor resources on normal and failed setup paths.
- Wire the built-in change feed's adaptive flush narrowing to target load. Expose `Runner.TargetUnderLoad` for injected feeds to wire the same callback; synchronous flush writes are separate from applier copy/repair workers.
- Report target throttling in progress and document the integration contract. Capacity bounds and monitor thresholds are derived at startup; restart after changing target instance size. Independent syncs do not share a global worker budget.

Validation:
- Full `pkg/datasync`, `pkg/copier`, and `pkg/checksum` suites passed with `-race` against local MySQL.
- Additional tests cover monitor cleanup on failure, capacity limits, non-Aurora fallback, controller growth and overload, checksum throttling/cancellation, and actual copy → verification → shutdown → checkpoint resume with an injected 16-worker applier.
- Repository-wide `go build ./...` and `golangci-lint run` passed; the final integration-test addition also passes targeted lint and race testing.

Actual Aurora end-to-end validation remains outstanding. GAP enablement is a separate follow-up.


aparajon:
## 🤖 What the new limiter and the injected-applier guard are pinned against, and where one connection budget is spent twice

Reviewed `6a83c0c415535fcc5eb72ac00c768a4914959a22` — `syncAutoscaleBounds` and the `setupAutoscaling`/`engageAutoscaling` pair, the limiter and throttle wait added to `ContinuousChecker.worker`, `StartWriteAutoscaler` and its join ordering against `applier.Stop`, the copier's new `defer cancelScale(); <-scaleDone`, and the four new test files. Ran seven mutations of the new code against the package tests; five died, two survived with a test each below.

**0 blocking, 4 non-blocking.**

The lifecycle work holds up where it is easiest to get wrong. `engageAutoscaling` owns the monitor on every exit: the `engaged` flag closes both the signal and `MonitorDB` on the `len(Throttlers) == 0` and `Open` failure paths, and because `r.loadSignal` is assigned only after `Open` succeeds (`autoscale.go:100-104`), `Close`'s own `r.loadSignal.Close()` (`runner.go:1566`) cannot double-close it — `TestSyncAutoscaleMonitorOwnership` pins `closes == 1` on both paths. The `stopScaling` defer in `runContinuousChecksum` (`runner.go:541`) is registered *after* the applier-`Stop` defer, so LIFO runs the join first, which is what the comment above it claims. And the checksum side is wired to the read ceiling (`Autoscale.MaxThreads: r.autoscale.MaxReadThreads`, `runner.go:557`), not the write one — the right pool, given the asymmetry `ReadBounds` documents. Five of the seven mutations died on the PR's own tests: dropping the `connections-6` reserve, dropping the read-start budget clamp, `TargetUnderLoad` without `GradualOnly`, removing the post-`BlockWait` cancellation check, and sizing the worker pool off `Concurrency` alone all fail.

`AGENTS.md`'s "Not yet unified" entry said autoscaling was *deliberately* absent from datasync; the PR rewrites that entry rather than leaving it to contradict the code. That is the failure mode of a registry, and it was handled.

### Non-blocking

**1. Mutation survivor: nothing distinguishes the limiter from the goroutine count.** Releasing the permit before the read instead of after leaves the whole suite green:

```go
-			res := c.executeWork(ctx, item)
-			limiter.Release()
+			limiter.Release()
+			res := c.executeWork(ctx, item)
```

`ok github.com/block/spirit/pkg/checksum 0.717s`, and green on five consecutive runs. Under that mutation the permit covers only the dispatch, so all `workers` goroutines read concurrently and `limiter.SetLimit` stops having any effect — the shed the controller performs under target load would be inert while every gauge still reported a reduced count. `TestContinuousAutoscaleConcurrency` cannot see it because its floor and its ceiling are both 3, which is also the goroutine count (`continuous.go:521-523`): the test proves growth happened, not that the limit is what bounds the reads. The distinguishing case is a limit *below* the pool size.

<details>
<summary>Test that fails under that mutation and passes on 6a83c0c4</summary>

```go
// The limiter, not the goroutine count, is what bounds concurrent checks. An
// overloaded target holds the controller at the configured start of 1 while the
// pass still runs the worker pool provisioned for the ceiling, so exactly one
// chunk may be in flight: a permit held only across the dispatch, rather than
// across the read, would let every worker read at once under the load the
// limiter exists to answer.
func TestChecksumLimiterBoundsInFlightReads(t *testing.T) {
	old := csTick
	csTick = 2 * time.Millisecond
	defer func() { csTick = old }()
	cfg := fastConfig()
	cfg.Concurrency = 1
	cfg.Autoscale = AutoscaleConfig{Enabled: true, MaxThreads: 3}
	cfg.Throttler = &gradualStub{util: 1.5} // Overloaded: never grows past start.
	entered := make(chan struct{}, 10)
	c := newTestChecker(t, newTestChunker(20), cfg, func(ctx context.Context, _ *table.Chunk, _ int) (int64, int64, uint64, error) {
		entered <- struct{}{}
		<-ctx.Done()
		return 1, 1, 1, ctx.Err()
	})
	stop, _ := runUntil(t, c)
	defer func() { require.ErrorIs(t, stop(), context.Canceled) }()

	select {
	case <-entered:
	case <-time.After(time.Second):
		t.Fatal("no chunk was read")
	}
	select {
	case <-entered:
		t.Fatal("a second chunk read started while the limit was 1")
	case <-time.After(200 * time.Millisecond):
	}
}
```

```
--- FAIL: TestChecksumLimiterBoundsInFlightReads (0.00s)
    a second chunk read started while the limit was 1
```

Passes on `6a83c0c4`.
</details>

**2. Mutation survivor: the injected-applier guard is pinned only by its type assertion.** Reducing `autoscale.go:41` to `if !ok` — keeping the type check, dropping both the target count and the target identity — leaves every test green, including the DB-backed ones, because `TestSyncAutoscaleInjectedApplierResume` calls `engageAutoscaling` directly and never runs `setupAutoscaling`, while `TestSyncAutoscaleUnsupportedApplier` supplies a type that the assertion alone rejects.

The dropped half is the part the comment at `autoscale.go:37-38` is about. A `*applier.SingleTargetApplier` built over a *different* target passes the mutated guard, and then the write controller grows that applier's pool from a signal that describes a database it never writes to — the one case the guard exists to refuse.

<details>
<summary>Test that fails under that mutation and passes on 6a83c0c4</summary>

```go
// An injected SingleTargetApplier that writes to a database other than the one
// the monitor polls must not enable autoscaling: the load signal describes the
// monitored target, so scaling the applier on it would grow writes against a
// server whose load is never measured. The applier is the right type, so the
// type assertion alone cannot reject it.
func TestInjectedApplierMustUseMonitoredTarget(t *testing.T) {
	elsewhere := applier.Target{DB: &sql.DB{}}
	a, err := applier.NewSingleTargetApplier(elsewhere, applier.NewApplierDefaultConfig())
	require.NoError(t, err)

	r, err := NewRunner(&Sync{Threads: 3, WriteThreads: 5, EnableExperimentalAutoscaling: true, Applier: a})
	require.NoError(t, err)
	var logs bytes.Buffer
	r.SetLogger(slog.New(slog.NewTextHandler(&logs, nil)))
	// A monitored target that is not the applier's, with a config present so
	// the target-identity guard is the only thing that can reject it. The DSN
	// is never dialled on this path; reaching the Aurora probe means the guard
	// let a foreign applier through.
	const monitoredDSN = "u:p@tcp(127.0.0.1:1)/monitored"
	cfg, err := mysql.ParseDSN(monitoredDSN)
	require.NoError(t, err)
	monitored, err := sql.Open(dbconn.DriverName, monitoredDSN)
	require.NoError(t, err)
	t.Cleanup(func() { _ = monitored.Close() })
	r.target = applier.Target{DB: monitored, Config: cfg}

	require.NoError(t, r.setupAutoscaling(t.Context()))
	require.False(t, r.autoscale.Enabled)
	require.Equal(t, 3, r.sync.Threads)
	require.Equal(t, 5, r.sync.WriteThreads)
	require.Contains(t, logs.String(), "injected applier must use the monitored single target")
}
```

```
--- FAIL: TestInjectedApplierMustUseMonitoredTarget (0.00s)
    "...msg=\"sync autoscaling disabled: target detection failed\" error=\"...dial tcp 127.0.0.1:1: connect: connection refused\"\n"
    does not contain "injected applier must use the monitored single target"
```

Passes on `6a83c0c4`. The unreachable DSN is what turns the bypass into an assertion rather than a panic.
</details>

**3. The per-pool connection reserve is spent twice on the target pool, but only during continuous verification.** `syncAutoscaleBounds` derives one `budget` (`autoscale.go:23`) and clamps the read ceiling and the write ceiling against it independently (`:28-29`), with the comment reading it as a per-pool figure — "in each fixed SQL pool". That holds during the initial copy, where reads go to the source and writes to the target. It stops holding in the phase this PR newly scales: `readChunkCRC` issues the source and target queries in parallel (`continuous.go:1179-1185`), so each in-flight check holds a **target** connection as well, and `StartWriteAutoscaler` keeps up to `MaxThreads` repair workers on that same pool.

So the target pool can be asked for up to `MaxReadThreads + MaxThreads` connections against a pool of `--max-connections`, where each term alone is allowed to reach `connections-6`. On a 192-vCPU target with 8 client cores and the default 128 connections that is 96 + 122 against 128 — and the six the comment reserves for replication, checkpoints, and metadata queries are exactly what the overshoot consumes. Nothing is corrupted: callers block in `sql.DB` waiting for a connection. But the feed and the checkpoint writer are the ones that starve, which is the opposite of the reserve's purpose, and the symptom (a sync that goes quiet on lag while the checksum keeps passing) does not point at a connection budget. Worth either splitting the budget across the two consumers that share the target pool, or saying in the comment that the sum is deliberate and bounded elsewhere.

**4. The write ceiling is derived without the rule that owns it, and the derived counts overwrite the caller's.** `autoscale.go:28` calls `autoscale.Ceiling(writeStart, true)` directly. The `true` is the "guarded" answer `throttler.ResolveMaxWriteThreads` computes from `redoAware && !commitLatencyEnabled` (`aurora_threads.go:200-203`), and `WriteStart`'s own doc says that rule "lives with the throttler". It is correct today only because `CommitLatencyThreshold` is the hardcoded `100 * time.Millisecond` at `autoscale.go:75`, so the backstop is always armed; the moment that becomes an option, as `--max-commit-latency` already is for migration, the ceiling silently oversubscribes the redo log unguarded. `pkg/move/autoscale.go:47` has the same shape, so this PR is following move rather than diverging from it — which makes it a good moment to either route both through `ResolveMaxWriteThreads` (it needs the two booleans, so the ceiling would move after `Build`) or narrow the claim in `WriteStart`'s comment.

Separately, `autoscale.go:108` assigns through to `r.sync.Threads` / `r.sync.WriteThreads`, and `Sync` is the caller's struct. A programmatic caller that builds a second `Runner` from the same `Sync` after a restart no longer has the numbers it configured — it has the previous run's derived ones, which is the wrong baseline if the second run does not qualify for scaling. The log line says the counts were overridden, but the override outlives the `Runner` that made it.

### Notes

`block/spirit` has no `docs/invariants.md`, so there is no registry entry to cite; `AGENTS.md`'s unified/not-yet-unified section is the nearest thing and the PR updates it. All 19 checks are passing on this head.

The DB-backed tests could not be exercised here — this host has MySQL listening but no `spirit` grant, so `TestSyncAutoscaleNonAurora`, `TestSyncAutoscaleMonitorOwnership`, `TestSyncAutoscaleInjectedApplierResume` and `TestWriteAutoscalerLifecycle` all fail on `Error 1045 (28000): Access denied for user 'spirit'@'localhost'`. Everything claimed above was run: `pkg/checksum`'s continuous tests, the connection-free `pkg/datasync` tests, `TestAutoscalerIfEnabled_Gating`, both proposed tests, and all seven mutations. The two survivors were confirmed against the tests that can run; the reasoning about `TestSyncAutoscaleInjectedApplierResume` not covering the guard is from reading it, since that one needs the grant.

aparajon (approved):
## 🤖 Stamped with comment

Approving `6a83c0c415535fcc5eb72ac00c768a4914959a22`. **0 blocking, 4 non-blocking** — the detail is in [the review comment](https://github.com/block/spirit/pull/1235#issuecomment-5722046817).

Nothing here can return a wrong answer about convergence: the checksum still compares whole signatures, a confirmed divergence still goes through the recopier, and every path that cannot establish the target's capacity or the applier's identity disables scaling and keeps the configured counts. The monitor's ownership is correct on all three exit paths, and the controllers are joined before the applier stops.

The four non-blocking items are two unpinned behaviours with a test each (the limiter's permit must span the read, not the dispatch; the injected applier must be the monitored target), one arithmetic note (the read and write ceilings each spend the whole per-pool budget, and both land on the target pool during continuous verification), and one rule-location note (`autoscale.Ceiling(writeStart, true)` reimplements what `ResolveMaxWriteThreads` owns).

morgo:
🤖 Addressed the review follow-ups:

- Added regressions that hold the checksum limiter below the worker ceiling and reject a single-target applier using a different database.
- Partitioned the target connection budget between checksum reads and repair writes, reserving the derived flush width and checkpoint/metadata headroom. Added bounds tests across target sizes and small pools.
- Routed the write ceiling through `ResolveMaxWriteThreads`, with the commit-latency threshold shared by setup and the ceiling policy. `NewRunner` now copies the caller's configuration before applying defaults or derived counts.
- Added pre-start worker sizing for injected single-target appliers, covering both initial start and restart without briefly spawning the constructor's larger pool.
- Applied capacity-derived flush concurrency and batch size to the built-in feed on fresh and resumed runs; documented the corresponding injected-feed configuration.

Validation: complete datasync, checksum, and applier suites passed with `-race` against local MySQL; lint passed for all three packages.

PR #1240, opened 2026-09-17
fix(table): strip the auto-increment counter without reformatting

aparajon:
## Why

`table.StripAutoIncrement` removes `AUTO_INCREMENT=N` from a `CREATE TABLE` statement. That number is a live counter on one server, so it should not be saved into DDL kept in a file.

#1210 rewrote it from a regex to parse / delete / `Restore`, fixing a real bug: the regex also matched `auto_increment=999` inside a table `COMMENT`. But `Restore` re-renders from the parse tree, and it is not a pretty-printer:

```sql
-- in, from SHOW CREATE TABLE
) ENGINE=InnoDB AUTO_INCREMENT=1000 DEFAULT CHARSET=utf8mb4

-- out: also one line, uppercase types, _UTF8MB4 on string defaults
) ENGINE = InnoDB DEFAULT CHARACTER SET = UTF8MB4
```

Only a table that *has* a counter takes that path, so one `LoadSchemaFromDB` call returns two formats: re-rendered one-liners for the busy tables, raw `SHOW CREATE TABLE` for the rest. Saved to disk, every busy table is a diff.

The doc comment also pointed at the wrong tool, which this fixes:

| If you want to | Use |
| --- | --- |
| compare two schemas | `statement.Diff` — already skips the counter |
| clean a schema file against a live server | `pkg/fmt` |
| strip the counter from DDL text, no server in hand | `StripAutoIncrement` |

## What

The parser records where the counter sits. `pkg/table` cuts that span and leaves every other byte alone.

```
) ENGINE=InnoDB FORCE AUTO_INCREMENT = 100 DEFAULT CHARSET=utf8mb4
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
                └─ OriginTextPosition, for len(OriginalText()) bytes
```

- **`feat(parser)`** — the `AUTO_INCREMENT` rule sets both. Table options used to report `0`.
- **`fix(table)`** — no counter, return the input unchanged; one counter, cut its span. Nothing here searches the text, so the keyword inside a `COMMENT`, a column default or a quoted identifier is never a candidate. The cut is re-parsed and compared with the original, so a bad one errors rather than returning mangled DDL — as does input that will not parse.

## Review notes

- **The end matters as much as the start.** An option can carry a `FORCE` prefix, odd spacing, or a comment before the value. An earlier version of this branch measured the extent by hand, left the `FORCE` behind, and could not handle `AUTO_INCREMENT=/* reset me */ 9`.
- **goyacc gives a reduce action no position of its own.** `$$` starts as a copy of `$1`, so a nonterminal's offset is really its first symbol's. Setting the span in `TableOptionList` looked tidier and made a `COMMENT` and an `AUTO_INCREMENT` option report the same offset. Both ends now sit on a terminal; `ast.TableOption` documents the limit.
- **The grammar is unchanged**, action body only: parse tables identical at 3145904 entries, 0 conflicts.

## Testing

Beyond CI: `./pkg/parser/...` with and without `-tags singleversion`, plus `./pkg/table/...`, `./pkg/statement/...` and `./pkg/fmt/...` against MySQL 8.0.45.

The pre-#1210 code still returns the re-rendered one-liner on the integration fixture, and the new assertions fail on it. `TestStripAutoIncrementUsesTheParserSpanPastADecoy` pins the decoy inside the `COMMENT` *earlier* in the text than the real counter, so the answer has to come from the span and not from a lucky search.

## After merge

An internal Block tool that writes schema files carries this logic and hit the bug. It picks up a new Spirit version and drops its local copy.


morgo:
I think I'd like to take a closer look at this. Now we control the parser it has a feature where you can set Restore context flags. I would like to see if it makes sense to move the normalizers we have + this under it.

morgo:
I took a closer look and no longer think the pkg/statement normalizers should move into pkg/parser. Restore flags are serialization policy, while the normalizers model MySQL’s stored schema form specifically for structured comparison. A restore flag also cannot preserve the original bytes around the removed option. The split here looks right: the parser owns the source span, and pkg/table owns the policy of removing the instance-specific counter.

aparajon:
🤖 Agreed, and thanks for digging into it. I put your reasoning in the `StripAutoIncrement` doc comment so the next reader does not have to re-derive it:

```go
// This is deliberately not a format.RestoreFlag. A restore flag is
// serialization policy: it decides how the parse tree gets written out, and
// everything it touches goes through Restore, which re-renders the whole
// statement. Re-rendering is the bug this function exists to avoid. Keeping the
// bytes around the removed option means editing the source text, which only a
// caller holding that text can do.
```

Also rebased past #1241, which moved the `StripAutoIncrement` call into `LoadSchemaAndExcludedTablesFromDB`. The error propagates through the three-value return now. CI is rerunning on the new head.

Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for spirit/pull/1240, 80e084cd.

**Verdict:** 3 findings — 3 non-blocking (all in the new post-strip verification path).

### Non-blocking

**The new post-strip verification cannot render `ENGINE_ATTRIBUTE`, so a correct strip becomes a hard error that aborts the whole schema load.** [auto_increment.go:93](https://github.com/block/spirit/blob/80e084cd/pkg/table/auto_increment.go#L93) re-renders the parsed statement through `ast.TableOption.Restore`, and [ast/ddl.go:2444](https://github.com/block/spirit/blob/80e084cd/pkg/parser/ast/ddl.go#L2444) has no case for `TableOptionEngineAttribute` (only `TableOptionSecondaryEngineAttribute` at 2410), so legal MySQL 8.0.21+ output with a table-level `ENGINE_ATTRIBUTE='{}'` returns `restore the parsed statement without its counter: ... invalid TableOption: 31` even though the sliced text was already correct. Via [table_schema.go:127](https://github.com/block/spirit/blob/80e084cd/pkg/table/table_schema.go#L127) one such table makes `LoadSchemaAndExcludedTablesFromDB` return `nil, nil, err` for the entire database, where the deleted function just returned the statement unchanged.

**A comma-separated table option list leaves a stray comma behind, so the re-parse fails where the old implementation succeeded.** MySQL allows `table_option [[,] table_option] ...`, but `trimSpacingBefore` at [auto_increment.go:134](https://github.com/block/spirit/blob/80e084cd/pkg/table/auto_increment.go#L134) walks back only over spaces and tabs, so `ENGINE=InnoDB, AUTO_INCREMENT=5, DEFAULT CHARSET=utf8mb4` becomes `ENGINE=InnoDB, , DEFAULT CHARSET=utf8mb4` and the re-parse at [line 88](https://github.com/block/spirit/blob/80e084cd/pkg/table/auto_increment.go#L88) errors with `near ", DEFAULT CHARSET=utf8mb4"`. Reproduced at head; the deleted implementation returned a correctly stripped statement for the same input, and no test in `auto_increment_test.go` uses comma-separated options.

**A duplicated `AUTO_INCREMENT` option can never verify, and the error blames the wrong thing.** [Line 53](https://github.com/block/spirit/blob/80e084cd/pkg/table/auto_increment.go#L53) strips only the first option (`slices.IndexFunc`) while [line 92](https://github.com/block/spirit/blob/80e084cd/pkg/table/auto_increment.go#L92) builds the expected form with `slices.DeleteFunc`, which drops all of them. MySQL 8.0.33 accepts `... AUTO_INCREMENT=1 ENGINE=InnoDB AUTO_INCREMENT=2` (last value wins) and this PR's own `TestTableOptionOriginTextSpanRepeated` pins spans for exactly that shape, yet at head it returns `removing the counter changed the rest of the statement` — nothing is stripped and the rest of the statement was not changed.

### General suggestions

- All three findings share one root cause: verifying a byte-level edit by re-rendering through `Restore` couples a narrow text slice to the fork's full (incomplete) render coverage. Comparing the stripped text against the original text minus the recorded span — or downgrading a failed verification to "return the original statement" rather than an error — would keep the safety check without turning render gaps into schema-load aborts.

The span offsets recorded by the new grammar action are the riskiest mechanism, since a wrong offset would silently slice real DDL. goyacc's reduce preamble is `parser.yyVAL = &yyS[yyp+1]` ([parser.go:13061](https://github.com/block/spirit/blob/80e084cd/pkg/parser/parser.go#L13061)), so `$$` aliases `$1`'s slot; the code reads `$1`'s offset only when `$1.(bool)` is true (the `FORCE` terminal was shifted) and otherwise anchors on the `AUTO_INCREMENT` keyword at `yyS[yypt-2]`. `literalEndOffset` is sound because the `LengthNum -> NUM -> intLit` chain never rewrites `ident`, and `counterSpan` re-checks `stmt[start:end] != text` before slicing, so a span from a different statement yields `ok=false` instead of a panic.

### Verified correct

- Generated `parser.go` case 2160 matches the `parser.y` action byte for byte; CI's "generated parser matches grammar" job is green.
- `parser.src` is exactly the string handed to `Parse`/`ParseOneStmt`, so recorded offsets index the caller's own text.
- New span fields are inert for existing consumers: only `pkg/lint/lint_unsafe.go` reads `OriginalText()`, and on the statement node, not a `TableOption`.
- No `TableOption` rule other than `AUTO_INCREMENT` calls `setNodeText`, and nothing in `pkg/statement` / `pkg/fmt` reads `Text()` off options.
- Realistic DDL strips cleanly against live MySQL 8.0.33: generated columns, SRID geometry + SPATIAL KEY, FULLTEXT, multi-valued functional index, CHECK, INVISIBLE, expression DEFAULT, escaped quotes, RANGE partitioning, COMPRESSED row format, FOREIGN KEY.
- No production caller passes `WithStrippedAutoIncrement` today, so the new error path and the `string` → `(string, error)` signature change cannot break an existing flow.
- The one updated call site at [table_schema.go:124](https://github.com/block/spirit/blob/80e084cd/pkg/table/table_schema.go#L124) propagates the error without shadowing `err`.
- `verifyOnlyCounterRemoved`'s in-place `original.Options` mutation is safe here, and the early `i < 0` return means a counter-free statement is returned byte-for-byte without needing `Restore`.

Kiran01bm (approved):
🤖 Approved on Kiran's (@kmuddukrishna) behalf by the scheduled review agent — no blocking findings at 80e084cd. See the review comment above; non-blocking findings and suggestions, if any, are not merge gates.

morgo (approved):

PR #1243, opened 2026-09-18
fix(checksum): avoid prepared LIMIT offsets for hot-range splits

### LINKED ISSUE #1242: Hot-range split LIMIT parameter fails through Vitess
🤖 Continuous checksum hot-range splitting fails against Vitess when the split-key query binds the LIMIT offset. The reported tablet SQL contains a bind variable for the offset, and MySQL rejects the resulting query near 'null, 1'.

Spirit currently executes SELECT ... ORDER BY ... LIMIT 1 OFFSET ? with rows/2 as a bound parameter, and uses the same parameterized query with zero for the stale-count fallback. Splitting falls back to bounded retries, leaving repeatedly changing ranges unresolved.

Render the internally computed unsigned offset as a decimal SQL literal for both paths. Add a real-MySQL regression test proving the split lookups do not prepare statements and still select the median/fallback keys. The Vitess failure is reported from a live import; local MySQL alone does not reproduce the proxy's NULL substitution.

morgo:
Fixes #1242.

Hot-range splitting can fail through Vitess because its prepared LIMIT offset reaches MySQL as NULL. Render the internally computed uint64 offset as a decimal literal for both the median lookup and the stale-count fallback. This avoids the prepared-statement path without changing global connection settings or checksum correctness rules.

The regression test runs against real MySQL with parameter interpolation disabled, checks that neither lookup increments the session's prepared-statement counter, and verifies the chosen pivots. Both cases fail against the previous implementation.

Validation: the full checksum suite passed with the race detector; checksum golangci-lint passed.


aparajon:
## 🤖 What the new test actually kills, and the one line that keeps it honest

Reviewed `7aefb95fdfb2b75914ef6fdb5c5d093e440e1853` — both offset sites, the completeness of the change across the repo, and the new test's discriminating power, by mutating the fix and re-running against a real MySQL 8.0.45 (`compose/compose.yml`, `MYSQL_DSN=tsandbox:msandbox@…/test`) rather than by reading the diff.

**0 blocking, 2 non-blocking.**

**The body's central claim holds, and I measured it.** Reverting both sites to the prior `LIMIT 1 OFFSET ?` with bound args fails *both* subtests on the `Com_stmt_prepare` delta — `median` expected 5 got 6, `stale count fallback` expected 6 got 8, one extra prepare per lookup that still binds. Reverting *only* the fallback site (`query+"0"` → `query+"?"` with a bound `0`) fails only `stale count fallback`, so the two sites are independently pinned rather than one assertion covering both. Formatting the median offset from `rows` instead of `rows/2` fails on the pivot assertion (`expected 20, actual 10`), because the wrong offset runs off the end and the `ErrNoRows` fallback silently returns the first key — so the pivot check at `:453` is load-bearing too, not decoration. That third mutation is the interesting one: without the pivot assertion, the fallback would have absorbed the error and the prepare counter would still have been satisfied.

**This converges the last divergent site onto a convention the repo already had.** `pkg/table/chunker_composite.go:93` and `:103`, and `pkg/table/chunker_optimistic.go:111`, already format the offset as a decimal literal (`LIMIT 1 OFFSET %d`), and `chunker_optimistic.go:111` keeps its key bound (`WHERE %s > ?`) in the same statement. So the established rule is *keys bound, offsets literal*, and `continuous_split.go` was the one place that bound an offset. Grepping the whole tree for any other bound `LIMIT`/`OFFSET` returns nothing, so the fix is complete for the class — there is no second site left to follow up on.

The concatenation is injection-safe on its own terms: `rows` is the function's own `uint64` parameter and `strconv.FormatUint` emits digits only, so nothing caller-influenced reaches the statement text. The comment at `:52-54` states exactly that, which is the right thing for it to say.

### Non-blocking

**1. `query` is no longer a statement, and the failure mode of forgetting that is a runtime error rather than a compile error.** After `:47`, `query` ends in a dangling `LIMIT 1 OFFSET ` with a trailing space, and it is only valid once one of the two call sites appends to it. Both current uses do. But the variable is still named `query` and still typed `string`, so a future edit that adds a third lookup and passes `query` straight to `QueryRowContext` compiles fine and fails at the server with a syntax error — and on a hot-range split path, that is a failure discovered in production rather than in review.

The repo's own convention closes this without a new concept: build the two full statements with `%d` in the existing `fmt.Sprintf`, exactly as `chunker_composite.go:93` does, which also drops the `strconv` import. If the single-format version is preferred for the shared prefix, naming it `queryPrefix` gets most of the protection for one word.

<details>
<summary>Test case (run here): the convention-matching form passes the new test unchanged</summary>

Replacing the dangling-prefix construction with two fully-formed statements keeps both subtests green, so this is a readability choice and not a behavior change:

```go
median := fmt.Sprintf("SELECT %s FROM %s WHERE %s ORDER BY %s LIMIT 1 OFFSET %d",
    strings.Join(projections, ","), parent.Table.QuotedTableName, parent.String(), keys, rows/2)
first := fmt.Sprintf("SELECT %s FROM %s WHERE %s ORDER BY %s LIMIT 1 OFFSET 0",
    strings.Join(projections, ","), parent.Table.QuotedTableName, parent.String(), keys)
```

`go test ./pkg/checksum/ -run TestHotSplitDoesNotPrepareOffset` stays green, and `strconv` is no longer imported.
</details>

**2. `cfg.InterpolateParams = false` at `:420` is the single line that makes the counter assertion mean anything, and nothing says so.** I flipped it to `true` while holding the *pre-fix* implementation in place, and the test passes both subtests — the Go MySQL driver interpolates the argument client-side, sends text protocol, and never prepares. So with that one line relaxed, the guard silently stops guarding while still looking like a strict assertion.

It is correct as written, and faithful to the configuration that actually ships: `pkg/dbconn/dbconn.go:88` defaults `InterpolateParams` to `false`, and the only way to change it is the hidden `--interpolate-params` flag at `pkg/migration/migration.go:111`. Which incidentally explains the bug's environment-dependence — an operator running with that flag set would never have reached the Vitess path at all.

Worth one clause on that line saying what it is for (something like "force the prepared path; with interpolation on, the driver sends text protocol and the counter can never move"), because it reads today like defensive DSN normalization rather than the precondition of the assertion four lines of setup later. The two things most likely to retire this test are someone tidying that line away as redundant-looking, and the driver changing its own default.

### Notes

Spirit has no `docs/invariants.md`, so there is no registry ID to cite; the nearest stated rule is the chunker convention above, which this change now makes uniform.

Opening the pool directly with `sql.Open("block-mysql", …)` rather than through `pkg/dbconn` is right here and I am not flagging it: the session-scoped `Com_stmt_prepare` counter is only readable if every query lands on one connection, which is what `SetMaxOpenConns(1)`/`SetMaxIdleConns(1)` at `:424-425` buy. Reading it through a normal pool would be a flaky test, not a stricter one.

What was actually run, against MySQL 8.0.45: the new test at the head (green, both subtests); four mutations (full revert to bound args; fallback site only; `rows` instead of `rows/2`; pre-fix implementation with `InterpolateParams=true`) — the first three died as described, the fourth passed, which is the finding above; the full `pkg/checksum` package (`ok`, 16.9s); `go build ./...`, `go vet ./pkg/checksum/`, and `gofmt -l` on both changed files, all clean. Every line number above was read at this head. CI at this head is 19/19 green.

aparajon (approved):
🤖 Approving `7aefb95fdfb2b75914ef6fdb5c5d093e440e1853` — nothing blocking.

The fix converges the one site that still bound a `LIMIT` offset onto the convention `pkg/table/chunker_composite.go:93` and `pkg/table/chunker_optimistic.go:111` already follow, and a repo-wide grep finds no other bound offset, so the class is closed rather than patched. The new test discriminates: reverting both sites fails both subtests on the `Com_stmt_prepare` delta, reverting only the fallback fails only that subtest, and formatting the offset from `rows` instead of `rows/2` fails on the pivot assertion. Full `pkg/checksum` green against MySQL 8.0.45.

Two non-blocking notes in [the review comment](https://github.com/block/spirit/pull/1243#issuecomment-5724184453): `query` is now a dangling prefix whose misuse would be a runtime rather than a compile error, and `cfg.InterpolateParams = false` is the one line that keeps the counter assertion non-vacuous (with it relaxed, the pre-fix implementation passes) and deserves a clause saying so. Neither changes behavior.

morgo:
🤖 Addressed both suggestions in 81345d0d: renamed the incomplete SQL string to queryPrefix, and documented why InterpolateParams must remain false for the prepared-statement counter assertion to detect regressions. Kept the shared prefix to avoid duplicating the projection/predicate construction. The focused regression test passed with -race, and checksum lint passed.

Issue #1165, opened 2026-08-26
datasync: resume can silently copy with a stale target schema after source DDL


morgo:
🤖 *Filed on behalf of @morgo via Claude Code.*

## Summary

On the resume path, `datasync` skips creating any target table that already
exists without comparing its schema to the source. Combined with the
source/target column *intersection* used to build the copy statement, a source
DDL that lands between attempts is silently ignored rather than failing: new
source columns are excluded from both the copy and the checksum, so the sync
converges "clean" while the target is missing data.

This affects any consumer that retries an errored sync.

## Mechanism

1. `Runner.createTargetTables` does an `information_schema.TABLES` existence
   check per table and, when the table is present, logs `target table already
   exists, skipping creation` and `continue`s. It never compares the existing
   target definition to the source.
2. `startResume` calls `createTargetTables` exactly as `startFresh` does, so a
   resumed sync inherits whatever schema the target already had.
3. `table.ColumnMapping.computeIntersection` builds the copy column list from
   the intersection of `sourceTable.NonGeneratedColumns` and
   `targetTable.NonGeneratedColumns`.

So a source `ADD COLUMN` that is checkpointed past produces a target table
without that column, the column is dropped from the copy, and — because the
checksum compares the same intersection — the comparison still passes. There is
no error and no divergence signal.

The aborted run leaves a resumable checkpoint even when it aborted *because* of
the DDL: `runContinuous`'s drain path writes the (`checkpoint.Persistent`)
checkpoint on a detached `context.WithoutCancel` budget after cancellation.

## What self-heals and what does not

- A **new table** on the source self-heals — it fails the existence check and
  gets created.
- **`ADD INDEX`** (secondary) largely self-heals: `restoreSecondaryIndexes` runs
  after every copy, re-derives missing secondary indexes by comparing source and
  target `SHOW CREATE TABLE`, and is documented as idempotent.
- **`ADD` / `DROP` / `MODIFY COLUMN`** does *not* self-heal, and is silent.

## Reproduction sketch

1. Start a continuous `datasync` and let the initial copy finish so a checkpoint
   is written.
2. Cause the run to error (any path that returns an error from `Run` while a
   checkpoint exists).
3. `ALTER TABLE ... ADD COLUMN` on a source table.
4. Re-run with the same target. `createTargetTables` skips the table, the new
   column is absent from the target, the copy omits it, and the continuous
   checksum reports convergence.

## Suggested directions

Any one of these would close the silent-failure hole:

[...]

PR #1245, opened 2026-09-18
fix(statement): preserve required auto-increment indexes during deferral

### LINKED ISSUE #1244: Deferred indexes must preserve AUTO_INCREMENT support
🤖 RemoveSecondaryIndexes can remove the sole supporting index for an AUTO_INCREMENT column when it is not the leading column of a retained primary or unique key. The resulting CREATE TABLE fails in InnoDB. Preserve one regular index beginning with the auto-increment column when no retained key already supports it. This affects callers enabling deferred secondary indexes.

morgo:
Fixes #1244.

Preserve a regular index required to keep an AUTO_INCREMENT column valid when deferring secondary indexes. A retained primary or unique key with that column first already satisfies the requirement; otherwise preserve the supporting regular index with the fewest key parts and defer the rest.

Datasync's existing-target schema comparison uses a separate helper that ignores all regular indexes, so equivalent support indexes with different names or trailing columns remain accepted. Restore comparisons match unnamed indexes by definition to avoid silently omitting deferred indexes.

Real-MySQL tests cover secondary-only support, composite primary keys, inline declarations, choosing the narrowest supporting index, functional indexes, inserts, and named/unnamed index restoration. Full statement and datasync suites passed with -race; statement, datasync, and move lint passed. Documentation and CLI help describe the AUTO_INCREMENT exception.


aparajon:
## 🤖 What else reads `RemoveSecondaryIndexes`, and which supporting index the deferral keeps

Reviewed `b2df363349` — the retention rule, the new test, and all four production callers of the two helpers it changes. **0 blocking, 5 non-blocking.** spirit carries no `docs/invariants.md`, so there is no registry entry to cite a disposition against.

The fix itself is right, and the shape of it is right: InnoDB needs the AUTO_INCREMENT column to lead *some* index, a retained PRIMARY or UNIQUE whose first key part is that column already satisfies it, and only when neither does must a regular index survive. The new test earns its keep — it executes the stripped DDL against a real server, inserts a row to prove the counter still works, and then round-trips the restore until `GetMissingSecondaryIndexes` comes back empty. Seven of my eight mutations die on it.

Eight mutations of `secondary_indexes.go` against the tests at this head, 7 killed:

| Mutation | Result |
|---|---|
| Strip every regular index (the pre-fix behaviour, `:45-54`) | KILLED — `AutoIncrement/{secondary,composite_primary_suffix,multiple_supporting}` |
| Never clear the need from a retained primary/unique (`:35`) | KILLED — `AutoIncrement/{primary,unique,inline_unique}` |
| Retain *every* supporting index, not one (drop `delete`, `:53`) | KILLED — `AutoIncrement/multiple_supporting` |
| Match on the index's **last** key part instead of its first (`:49`) | KILLED — `AutoIncrement/composite_primary_suffix` |
| Only a primary key clears the need, not a unique (`:35`) | KILLED — `AutoIncrement/{unique,inline_unique}` |
| Clear the need from a primary/unique's **last** column (`:36`) | KILLED — `AutoIncrement/composite_primary_suffix` |
| Treat every column as AUTO_INCREMENT (`:30`) | KILLED — `TestRemoveSecondaryIndexes/{Table_with_all_index_types,…}` |
| Drop the nil-column guard on a functional index (`:46-48`) | **SURVIVED** — finding 4 |

---

### Non-blocking

#### 1. This function has two callers wanting two different things, and the comparison caller now gets the wrong one

`pkg/datasync` calls `RemoveSecondaryIndexes` on **both** sides to decide whether an existing target may be copied into (`runner.go:1092-1096`, inside `verifyExistingTargetTable`). Its doc comment states the premise the comparison rests on (`runner.go:1085-1090`):

> Regular secondary indexes are excluded from the comparison on both sides. With DeferSecondaryIndexes the target is deliberately created without them, and an attempt that died mid-copy leaves it that way

After this change that is no longer true: one regular index survives on each side, and if the two sides disagree about *which*, the copy is refused. Measured by running the function's own two calls plus `DiffCreateTables` over four pairs:

| source / target | diff before | diff now |
|---|---|---|
| `KEY ai(id)` / `KEY target_ai(id)` | *(empty)* | ``DROP INDEX `target_ai`, ADD INDEX `ai` (`id`)`` |
| `KEY ai(id)` / `KEY ai(id,p)` | *(empty)* | ``DROP INDEX `ai`, ADD INDEX `ai` (`id`)`` |
| `KEY extra(p), KEY ai(id)` / `KEY ai(id), KEY extra(p)` | *(empty)* | *(empty)* |
| same pair as row 1 but **no** AUTO_INCREMENT column | *(empty)* | *(empty)* |

The last two rows are the controls: declaration order among non-supporting indexes still does not matter, and dropping the AUTO_INCREMENT column makes the divergence disappear — so the retention is precisely what introduces it. The first two are the reachable ones, because `verifyExistingTargetTable` exists for a target this run did not create (the `#1165` case in its own comment): an operator-provisioned target, or one built by another tool, can carry an equivalent supporting key under a different name or with an extra trailing column. That target is now rejected with "its schema has diverged from the source; copying into it is unsafe", and the remedy the message offers is a `DROP INDEX` / `ADD INDEX` pair that changes nothing an operator would recognise as wrong.

It fails closed and the message is actionable, so this is not blocking. But the two call sites now want opposite things from one function: the CREATE path needs the support kept, and the comparison path needs the blanket exclusion its comment promises. Splitting them — a second entry point, or an option on this one — keeps both true, and is the version where the comment above needs no edit. (Copilot's review names this call site too; the measurements above are what I would want before choosing between splitting it and narrowing the comparison.)

#### 2. Which supporting index survives is decided by declaration order, and the new test pins only how many

The keep branch takes the first regular index whose first key part is the AUTO_INCREMENT column and deletes the need, so every later candidate is dropped. With two candidates, source order alone decides which one the copy pays for:

```
source: …PRIMARY KEY(p), KEY narrow(id), KEY wide(id,p,q,r)
kept:   …PRIMARY KEY(`p`),INDEX `narrow`(`id`)

source: …PRIMARY KEY(p), KEY wide(id,p,q,r), KEY narrow(id)
kept:   …PRIMARY KEY(`p`),INDEX `wide`(`id`, `p`, `q`, `r`)
```

The flag's own rationale is what this trades against — "Bulk-loading an index-free table is faster and lighter on temporary space" (`pkg/datasync/sync.go:67-68`, repeated at `docs/sync.md:162`). A four-part index maintained across the whole copy, chosen over a one-part index that would have satisfied InnoDB identically, is the cost the option exists to avoid, and nothing about the source's declaration order predicts which one an operator gets. Preferring the candidate with the fewest key parts costs one comparison and makes the choice a property worth asserting.

`TestRemoveSecondaryIndexesAutoIncrement/multiple_supporting` has exactly this shape (`KEY ai(id), KEY redundant(id,p)`) but asserts only `regular == 1`, so it passes whichever is kept — which is why none of my mutations needed to defend the choice.

<details>
<summary>Test case — pin the retained index by name, not by count</summary>

```go
{"prefers the narrowest support", "id INT AUTO_INCREMENT, p INT PRIMARY KEY, KEY wide(id,p), KEY narrow(id)", 1, "narrow"},
```

…with the assertion reading the name rather than counting:

```go
var kept []string
for _, idx := range ct.GetIndexes() {
    if idx.Type == "INDEX" {
        kept = append(kept, idx.Name)
    }
}
require.Equal(t, tc.regular, len(kept))
if tc.support != "" {
    require.Equal(t, []string{tc.support}, kept)
}
```

Fails at this head (keeps `wide`, declared first); passes once the branch prefers the fewest key parts. The existing six cases all have a single candidate, so they are unaffected by the new column.

</details>

#### 3. Five places still promise that deferral removes every regular index

The PR updates the function's own doc comment (`secondary_indexes.go:16-18`) and stops there. Still stating the old contract, none of them in the diff:

- `pkg/statement/README.md:382` — "Removes regular secondary indexes … while preserving PRIMARY KEY, UNIQUE, and FULLTEXT indexes", and the explicit two-column list at `:398-404`, whose **What's Removed** entry is "Regular INDEX (non-unique secondary indexes)".
- `docs/sync.md:159-163` — "the target tables are created **without their regular secondary indexes**" and "only regular secondary indexes are deferred".
- `docs/move.md:69` — "target tables are created without secondary indexes".
- The two flag help strings an operator reads from `--help`: "Create target tables without secondary indexes, then add them after the initial copy." (`pkg/datasync/sync.go:73`) and "Create target tables without secondary indexes, add them before cutover" (`pkg/move/move.go:32`).
- `pkg/datasync/runner.go:1085-1090`, quoted in finding 1 — the one that is load-bearing rather than descriptive, since it is the stated reason the comparison is safe.

The README pair is the one most likely to mislead a caller, because it reads as an exhaustive partition of index types and the new behaviour belongs to neither column.

#### 4. The nil-column guard is the only thing between a functional index and a panic, and nothing exercises it

`if len(constraint.Keys) == 0 || constraint.Keys[0].Column == nil` (`:46-48`) is correct and necessary: a functional index part carries `Expr` and leaves `Column` nil, and the line below it dereferences `Column.Name.O` for *every* regular index, not only candidates. Removing the guard is the one mutation that survives the suite — measured, on a table that is otherwise entirely ordinary:

```
CREATE TABLE t (id INT AUTO_INCREMENT, p INT, b VARCHAR(32),
                PRIMARY KEY(p), KEY ai(id), KEY fn((lower(b)))) ENGINE=InnoDB
```

At this head that strips to ``PRIMARY KEY(`p`),INDEX `ai`(`id`)`` and restores as ``ALTER TABLE `t` ADD INDEX `fn` ((LOWER(`b`)))`` — correct on both halves, since `GetMissingSecondaryIndexes` already handles the expression case (`:166-175`). With the guard removed the same input panics with a nil dereference. A guard whose absence is a panic rather than a wrong answer is worth one case.

<details>
<summary>Test case — a functional index through deferral</summary>

```go
func TestRemoveSecondaryIndexesFunctionalIndex(t *testing.T) {
	src := "CREATE TABLE t (id INT AUTO_INCREMENT, p INT, b VARCHAR(32), " +
		"PRIMARY KEY(p), KEY ai(id), KEY fn((lower(b)))) ENGINE=InnoDB"
	stripped, err := RemoveSecondaryIndexes(src)
	require.NoError(t, err)
	require.NotContains(t, stripped, "fn")
	require.Contains(t, stripped, "`ai`")
	restore, err := GetMissingSecondaryIndexes(src, stripped, "t")
	require.NoError(t, err)
	require.Contains(t, restore, "ADD INDEX `fn`")
}
```

Passes at this head; panics with the nil-column guard removed. No server needed.

</details>

#### 5. A retained *unnamed* index collides with every other unnamed index in the restore comparison

`GetMissingSecondaryIndexes` keys the target's index set on `constraint.Name` (`:97`), so an index with the empty name claims the empty-name slot for all of them. Until this change a deferred target had no regular indexes at all, and the retained one is the first that can arrive unnamed:

```
source:   CREATE TABLE t (id INT AUTO_INCREMENT, p INT, PRIMARY KEY(p), KEY (id), KEY (p))
stripped: CREATE TABLE `t` (…,PRIMARY KEY(`p`),INDEX(`id`))
restore:  ""        ← KEY (p) is never added back
```

Measured. `KEY (p)` is silently dropped rather than restored, which is the worst failure mode this pair has.

It is **not reachable through any current caller**: all four pass `SHOW CREATE TABLE` output on both sides (`datasync/runner.go:999`, `:1012`, `:1132`, `:1137`; `move/runner.go:393`, `:1720`, `:1735`), and MySQL always names indexes there, so `Name` is never empty in production. So this is a boundary defect in an exported pair rather than a live bug — but the pair is exported from a public package, `pkg/move` and `pkg/datasync` are not its only possible callers, and the guard is one line: skip the empty name when building `targetIndexes`, or compare unnamed indexes by column set the way `diffIndexes` already does for inline-derived uniques.

---

<details>
<summary>Smaller notes</summary>

- **The inline forms work, but not for the reason the comment gives.** "GetIndexes also includes inline PRIMARY KEY and UNIQUE clauses" (`:27`) is true, and the `inline_unique` and `primary` cases pass — but `GetIndexes`' own column-level synthesis (`create_table.go:273-292`) sets only the deprecated `Columns` field and leaves `ColumnList` empty, so those entries are silently skipped by the `len(index.ColumnList) > 0` guard. What actually clears the need is that `ParseCreateTable` has already normalised both inline forms into `ct.Indexes` with a populated `ColumnList` — I confirmed it on the raw DDL, where the parse reports `name="PRIMARY" type="PRIMARY KEY" columnList=1` and `name="id" type="UNIQUE" columnList=1`. The guard is right to be defensive; it is the comment that credits the wrong mechanism, and the cost of the confusion is a silently retained redundant index rather than an error.
- `Index.Type`'s own field comment lists the values as "PRIMARY, UNIQUE, INDEX, FULLTEXT, SPATIAL" (`create_table.go:73`) while the parser writes `"PRIMARY KEY"` (`:600`). The new code matches the parser, so it is correct; the stale comment is pre-existing and is the kind that makes a string comparison look wrong on review.
- The test's decision to run each case twice — once on the authored DDL and once on `SHOW CREATE TABLE` — is the right call and is what makes the inline cases meaningful. Worth keeping if the cases are extended.

</details>

<details>
<summary>What was run</summary>

`go build ./...` clean at `b2df363349`. Baseline `ok github.com/block/spirit/pkg/statement 13.165s` against a local MySQL. The eight mutations above applied one at a time against a restored source, each run against the whole package. Three probes: the retained-index order measurement in finding 2, the functional-index shape in finding 4 (run at head and with the guard removed), and the target-comparison table in finding 1 (`RemoveSecondaryIndexes` on both sides plus `DiffCreateTables`, the two controls included). No probe files left in the tree.

</details>

aparajon (approved):
🤖 Approving — **0 blocking, 5 non-blocking.**

The retention rule is right and the new test is unusually good for a parser change: it executes the stripped DDL against a real server, inserts a row to prove the AUTO_INCREMENT counter still works, and round-trips the restore until nothing is missing. Seven of my eight mutations die on it, including the pre-fix behaviour and every off-by-one on which key part or which column is consulted.

The non-blocking findings are in the comment above. The one worth a second look before this lands is that `pkg/datasync` uses `RemoveSecondaryIndexes` as a *comparison* primitive on both sides of an existing target (`runner.go:1092-1096`), so retaining an index narrows what that comparison forgives — measured, with two controls. It fails closed with an actionable message, so it is not a blocker, but the two callers now want opposite things from one function.

morgo:
🤖 Addressed all five findings and both inline comments in 496dca93:

- Added a comparison-only helper that strips every regular index, and switched datasync's existing-target check to it. Regression tests accept differently named and differently sized AUTO_INCREMENT support indexes.
- Prefer the supporting regular index with the fewest key parts; test the retained name with both declaration orders.
- Updated statement documentation, sync/move documentation, flag help, and the inline-normalization/type comments.
- Added a functional-index case through the real-MySQL create/insert/restore test.
- Match unnamed indexes by their definition rather than the shared empty name. A real-MySQL regression verifies the omitted index is restored without duplicating the retained index, including comparison against SHOW CREATE TABLE.

Validation: full statement and datasync suites passed with -race; lint passed for statement, datasync, and move. Enabling auto-merge pending CI.

PR #1247, opened 2026-09-18
feat(datasync): distinguish scan progress from remaining verification

### LINKED ISSUE #1246: Separate continuous verification scan progress from remaining ranges
🤖 Continuous verification's passed/emitted counters resemble a completion fraction, but hot splitting retires parents and emits children, making that interpretation misleading.

Separate estimated scan progress (or confirmed scan completion) from remaining retrying, in-flight, and deferred ranges. Show splits, mismatch observations, and recopies as per-pass activity. Keep raw counters available in the stats API and detailed pass logs. Scan completion must not imply verification, and repaired ranges still need verification in the next pass.

morgo:
Fixes #1246.

Separate continuous verification's scan progress from unresolved ranges. Hot splitting retires parents and emits children, so passed/emitted is not a completion fraction. The datasync status block now renders:

```text
verify  pass=1  scan complete
        remaining: 1 retrying (0 hot), 0 in flight, 0 deferred
        pass activity: 63 chunks mismatched: 49 split, 0 recopied
        permanent failures: 0  walker stalls: 0
```

Before successful walker exhaustion, show scan≈N%. Scan completion does not imply verification. Between passes, show the completed pass and next scheduled start; show the historical first clean-pass timestamp once available. Repaired ranges explicitly require verification in the next pass. Raw counters remain available in the stats API, with outcome breakdowns in the pass-complete logs.

Validation: full checksum and datasync suites passed with -race; checksum and datasync lint passed. Regression coverage includes hot splits, 100% estimates before completion, outstanding retries, repairs/deferrals, walker failures, inter-pass scheduling and cancellation, and first clean-pass evidence.


aparajon:
## 🤖 What the verify block asserts between passes, and how its three activity counts relate

Reviewed `b422dc48f4` — `appendVerificationStatus`, the `ScanComplete` plumbing, both new tests, and the stats fields the row now reads. **0 blocking, 6 non-blocking.** spirit carries no `docs/invariants.md`, so there is no registry entry to cite a disposition against.

The premise holds up: `passed/emitted` really is not a completion ratio, and the code says so at the source. `ChunksThisPass` counts split parents *and* their children (`continuous.go:239-241`); `handleResult` records a split parent as split and never as passed (`:1116-1120`); `ProgressBasisPoints` exists precisely because "ChunksThisPass ... can never be an honest denominator" (`:247-252`). Splitting the row into traversal (`scan≈`/`scan complete`) and unresolved work (`remaining:`) is the right cut, and `ScanComplete` is a real fact the old row could not express — 100% progress with an outstanding retry is exactly the state `TestScanCompleteWithOutstandingRetry` pins.

Two things I went looking for and did not find, so they are not findings: `remaining: %d retrying (%d hot)` is **correct** — `HotChunkCount` is defined over entries *currently in the retry queue* with `consecutiveSrcChanged >= 2` (`:298-301`), so hot is a subset of retrying, not a separate population. And there is no second renderer left holding the old format: `appendVerificationStatus` is called once (`runner.go:1724`), and the only other exit for these counters is `ChecksumStats()` (`runner.go:648-654`).

Eight mutations of the changed code against the tests at this head, 6 killed:

| Mutation | Result |
|---|---|
| Invert the `ScanComplete` branch (`runner.go:1781`) | KILLED — `TestSyncProgressAndLogFormat`, `TestVerificationStatusAfterHotSplits` |
| Mis-scale the scan estimate (`/10` instead of `/100`) | KILLED — `TestVerificationStatusAfterHotSplits` |
| Swap `retrying` and `hot` (`runner.go:1785`) | KILLED — `TestVerificationStatusAfterHotSplits` |
| Swap `splits` and `mismatch observations` (`runner.go:1787`) | KILLED — `TestVerificationStatusAfterHotSplits` |
| Drop `scanComplete.Store(false)` from the per-pass reset (`continuous.go:618`) | KILLED — `TestScanCompleteResetsAndExcludesWalkerFailure` |
| Store scan-complete *before* the walker-error drain (`continuous.go:859`) | KILLED — `TestScanCompleteResetsAndExcludesWalkerFailure` |
| Always emit the repair advisory, not just when `RecopiesThisPass > 0` (`runner.go:1789`) | **SURVIVED** — finding 5 |
| Store scan-complete even when the pass was cancelled (`continuous.go:859`) | **SURVIVED** — unobservable, see the smaller notes |

---

### Non-blocking

#### 1. The block freezes on the finished pass for up to an hour, and the advisory says "in the next pass" without saying when

`datasync` passes `ContinuousMinPassInterval` = **1h** (`runner.go:562`, `continuous.go:151`), and the per-pass reset runs at the *start* of the next pass (`continuous.go:618-631`), **after** the wait (`:603-612`). So for the whole gap between passes — 59 minutes for a small table whose pass finishes in seconds, which is the case the interval exists for — the snapshot a status tick reads is the **ended** pass's counters:

```
verify  pass=1  scan complete
        remaining: 0 retrying (0 hot), 0 in flight, 0 deferred
        activity this pass: 0 splits, 0 mismatch observations, 0 recopies
```

Measured, sampling `Stats()` through the wait after a 4-chunk clean pass: `pass=1 passes_completed=1 scan_complete=true chunks_this_pass=4`, unchanged across every sample. Two of the new row's words become wrong there. `pass=1` names the pass that ended, not one in flight — which `CurrentPass`'s own doc is careful about ("the 1-indexed pass number **in flight**", `:235-237`). And "activity **this pass**" labels a pass that is over.

The sharper case is the advisory. A pass with `RecopiesThisPass > 0` leaves

```
        remaining: 0 retrying (0 hot), 0 in flight, 0 deferred
        activity this pass: 0 splits, 12 mismatch observations, 12 recopies
        repaired ranges need verification in the next pass
```

on screen for up to an hour with nothing in flight and nothing scheduled that the operator can see. The statement is true — the next pass does re-read them — but "next" reads as *imminent* when the row above it says nothing is left to do, and the decision it feeds (is this data safe to cut over onto?) turns on *when*. The checker already logs the wait, at Debug with the remaining duration (`:604-605`); the block is where the operator is looking. Rendering the wait, or naming the pass boundary (`pass=1 complete; next pass in 52m`), is what makes the line checkable. The general shape is worth naming because the row is new: a status line's truth conditions include how long the phase underneath it lasts, so a line that reads correctly while a phase takes milliseconds can be false once the same phase takes an hour.

#### 2. The three activity counts are not three independent counts — two of them are inside the third

```
activity this pass: 49 splits, 63 mismatch observations, 0 recopies
```

Those are the numbers in the PR's own fixture (`TestVerificationStatusAfterHotSplits`), and the row invites reading them as 49 + 63 + 0 = 112 events. `MismatchesThisPass`'s doc states the partition (`continuous.go:259-266`):

> At the end of a completed pass this equals PassedSecondAttemptThisPass + PassedUnder5AttemptsThisPass + PassedUnder10AttemptsThisPass + RecopiesThisPass + HotChunksDeferredThisPass + HotChunksSplitThisPass.

So splits and recopies are both **terminal dispositions of** the mismatch count, not siblings of it: the honest reading of the fixture is "63 chunks mismatched, 49 of them were split". Rendering the total first, or as a denominator (`63 mismatched: 49 split, 0 recopied`), makes the containment visible at no extra width. The `deferred` count under `remaining:` is in that same partition, which is worth knowing when reading the two rows together.

"Observations" is the other half of this. `MismatchesThisPass` counts a chunk **once**, on its initial fresh-walk read, and its lifetime twin spells out the exclusion (`:317-318`): "does not include re-failures within a single retry sequence". A chunk seen changing across five retry windows contributes 1, not 5 — so a plural noun of observation is the one reading the counter cannot support. "63 chunks mismatched" is both shorter and what the field holds.

#### 3. The block still has no rendering of the verdict the checker actually computes

The suppression of "verified" is right, and the test asserting its absence is the right instinct: nothing here is verified while a scan estimate is all you have. But the checker *does* compute the verdict — `recopies == 0 && deferredHot == 0` at the end of each pass, firing `FirstCleanPass` (`continuous.go:656-659`) — and the accessor's doc points programmatic callers straight at it (`runner.go:627-629`):

> Programmatic callers that gate on "data is known consistent" (e.g. the import feature) should block on this channel.

A human reading the status block is gating on the same question with none of that: `PassesCompleted` is rendered nowhere in `datasync`, and `pass=N  scan complete  remaining: 0` is *not* the verdict — a pass with deferrals or recopies produces exactly that line and is ineligible. So the operator has to reconstruct from three rows what a caller gets as one signal, and the reconstruction is the part this PR is otherwise removing the need for. `passes=%d` on the verify row, or a `clean` token once the signal has fired, is a fact rather than an estimate, so it does not reopen what the PR closed.

#### 4. `docs/sync.md:241-243` now documents a status surface that no longer exists

```
Split children do not feed the main chunker's walk-progress estimate. Status logs show `hot-split`;
emitted counts include both split parents and their children, while passed counts exclude split parents.
```

That is the doc section for hot-range splitting — the feature this row reworks — and after this PR neither `passed` nor `emitted` is rendered anywhere an operator sees. The sentence is also the clearest statement in the repo of *why* the ratio was misleading, so it is worth moving rather than deleting: it now describes the stats fields (`ChunksThisPass` / `ChunksPassedThisPass`) rather than the status output, and the block's new shape is what the paragraph should show.

Its first clause was already stale before this PR: the only `hot-split` in the tree is that doc line. The code logs `"continuous checksum: splitting hot range"` per split (`continuous.go:1120`) and `hot_chunks_split` in the pass summary (`:682`).

#### 5. The advisory line is pinned in one direction only, and the mutation that removes its condition survives

`TestVerificationStatusAfterHotSplits` asserts the advisory appears when `RecopiesThisPass = 1`, and the no-recopy leg asserts `require.NotContains(t, text, "verified")` — which does not catch `"repaired ranges need verification in the next pass"`, since that string contains "verification", not "verified". So making the line unconditional passes the whole suite, and a block that always warns about repairs is a block an operator learns to ignore.

<details>
<summary>Test case — one assertion closes it</summary>

In the first leg of `TestVerificationStatusAfterHotSplits`, where the fixture has `RecopiesThisPass: 0`:

```go
require.NotContains(t, text, "passed=")
require.NotContains(t, text, "emitted=")
require.NotContains(t, text, "verified")
require.NotContains(t, text, "repaired ranges") // a clean pass must not advise re-verification
```

Fails with the advisory's condition removed; passes at this head. The existing third assertion is worth keeping — it pins a different claim (no verification verdict from a scan estimate) — but it is not load-bearing for this one.

</details>

#### 6. The counters that invite the misreading are still exported without the caveat, where a consumer reads them

This PR adds exactly the right warning twice, on the fields it touches: `ScanComplete` says "an estimate of 100% does not imply ScanComplete" (`:254-257`), and `ProgressBasisPoints` says `ChunksThisPass` "can never be an honest denominator" (`:247-252`). But the pair that produced the row being deleted did not get the same treatment. `ChunksThisPass` documents that it includes split parents and children (`:239-241`); `ChunksPassedThisPass` is just "how many chunks have gone clean in the current pass (either initially or via retry)" (`:243-245`) and never says parents are excluded from it. The asymmetry that makes the ratio dishonest is therefore documented on one side only, in the two fields a downstream renderer reaching `ChecksumStats()` will reach for first. One sentence on `ChunksPassedThisPass` — split parents are never counted here, so this is not a numerator over `ChunksThisPass` — puts the fix at the source rather than in one renderer.

---

<details>
<summary>Smaller notes</summary>

- **The surviving cancellation mutation is unobservable, not untested.** Removing `if ctx.Err() == nil` around `scanComplete.Store(true)` (`continuous.go:859-861`) survives, and it should be reported as defense in depth rather than a test gap: I wrote a 30-iteration probe with a chunker that parks in `Next()` until released, cancels the context, then returns `table.ErrTableIsRead` so the walker exits cleanly and closes `walkCh`. It passes identically at head (0.777s) and mutated (0.789s) — the dispatcher takes its `<-ctx.Done()` arm before the walker's close becomes visible, so no probe through `Run` can distinguish them. The guard is still worth keeping (it is what protects a future dispatcher that drains `walkRecv` before checking cancellation); nothing to do.
- The body's claim that the raw counters "remain in the stats API and detailed pass logs" holds, with one caveat worth knowing: in the pass-complete log, emitted is `total_chunks` and passed is the *sum of five fields* (`first_attempt` + `second_attempt` + `under_5_attempts` + `under_10_attempts` + `recopies` — `bucketPassed` routes every pass into exactly one of them, `:1223-1245`). There is no single `passed` key, and the line only lands at pass end, so mid-pass the numbers are reachable only through the stats API.
- `b.Row("", ...)` for the continuation rows relies on the block renderer padding an empty label; it lines up in the test output, and the four-row shape is a real readability gain over the 13-field single line.
- `ScanComplete` is stored in the walker-closed branch after the error drain, which is the correct order — a walker that failed populates `walkErrCh` before closing `walkCh` (`runWalker`'s own contract, quoted in the comment), and mutating the order is killed.

</details>

<details>
<summary>What was run</summary>

`go build ./...` clean at `b422dc48f4`. Baselines with a local MySQL: `ok github.com/block/spirit/pkg/checksum 36.683s`, `ok github.com/block/spirit/pkg/datasync 6.344s`. The eight mutations above applied one at a time against a restored source, each run against `./pkg/checksum/` and `./pkg/datasync/`. Two probes: the inter-pass freeze measurement in finding 1 (`MinPassInterval = 1h`, sampling `Stats()` through the wait), and the cancellation probe in the smaller notes, run at head and mutated. No probe files left in the tree.

</details>

aparajon (approved):
🤖 Approving — nothing blocking at `b422dc48f4`. The premise checks out in the code rather than only in the summary: `ChunksThisPass` includes split parents and their children while a split parent is never counted as passed, so the ratio the old row rendered really was not a completion fraction, and `ScanComplete` expresses something the old row could not — 100% walked with a range still unresolved. 6 of 8 mutations of the changed code are killed, and both survivors are explained in the comment (one is a real one-assertion test gap, the other is unobservable through `Run` and worth keeping as defense in depth).

Six non-blocking findings are in the review comment. The two worth a look before merge are both about what the new rows assert when nothing is happening: the per-pass counters reset at the *start* of the next pass, so between passes — up to an hour with `ContinuousMinPassInterval` — the block sits frozen on the pass that ended, with `pass=N`, `activity this pass`, and `repaired ranges need verification in the next pass` all describing work that is over and a pass that may be 59 minutes away. And `activity this pass: 49 splits, 63 mismatch observations, 0 recopies` reads as three independent counts when the field's own doc makes splits and recopies dispositions *inside* the mismatch total, so 49 of those 63 are the same chunks counted twice.

morgo:
🤖 Addressed all six review points in 67894110:

- Show the completed pass and next scheduled start during the inter-pass wait; clarify CurrentPass semantics.
- Render mismatch totals with split/recopy outcomes contained within them.
- Show FirstCleanPassAt as historical evidence, without deriving verification from scan progress.
- Update the sync documentation and raw-counter caveat.
- Assert the repair advisory is absent when there were no recopies.
- Document that split parents are excluded from ChunksPassedThisPass.

Full checksum and datasync race suites passed; package lint is clean. Added scheduling/cancellation and status-rendering regressions. Enabling auto-merge, pending the new CI run.

PR #1250, opened 2026-09-18
feat(migration): add experimental lockless checksum

### LINKED ISSUE #1233: Add experimental support for lockless checksum
The checksum algorithm currently requires a ftwrl so that it can create several threads in repeatable-read, and then consistently compare data.

For the datasync package, we've added a continuous checksum which is optimistic in comparing data with retries. As long as there are not hot spots, it should in theory work fine.

It would be nice to offer this as an option for regular schema changes. The advantage besides the lock is no long running transactions, so no HLL growth.

morgo:
Fixes #1233.

Add `migrate --enable-experimental-lockless-checksum` (default off). It runs the lockless checksum as a finite gate using optimistic reads, retries, wide hot-range splitting, and finite per-row snapshot draining instead of the main checksum's setup lock and long-lived repeatable-read snapshots. Binlog application stays active; the final drain and cutover synchronization are unchanged.

For small unresolved hot ranges, take a bounded target-key census first, then freeze the source PK/CRC32 images once. Retry only the frozen obligations: each source image must actually match a target read, and each observed target-only key must become absent. Later inserts do not extend the work set, allowing append-heavy tails to converge. There are no stream-backed or soft passes. Continuously modified/deleted rows whose frozen image never matches remain unresolved and defer without authorizing cutover.

Snapshots are bounded to 128 rows per side and 64 KiB of key data, with 30-second query timeouts. Oversized captures fall back to ordinary retries. Snapshots travel through the existing retry queue, releasing workers between attempts and preserving load throttling/autoscaling. SQL collation matches with an unexpected key representation remain unresolved rather than being mistaken for absent orphans. The shared checker exposes SnapshotHotChunks; this PR enables it only for the experimental migration path for burn-in.

A complete clean pass is required before proceeding. Stable divergence fails closed without repair. Copy checkpoints remain available, but experimental checksum watermarks are neither saved nor resumed, including when switching from a checkpoint written by the default checker. Progress distinguishes traversal from verification.

The checker API, implementation files, logs, tests, and documentation use `LocklessChecker`, `LocklessCheckerConfig`, and `LocklessCheckerStats`. Downstream consumers must update the renamed exported symbols when bumping Spirit.

Validation:
- Full checksum suite with `-race`.
- Focused migration/checkpoint regressions with `-race`, including column rename and detection of missing rows before an old checksum watermark.
- Real-MySQL snapshot tests for growing tails, lag, orphans, deletes, composite/binary/temporal/unsigned keys, collation changes, renamed columns, NULL values, row/byte limits, cancellation, and clean-pass gating.
- Build and checksum/migration lint passed; CLI flag verified.

This is an opt-in experimental algorithm, not a common-snapshot comparison. Stream-backed reconciliation remains a follow-up. Auto-merge is disabled.


aparajon:
Reviewed `6c5d99eef740c311b4c7bffd8a09355e24e68b65`: the new per-row snapshot drain (`pkg/checksum/lockless_snapshot.go`), its gating inside the lockless checker (`tryHotSnapshot` / `checkHotSnapshot` / `RunUntilClean`), the migration wiring behind `--enable-experimental-lockless-checksum`, the checkpoint/resume changes, and the `Continuous*` → `Lockless*` rename. Built, ran `./pkg/checksum/` with and without `-race` and the migration lockless/checkpoint tests against a local MySQL, and ran 25 mutations of the changed decision points.

**0 blocking, 6 non-blocking.**

The core design holds up under mutation. I tried eleven ways to make the snapshot declare a range verified when it should not — dropping the target census, seeding census rows as source obligations, letting the census overwrite the source image, dropping the CRC comparison, accepting absence for a source obligation, dropping the collation guard, making the point predicate exclusive or leading-column-only, removing the `+1` overflow probe, removing the identity length prefix, and letting `RunUntilClean` ignore deferrals or recopies. All eleven die. The two guards I could not kill are both real guards with no test behind them, which is what the first two findings are about.

### Non-blocking

**1. The target census's own row-budget guard is untested, and dropping it silently discards every orphan.**

`captureHotSnapshot` bails to `nil` when either side overflows 128 rows (`lockless_snapshot.go:46-53`). `TestHotSnapshotBoundsAndBudget:123-125` only overflows the **source** side (it inserts ids 4..132 into `src`; `dst` has 3 rows). Mutating just the target read to ignore `oversized` leaves the whole suite green — `readHotSnapshotRows` returns a `nil` map on overflow, so the mutant proceeds with zero target obligations while the source obligations still resolve normally. That is precisely the "snapshot passes with orphans on the target" outcome the census exists to prevent, and nothing currently fails when the guard goes away.

<details>
<summary>Test that fails on the mutant and passes on 6c5d99ee</summary>

```go
func TestHotSnapshotTargetCensusOverflow(t *testing.T) {
	db, chunk := snapshotTestTables(t, "id INT PRIMARY KEY, value INT", []string{"id"})
	snapshotExec(t, db, "INSERT INTO src VALUES (1,10),(2,20)")
	var values []string
	for i := 1; i <= 200; i++ {
		values = append(values, fmt.Sprintf("(%d,%d)", i, i*10))
	}
	snapshotExec(t, db, "INSERT INTO dst VALUES "+strings.Join(values, ","))
	snapshot, err := captureHotSnapshot(t.Context(), db, db, chunk)
	require.NoError(t, err)
	require.Nil(t, snapshot, "a target census over the row budget cannot truncate orphan evidence")
}
```

Passes at `6c5d99ee`. With the target read's `oversized` check removed it fails, and the failure message is the point — the returned snapshot carries `pending` of length 2 (both source rows, `present:true`) and not one of the 198 target-only keys:

```
--- FAIL: TestHotSnapshotTargetCensusOverflow (0.11s)
    Error: Expected nil, but got: &checksum.hotSnapshot{pending:map[string]checksum.hotSnapshotRow{...:{crc:0xa487c06b, present:true}, ...:{crc:0x38bd6866, present:true}}, ...}
```
</details>

**2. The temporal `CAST(... AS CHAR)` is unexercised: the snapshot tests never open a `parseTime=true` connection, and removing the cast keeps the suite green.**

`lockless_snapshot.go:122-126` casts DATE/DATETIME/TIMESTAMP/TIME key columns to `CHAR`, and the comment on `:106-108` says why: "to preserve fractional seconds/zero dates with parseTime=true". The sibling mechanism in `lockless_split.go:82-95` is identical and *is* pinned — `TestHotSplitTemporalKeys` (`lockless_split_test.go:333-340`) parameterises over `parseTime ∈ {false, true}` and includes a zero date. The snapshot side has only `TestHotSnapshotKeyIdentity/temporal`, which runs on the default `testutils` connection; nothing in the repo sets `parseTime`, so that case passes with the cast deleted. The assertion is real; the configuration it exists to discriminate is never instantiated.

<details>
<summary>Test that fails on the mutant (parseTime=true only) and passes on 6c5d99ee</summary>

```go
func TestHotSnapshotTemporalParseTime(t *testing.T) {
	for _, parseTime := range []bool{false, true} {
		t.Run(fmt.Sprint(parseTime), func(t *testing.T) {
			schema, _ := testutils.CreateUniqueTestDatabase(t)
			cfg, err := mysql.ParseDSN(testutils.DSNForDatabase(schema))
			require.NoError(t, err)
			cfg.ParseTime = parseTime
			if cfg.Params == nil {
				cfg.Params = make(map[string]string)
			}
			cfg.Params["sql_mode"] = "'NO_ENGINE_SUBSTITUTION'"
			cfg.Params["time_zone"] = "'+00:00'"
			db, err := sql.Open("block-mysql", cfg.FormatDSN())
			require.NoError(t, err)
			defer func() { _ = db.Close() }()
			for _, name := range []string{"src", "dst"} {
				_, err = db.ExecContext(t.Context(), "CREATE TABLE "+name+" (id DATETIME(6) PRIMARY KEY, value INT)")
				require.NoError(t, err)
			}
			_, err = db.ExecContext(t.Context(),
				"INSERT INTO src VALUES ('0000-00-00 00:00:00',0),('2026-09-18 01:02:03.123456',10)")
			require.NoError(t, err)
			source, target := table.NewTableInfo(db, schema, "src"), table.NewTableInfo(db, schema, "dst")
			require.NoError(t, source.SetInfo(t.Context()))
			require.NoError(t, target.SetInfo(t.Context()))
			chunk := &table.Chunk{Key: []string{"id"}, Table: source, NewTable: target,
				ColumnMapping: table.NewColumnMapping(source, target, nil)}
			snapshot, err := captureHotSnapshot(t.Context(), db, db, chunk)
			require.NoError(t, err)
			require.NotNil(t, snapshot)
			require.Len(t, snapshot.pending, 2)
			_, err = db.ExecContext(t.Context(), "INSERT INTO dst SELECT * FROM src")
			require.NoError(t, err)
			passed, err := snapshot.check(t.Context())
			require.NoError(t, err)
			require.True(t, passed, "temporal obligations must resolve regardless of parseTime")
		})
	}
}
```

Both subtests pass at `6c5d99ee`. With the `switch base { case "DATE", ... }` block removed:

```
--- FAIL: TestHotSnapshotTemporalParseTime (0.23s)
    --- FAIL: TestHotSnapshotTemporalParseTime/true
        Error: Should be true
```

The `false` subtest still passes without the cast, which is why the existing temporal case cannot see the difference.

</details>

**3. The comment on `hotSnapshot` states a read-ordering safety property that nothing pins, and swapping the two reads passes the whole suite.**

`lockless_snapshot.go:18-21` says target membership is read *before* the source images "so an already-present orphan is not hidden by checking only source rows". Swapping `:46-53` so the source census is read first leaves `./pkg/checksum/` green. Working through the interleavings I could not find a case where the shipped order is *stricter*: an orphan present across both reads is captured either way, and an orphan deleted between them is a satisfied obligation either way. The one case where the orders differ is an orphan that *appears* between the two reads, and there the shipped order is the weaker one — the target census has already been taken, so the new orphan is not an obligation at all, whereas source-first would have caught it. Neither order closes that window, which is fine and matches the second half of the comment ("later changes remain replication's responsibility"), but the first half reads as a load-bearing guarantee and is not one. Worth either demoting the sentence to "the two reads are independent point-in-time samples; neither order closes the interleaving window" or, if the ordering really is meant to be load-bearing, pinning it.

**4. Two defensive branches in the snapshot reader are unreachable as written.** Both survive mutation, and both are benign; flagging them so the next reader does not assume they are covered.

- `lockless_snapshot.go:86-88` returns `false` when the target re-read is oversized. `pending` is bounded by `len(target)+len(source) ≤ 256`, the predicate is a disjunction of point ranges over a unique key, and the limit passed is `2*hotSplitTargetRows = 256` — so at most 256 rows can come back and `oversized` cannot fire. Inverting it to `return true, nil` (the dangerous direction) changes nothing in the suite.
- `lockless_snapshot.go:165-167` errors on a duplicate identity. Same reason: the key is unique, so two rows cannot share an identity. Deleting the guard changes nothing.

**5. `tryHotSnapshot`'s two admission guards are untested.** `lockless.go:1053` (neither side over 128 rows) and `lockless.go:1056` (at depth 0, require `consecutiveSrcChanged >= 1`) both survive removal. The size guard is partly redundant with the reader's own budget — dropping it still yields a `nil` snapshot for an oversized range, so it costs two wasted queries rather than correctness — but the evidence guard is the same kind of gate #1249 turned out to be load-bearing for the split budget: at depth 0 it is the only thing stopping a first-retry mismatch from paying for a full census before the ordinary retry has had a chance to reconcile replication lag. A case asserting `snapshotChunk` is not called on the first retry of a depth-0 chunk would pin it cheaply.

**6. Operability: nothing bounds how long the experimental gate runs, and the one knob an operator might reach for is silently inert.** `RunUntilClean` (`lockless.go:735-737`) loops passes until one is clean, and the migration wiring (`runner.go:982-993`) leaves `MinPassInterval` at zero, so passes run back-to-back. On the workload the PR itself documents as unsupported — continuously updated hot rows — the Checksum phase runs forever with no error and no deadline. `docs/migrate.md` does say "cancel the run or resume with the default checker", so this is documented rather than surprising, and it fails closed (cutover is never authorised), which is why it is not blocking. Two things would make it easier to live with during burn-in: the status line reports `deferred=N` for the *current* pass only, so an operator watching a long run cannot tell a range that has been stuck for twenty passes from one that just deferred for the first time; and the summary's claim that `--checksum-yield-timeout` "applies only to the default snapshot checksum" is accurate but worth reading as "there is no equivalent bound here", since that flag is the only phase-level timing control the traditional path exposes.

### What I checked and found correct

- `ChecksumExprs` wraps every column in `IFNULL(...,'')` plus an `ISNULL()` marker, so `CRC32(CONCAT(...))` cannot be `NULL` and the `uint64` scan at `:147` is safe; `TestHotSnapshotColumnMapping` covers the renamed-column and NULL-value path.
- The frozen work set does what the summary claims: `TestHotSnapshotFiniteTail` shows later inserts and an update to an already-retired row leaving the remaining obligation untouched, while the aggregate signature still disagrees.
- Both `tryHotSnapshot` call sites (`lockless.go:1149` pre-drain, `:1219` post-drain) sit on the `newSrc != item.originalSrc` branch, so a snapshot is only ever taken for a range the aggregate read has already shown to be changing, and `res.newTgt` is the post-drain value at the second site.
- The snapshot re-enqueue at `lockless.go:1277-1281` sits *before* the `res.permanent` branch, but a snapshot result can never carry `permanent` — `executeWork` short-circuits to `checkHotSnapshot` and returns — so the ordering is not hiding a divergence verdict.
- All the per-pass counters, including `hotChunksDeferredThisPass`, are reset at `lockless.go:669-681`, so `RunUntilClean`'s criterion cannot be poisoned by an earlier pass.
- `DivergenceIsFatal: true` in the migration config is unobservable (no `Recopier` is configured, so `executeWork` reaches the fatal path either way) — correct as belt and braces, not a gap.
- The checkpoint story is well pinned: `TestLocklessCheckpointNeverPersistsChecksumWatermark` and `TestLocklessResumeIgnoresSnapshotChecksumWatermark` both die when their respective guards in `runner.go` are removed, and the second also dies if `GetProgress` reports traversal as verification.
- The `Continuous*` → `Lockless*` rename is complete across `pkg/datasync`, `pkg/move`, `AGENTS.md` and `pkg/checksum/README.md`, and the `#enable-experimental-lockless-checksum` anchor in `docs/migrate.md` resolves.
- No internal infrastructure detail in the diff.

### Suggested follow-ups (not required for this PR)

- The `docs/migrate.md` section says snapshot reads have "a 30-second timeout" — worth noting it is 30s *per call*, so an unresolved range can spend up to `MaxHotAttempts` separate 30s windows plus `RetryDelay` between them.
- A source row deleted after capture can never satisfy its frozen obligation (`present: true`, `exists: false` matches neither arm at `:99`), so a single ordinary delete inside a hot range costs the whole remaining attempt budget and a full extra pass. `TestHotSnapshotOrphansAndDeletes:73-79` asserts exactly this behaviour deliberately; the follow-up would be retiring an obligation whose key is confirmed absent on *both* sides, once there is a cheap way to re-read the source key without reopening the work set.

aparajon (approved):
Approving on `6c5d99eef740c311b4c7bffd8a09355e24e68b65`. 0 blocking, 6 non-blocking — see the review comment above. The snapshot drain's correctness core survives every mutation I could aim at it; the findings are two real guards with no test behind them (the target census's row budget, the temporal `CAST` under `parseTime=true`), two unreachable defensive branches, an untested admission gate, and the unbounded-runtime/observability note on the experimental gate. Nothing here blocks merge.

PR #1254, opened 2026-09-20
fix(table): carry rows copied forward across an optimistic resume

aparajon:
The optimistic chunker keeps two counters. `rowsCopied` is the distance travelled through the auto-increment key space, which is what `Progress()` renders a percentage from. `actualRowsCopied` is the number of rows the applier really settled, which is what `RowsCopied()` returns. On a sparse key space the two are nothing alike, which is why they are separate.

Only the first survived a resume. `OpenAtWatermark` reseeds `rowsCopied` from the restored chunk pointer, so the percentage stays continuous across a stop and restart, but the watermark stored key positions and nothing else, so the settled count restarted from zero. From that point on the two disagreed for the rest of the run, and a consumer deriving a rate, an ETA, or a "rows copied of rows total" line from `RowsCopied()` computed it against a fraction of the work.

This persists the count in the watermark, as the composite chunker already does.

A table of 800,000,000 rows. The first run settles 120,000,000 rows and is stopped; the second resumes from the checkpoint and settles 30,000,000 more.

```
Before

  run 1 settles 120,000,000 rows
    │
    ▼ checkpoint
  ┌──────────────────────────┐
  │ {"Key":["id"],           │
  │  "LowerBound":{...},     │
  │  "UpperBound":{...}}     │
  │                          │
  └─────────────┬────────────┘
                │ resume
                ▼
  ┌──────────────────────────┐
  │ RowsCopied()          0  │  <-- restarts from zero
  │ Progress()          15%  │
  └─────────────┬────────────┘
                │ settles 30,000,000
                ▼
  ┌──────────────────────────┐
  │ RowsCopied()  30,000,000 │  <-- 5x low for the rest of the run
  │ Progress()          17%  │
  └──────────────────────────┘

After

  run 1 settles 120,000,000 rows
    │
    ▼ checkpoint
  ┌──────────────────────────┐
  │ {"Key":["id"],           │
  │  "LowerBound":{...},     │
  │  "UpperBound":{...},     │
  │  "RowsCopied":120000000} │  <-- the count travels too
  └─────────────┬────────────┘
                │ resume
                ▼
  ┌──────────────────────────┐
  │ RowsCopied() 120,000,000 │
  │ Progress()          15%  │
  └─────────────┬────────────┘
                │ settles 30,000,000
                ▼
  ┌──────────────────────────┐
  │ RowsCopied() 150,000,000 │
  │ Progress()          17%  │
  └──────────────────────────┘
```

## Watermark format

The count is a sibling of the chunk's own fields rather than an envelope around them:

```json
{"Key":["id"],"ChunkSize":1000,"LowerBound":{},"UpperBound":{},"RowsCopied":120000000}
```

That keeps the watermark a plain chunk to every reader. `newChunkFromJSON` validates by shape and ignores unknown keys, so it and `WatermarkRecopyClause` parse the new format unchanged, and a binary that predates the field resumes from the bounds exactly as before. A checkpoint written by either version is readable by the other, so the two can run side by side and no rollout ordering is required. The composite chunker's envelope could not offer that: a reader that does not know to unwrap it fails the resume outright and re-copies the table from scratch.

`RowsCopied` is the count as it stood when the checkpoint was written, not the count as of the chunk the watermark points at, so chunks that completed out of order ahead of the watermark are included and are counted again when the resume re-copies them. The composite chunker's envelope has the same property. The overcount is bounded by the in-flight window, and it is consistent with `RowsCopied` being defined as rows settled rather than rows present in the new table.


morgo:
LGTM. If you are ready for review, lets ship it.

JashLal:
🤖 Reviewed at `f483903f05eaff8e9cb1a9f8811b4f9f1be339d7`. No blocking findings.

The new watermark keeps the original chunk fields at the top level and adds only `RowsCopied`, so the existing chunk parser, per-table watermark extraction, and recopy-clause builder retain their previous interpretation. On restore, the actual settled-row counter is loaded after `open()` resets it; a missing field restores zero, while key-space progress continues to use the existing separate calculation. `Feedback` and watermark serialization hold the same chunker mutex, so the saved count and bounds are read from one synchronized state.

The sparse-key round-trip test verifies that the restored value is the actual row count rather than the key-space distance, and the compatibility tests cover older watermarks and the existing recopy readers. I ran `go test -race ./pkg/table -run 'Test(LowWatermark|OptimisticDynamicChunking|OptimisticResumeProgressAccounting|OptimisticWatermark.*)$' -count=1 -timeout=3m`; it passed. All reported GitHub checks are passing. I did not run the MySQL-backed migration/checkpoint suite locally.

One nonblocking documentation clarification: `pkg/table/chunker_optimistic.go:98` says the total can exceed table rows by up to the in-flight window. This is a per-resume accounting effect, not a lifetime cap: repeated resumes can accumulate re-settled work. Describe it as settled work including replay, without implying a fixed overall overcount bound. Consumers should continue to use `IsComplete`, rather than row-count equality, to determine completion.

JashLal (approved):

aparajon:
🤖 Fixed in `72eabb13`.

You're right that the bound was wrong. The sentence stated the overcount as a lifetime cap, but it is per resume: the restored count already includes chunks that completed ahead of the watermark, and every further resume replays its own window on top of the last, so there is no overall bound to state. The comment now describes the value as settled work including replay, and points at `IsComplete` for completion rather than comparing the count against the table's rows.

*Replied by Claude Code (Claude Opus 5) on behalf of @aparajon.*

PR #1263, opened 2026-09-22
Decouple blocker kill delay from lock wait timeout

### LINKED ISSUE #1262: Decouple blocker kill delay from lock wait timeout
🤖 Spirit starts killing transactions that block metadata locks after 90% of `lock-wait-timeout`. With a 10-second timeout, this leaves roughly one second for killed connections to release their locks before the original statement times out.

Please consider an independent `--force-kill-after` duration for `migrate`. For example, `--lock-wait-timeout=10s --force-kill-after=5s` would allow five seconds of grace before killing blockers, with more time left for lock acquisition. Keeping zero as the existing 90% default preserves current behavior.

The proposed change applies to native DDL and table locks, rejects negative delays and delays at or beyond MySQL's effective timeout, and includes tests against blocked INSTANT DDL and table-lock acquisition. Earlier killing can interrupt application transactions sooner; it does not guarantee rollback completes within the remaining wait.

Written by the user's Codex AI agent.

deanbaker:
Closes #1262

Spirit starts killing metadata-lock blockers at 90% of `lock-wait-timeout`. A 10-second timeout leaves roughly one second for killed connections to release locks before the statement times out.

Add `migrate --force-kill-after` and `DBConfig.ForceKillAfter` so the delay can be set independently. For example, `--lock-wait-timeout=10s --force-kill-after=5s` starts killing blockers after five seconds. Zero preserves the current 90% default.

Both native DDL and table locks use the delay. Validation rejects negative values and delays at or beyond the timeout MySQL receives, including whole-second truncation. Existing blocker selection and retry behavior remain unchanged.

Earlier killing interrupts application transactions sooner and does not guarantee rollback finishes before the timeout. The default behavior is unchanged.

Manual check: `go run ./cmd/spirit migrate --help` exposes the new flag and its default. No production testing.

Generated with Codex


morgo:
Makes sense to me. I think high connection counts and very short timeouts are probably the combination which are causing problems.

May I ask for the following docs fixes:
- Remove mention from README.md in parent directory.
- Add to docs/migrate.md.

Otherwise LGTM.

morgo:
🤖 Pushed 5cc6bb05 to this branch with both docs fixes:

- Removed the `Blocker kill timing` section from the top-level `README.md`.
- Added a `force-kill-after` section to `docs/migrate.md` (TOC entry, example, validation rules, caveats) plus a cross-reference from the `lock-wait-timeout` section.

No code changes.

morgo (approved):

